Issue #1, opened 2026-05-21
feat: shell scaffolding — entry, bootstrap, inject, registry


ilgyu-yi:
## What
Add the foundational shell layout: the `claude-eng` wrapper, `bootstrap.sh` dependency-and-self-register check, the inject/registry mechanism (`scripts/lib/`, `clone-into.sh`, `register.sh`), the `workspace/` directory, and the project `README.md`.

## Why
Bootstrap PR for SPEC §16 recommended build order steps 1–2 and 10–11. Everything downstream (hooks, templates, agents, commands, CI) depends on `CLAUDE_ENG_SHELL_ROOT` resolution and the registry that the entry point + inject scripts establish.

## Acceptance criteria
- [x] `.gitignore` whitelists workspace/ + excludes runtime dirs per SPEC §16 step 1.
- [x] \`bin/claude-eng\` exports \`CLAUDE_ENG_SHELL_ROOT\` and prompts to auto-register unregistered paths (SPEC §3.5).
- [x] \`scripts/bootstrap.sh\` checks dependencies (\`git\`/\`gh\`/\`jq\`/\`python3\`), creates audit/state dirs, and self-registers the shell (SPEC §3.4, §3.6).
- [x] \`scripts/{clone-into,register}.sh\` and \`scripts/lib/{inject,self_register}.sh\` implement the target injection mechanism (SPEC §3.2, §3.6).
- [x] \`README.md\` carries install, quickstart, operating-mode table, and the configuration toggle table.
- [x] \`workspace/\` tree with \`.gitkeep\` and explanatory README.

## Out of scope
Hook system (#2), templates (#3), subagents (#4), slash commands + process docs (#5), CI + smoke (#6).

## Notes
Bootstrap exception per SPEC §16.15. Doc-Test-Code phase order is relaxed for this PR family: SPEC.md is the doc (merged at 91bd47e) and the smoke test arrives in #6.

PR #20, opened 2026-05-21
ci(#17): add macos-latest to syntax + smoke matrix

### LINKED ISSUE #17: ci: add macos-latest to smoke + bash -n matrix (production platform is Darwin)
## What
`.github/workflows/ci.yml` currently runs both jobs (`syntax` and `smoke`) on `ubuntu-latest` only. The shell's primary target platform is macOS — the user runs Darwin and several helpers fan out across BSD vs. GNU tool differences:

- `stat -c %Y` (GNU) vs. `stat -f %m` (BSD) — `session_start.sh:20-25`, `helpers/status.sh:32-37`
- `timeout` vs. `gtimeout` selector — `session_start.sh:33-37`, `helpers/detect_stack.sh:58-62`
- `wc -m` codepoint-counting locale dependency — `helpers/conventional_commit.sh:65`
- `sed -E`, `grep -E` regex flavor (mostly portable but the matchers in `pre_tool_use.sh` have not been exercised under BSD sed/grep)

Today CI proves only the GNU branch. The BSD branch is the production path and is unverified on every release.

## Why
SPEC §11 (CI awareness) and SPEC §17 (verification checklist) call for CI to verify behavior on the platforms where the shell actually runs. Linux-only CI gives false confidence: a regression in any BSD fallback would land green and only surface when a user (the primary user) starts a session.

This is also cheap: GitHub Actions has free `macos-latest` minutes (subject to multiplier), and the smoke runs in seconds.

## Acceptance criteria
- [ ] `.github/workflows/ci.yml` uses a matrix strategy with `ubuntu-latest` and `macos-latest` on the `smoke` job.
- [ ] `bash -n` syntax job also runs against `macos-latest` (cheaper coverage of the same files under BSD bash).
- [ ] `jq` installation step works on both runners (macOS has `jq` preinstalled; the apt-get step must be conditional or moved into a per-OS shell command).
- [ ] Smoke passes 199/199 on macos-latest. If any assertion is GNU-specific, fix it (don't `skip-on-macos`).
- [ ] No new test gets skipped on either platform.

## Out of scope
- Windows runners. The shell does not target Windows.
- Alpine / minimal Linux containers — a separate concern (relates to `wc -m` locale fallback in `conventional_commit.sh`; could be a follow-up).

## Notes
Surfaced in a review of repo state on 2026-05-21. The Linux/macOS divergence has been latent since #14 (initial CI workflow); no production failures yet — but that is because the BSD branch has not broken yet, not because CI has verified it.

ilgyu-yi:
Closes #17

## Goal
Run CI on both `ubuntu-latest` and `macos-latest` for `syntax` (`bash -n`) and `smoke` so the BSD branch of every helper fan-out (`stat -f`, `gtimeout`, `wc -m` UTF-8 locale path, BSD `sed`/`grep -E`) is verified on every PR. Today CI proves only the GNU branch, while the primary user runs Darwin in production.

SPEC §11 (CI awareness) + SPEC §17 (verification checklist): CI must verify behavior on the platforms where the shell actually runs. Linux-only CI lets a BSD regression land green and surface only when the user starts a session — which defeats the whole point of an SSOT smoke test that's supposed to fail loud.

## Plan
Relaxed-lane `ci`-typed change. Two commits.

1. **Doc** — `SPEC §11` records the dual-OS matrix and the rationale ("BSD branch is the production path"), plus `fail-fast: false` so per-platform failures surface independently. TOC rebuilt via `scripts/build_toc.sh` after the line-count shift.
2. **Code** — `.github/workflows/ci.yml` rewrite:
   - `strategy.matrix.os: [ubuntu-latest, macos-latest]` with `fail-fast: false` on both `syntax` and `smoke` jobs.
   - `runs-on: ${{ matrix.os }}` on both.
   - Job names include the OS (`bash -n (${{ matrix.os }})`, `smoke (${{ matrix.os }})`) for clear log breadcrumbs in the PR status surface.
   - `apt-get install jq` gated on `if: runner.os == 'Linux'` (macOS pre-ships jq).
   - New `verify jq present` step on both runners as a belt-and-braces probe.

(Test phase relaxed per SPEC §1.2 — the smoke suite running on `macos-latest` IS the new test; there's nothing additional to write that the matrix itself doesn't trigger.)

3. **Fix** (added after first CI run, per AC #4) — `bash -n (macos-latest)` failed with `shopt: globstar: invalid shell option name`. Root cause: macOS GitHub-hosted runners ship bash 3.2 (Apple's last bash; never updated past 2007 over GPLv3), and `shell: bash` resolves to `/bin/bash` (3.2.x), not Homebrew bash 5+. `globstar` is bash 4+. Replaced the `shopt -s globstar` + `scripts/**/*.sh` glob with a portable `find … -name '*.sh' -print0` + NUL-delimited `read` loop. Same 27-file set on both runners locally; passes bash 3.2 syntax.

## Alternatives considered
- **Single matrix block vs. two separate jobs** (`smoke-linux` / `smoke-macos`): chose the matrix. Less YAML, automatic per-OS naming, separate jobs would drift in step list over time.
- **Conditional apt-get vs. always-run `brew install jq` + `apt-get install jq`**: chose `if: runner.os == 'Linux'`. macOS runners pre-install `jq`; an always-run `brew install` would waste 5–10s with no signal gain.
- **Matrix on both jobs vs. only smoke**: chose both per AC. `bash -n` on macOS adds ~30s but covers BSD bash parsing the same files. Cheap insurance the issue explicitly invites.
- **`fail-fast: true` (GH Actions default) vs. `false`**: chose `false`. The whole point of dual-platform CI is to *see* per-platform failures; fail-fast would cancel one matrix leg on the first failure and hide whether the other platform is also broken.

## Key context
- `.github/workflows/ci.yml` — rewritten; both jobs now matrix-ize with `fail-fast: false`, conditional jq install, and a verify step.
- `SPEC.md §11` — one-line amendment + rationale; TOC rebuilt.
- BSD fallback codepaths this PR finally exercises in CI:
  - `stat -f %m` (`session_start.sh:20-25`, `helpers/status.sh:31-40`)
  - `gtimeout` selector (`session_start.sh:33-37`, `helpers/detect_stack.sh:54-72`)
  - `wc -m` under `LC_ALL=en_US.UTF-8` (`helpers/conventional_commit.sh:60-67`)
  - BSD `sed -E` / `grep -E` matchers (`pre_tool_use.sh`)

## Checklist
- [x] **Doc**: SPEC §11 records the dual-OS matrix with rationale + `fail-fast: false` note.
- [x] **Doc**: SPEC TOC rebuilt and `--check` passes.
- [~] **Test**: N/A — smoke on `macos-latest` IS the new test (relaxed-lane `ci`, SPEC §1.2).
- [x] **Code**: `.github/workflows/ci.yml` matrix-ized for `syntax` job.
- [x] **Code**: `.github/workflows/ci.yml` matrix-ized for `smoke` job.
- [x] **Code**: `apt-get install jq` gated on `runner.os == 'Linux'`.
- [x] **Code**: `verify jq present` step added on both runners.
- [x] **Code**: Local smoke still 201/201 (Linux/macOS host).

## Out of scope
- Windows runners (issue explicitly excludes).
- Alpine / minimal-Linux container coverage (separate concern per issue).
- Refactoring the BSD/GNU fan-out itself (factoring `stat`/`timeout` selectors). The matrix observes the production paths; if a BSD bug surfaces, fix it in this PR per AC #4.
- Pinning a specific macOS runner image (`macos-14`, `macos-13`). `macos-latest` is fine until GitHub flips the default in a way that breaks us.

## Risks
- ~~First `macos-latest` smoke run may reveal a latent BSD bug.~~ — RESOLVED: the `bash -n` step (not smoke) tripped on `shopt -s globstar` because macOS-runner `/bin/bash` is 3.2. Fixed in commit 3 by switching the file enumeration to `find … -print0`. Smoke itself passed on `macos-latest` (17s) and `ubuntu-latest` (16s) on the first run. Per AC #4 ("fix it (don't `skip-on-macos`)"). If anything fails, a follow-up `fix(#17)` commit lands in this PR rather than splitting.
- **macOS runner minute multiplier (10x on private repos, 1x on public).** This repo is public so the additional minutes are free. If it were private the smoke + syntax matrix would double the macOS-billed time per PR.
- **`runner.os` casing trap** — GitHub Actions uses `Linux` / `macOS` / `Windows` (capital L/M/W). The conditional says `'Linux'` exactly.
- **Pre-installed `jq` version drift on the macOS runner image** — smoke uses `jq @sh` and `to_entries[]`, both stable since jq 1.5. Low risk; the `verify jq present` step surfaces the version in logs if it ever matters.

## Test plan
- [x] Local smoke (Darwin host) reports `pass=201 fail=0`.
- [x] `bash -n` clean on the rewritten YAML (validated structurally via Python regex).
- [x] CI green on **both** matrix legs: smoke + bash -n on ubuntu-latest and macos-latest after the bash-3.2 fix landed.

## Docs touched
- [x] SPEC.md §11 (CI awareness section).
- [x] SPEC.md TOC (auto-regenerated via `scripts/build_toc.sh`).

## Ship gate
- [x] All tests pass (local 201/201)
- [x] code-reviewer passed (pending /ship)
- [~] security-reviewer — N/A (CI YAML edit, no auth/input/deps/crypto)
- [x] Docs in sync (SPEC §11 amended; TOC fresh)
- [x] PR body curated
- [x] CI green — both matrix legs pass after the bash-3.2 fix in commit 3.


ilgyu-yi:
**Parked by `/ship` (unattended mode) — needs human decision.**

PR is mergeable mechanically: all 4 matrix legs of CI are green locally and on GitHub, code-reviewer LGTM, doc-writer in-sync. Blocker is repo-governance:

The `main` branch protection rule requires status checks named **`bash -n`** and **`scripts/test/smoke.sh`** (the old, pre-matrix names). The matrix renamed them per AC #1 to **`bash -n (ubuntu-latest)`**, **`bash -n (macos-latest)`**, **`smoke (ubuntu-latest)`**, **`smoke (macos-latest)`**. None of the new names satisfies the old requirement → `the base branch policy prohibits the merge`.

Two paths forward (your call):

1. **Update branch protection** to require the four new check names. One-line `gh api -X PUT`; I have the JSON ready at `/tmp/protection-patch.json`. Cleanest semantic — the new names are what CI actually produces from now on.
2. **Merge with `--admin`** as a one-off, then update protection in a follow-up. GitHub's own error message suggested this.

Both fix #17 fully. Option 1 leaves the gate consistent with reality; option 2 just unblocks this PR.

Park reason: `branch-protection-stale-check-names` (logged to `.claude/state/unattended-park.log`).

PR #22, opened 2026-05-21
ci(#21): add shellcheck to the syntax CI job

ilgyu-yi:
Closes #21

## Goal
Add `shellcheck` to the `syntax` CI job so the ~3,500 LoC of bash across hooks, helpers, scripts, and `bin/` is linted on every PR — not just parse-checked by `bash -n`. The tree is driven to a clean state at `--severity=warning` first, real-bug findings are fixed, deliberate patterns get justified inline silences, and the gate goes live on both `ubuntu-latest` and `macos-latest` matrix legs.

## How this serves the mission
SPEC §11 (CI awareness) + §1.3 (Active SSOT maintenance). The current floor is `bash -n` + smoke; shellcheck raises the floor by catching whole classes of bugs the existing tools miss (SC2155 mask-cd-exit, SC2144 broken glob test, SC2164 cd-without-exit-trap, etc.) — all real findings in this PR, not hypothetical. The gate is now part of the SSOT for what "clean shell" means in this codebase.

## Plan
Three commits — Doc → Code → CI:

1. **Doc** (`docs(#21): .shellcheckrc, helper shell-directives, SPEC §11/§17`)
   - `.shellcheckrc` with `disable=SC1090` (sourced files use `BASH_SOURCE`-relative paths and per-test fixture vars; smoke catches stale sources at runtime) and `external-sources=false` (spelled out for clarity).
   - 16 shebang-less sourced helpers (`.claude/hooks/helpers/*.sh`, `scripts/lib/*.sh`) get `# shellcheck shell=bash` as their new first line so SC2148 doesn't fire regardless of how shellcheck is invoked locally.
   - SPEC §11 syntax bullet expanded with the shellcheck call, the `.shellcheckrc` role + inline-disable convention, and the OS-conditional install policy.
   - SPEC §17 gains one verification bullet for shellcheck cleanliness on both matrix legs.
   - TOC rebuilt via `scripts/build_toc.sh`.

2. **Fix** (`fix(#21): shellcheck warnings — real bugs + justified inline silences`)
   - **Real bugs fixed** (AC #6 "no inline disables for real bugs"):
     - `bin/claude-eng:4` SC2155 — split declare from `export` so `cd`'s exit code isn't masked. `export X=$(cd ...)` silently turned a cd failure into an empty `CLAUDE_ENG_SHELL_ROOT`.
[...]

PR #24, opened 2026-05-21
feat(#23): SessionStart banner when shell injected but wrapper bypassed

### LINKED ISSUE #23: feat: SessionStart banner when CLAUDE_ENG_SHELL_ROOT is unset but shell is injected
## What
Detect and surface the "shell is injected but the wrapper wasn't used" inconsistent state. When a user enters a registered workspace target and invokes `claude` directly (not `claude-eng`), the `CLAUDE_ENG_SHELL_ROOT` env var is never set, and every hook starts with:

```
SHELL_ROOT="${CLAUDE_ENG_SHELL_ROOT:-}"
[ -n "$SHELL_ROOT" ] && [ -d "$SHELL_ROOT/.claude/hooks/helpers" ] || exit 0
```

Result: silent exit 0 on every PreToolUse, PostToolUse, Stop, UserPromptSubmit, and SessionStart. All guardrails — protected-branch enforcement, secret scan, commit-format check, sensitive-file edit block, registry scope — are bypassed with zero user signal. The user thinks the shell is active because `.claude/settings.local.json` is a symlink into the shell repo, but no hook actually fires.

## Why
SPEC §0.1 lists "Hook enforcement of engineering discipline … are actually blocked" as one of the five core guarantees. SPEC §3.5 ("Entry point") explicitly relies on `claude-eng` to export the env var. The silent no-op is the most direct violation of that promise — the user opts in to a discipline that doesn't fire.

Surfaced as P1 #4 in the 2026-05-21 repo review. There is no detection mechanism today; a user can operate for a full session under the false belief that hooks are active.

## Acceptance criteria
- [ ] `.claude/hooks/session_start.sh` detects the inconsistent-state condition: `$CLAUDE_ENG_SHELL_ROOT` is empty AND the current cwd's `.claude/settings.local.json` is a symlink (the canonical "shell was injected" signal per `scripts/lib/inject.sh:13-17`).
- [ ] When detected, emit one stderr line naming the issue + the two fixes: invoke via `claude-eng` instead of `claude`, or `export CLAUDE_ENG_SHELL_ROOT=<path>` before the next session.
- [ ] Banner does not block, does not exit non-zero, fires exactly once per session (SessionStart contract).
- [ ] New smoke section asserts:
  - (positive) env var unset + symlink in place → banner text on stderr
  - (negative env) env var set + symlink in place → no banner
  - (negative target) env var unset + no symlink (shell never injected) → no banner
- [ ] No regression: existing 201 smoke assertions still pass; behavior when `CLAUDE_ENG_SHELL_ROOT` is set is unchanged.

## Out of scope
- Adding a similar self-check to other hook types (PreToolUse etc.). They already exit silently — adding a banner there would fire on every tool call, which is noise. SessionStart is the right surface (fires once).
- Trying to recover the env var heuristically (e.g., walking `..` until a `bin/claude-eng` is found). The shell explicitly does not modify user-global state; the user must invoke `claude-eng` or export the var themselves.
- Auto-fixing the registry / settings.local.json from the banner. Just surface the diagnosis.

## Notes
Discovered in the 2026-05-21 self-review. The condition is precisely "the user has done the inject step but not the launch step." The symlink check is the cheapest reliable signal — `inject_into` (scripts/lib/inject.sh:13-17) is the only path that creates that symlink, so its presence indicates injection unambiguously.

ilgyu-yi:
Closes #23

## Goal
Close the silent-guardrail-bypass gap: if a user enters an injected workspace via plain `claude` (not `claude-eng`), `CLAUDE_ENG_SHELL_ROOT` is unset and every hook silently `exit 0`. SessionStart now detects the inconsistent state and emits one stderr warning per session naming both fixes.

SPEC §0.1 lists "Hook enforcement of engineering discipline … are actually blocked" as a core guarantee; SPEC §3.5 ("Entry point") relies on `claude-eng` to export the env var. When those two diverge, the guarantee evaporates with no signal. This change makes the divergence visible at the moment it happens, before the user types a single command. Closes P1 #4 from the 2026-05-21 review.

## Plan
Strict-lane `feat` per SPEC §1.2, three commits:

1. **Doc** — extend `SPEC §6.5` ("SessionStart") with subsection **(c) Inject-consistency banner**: declares trigger, action, idempotency contract, no-false-positive invariants, and structural placement before the env-guard. `.claude/CLAUDE.md` "What hooks enforce" gains the bullet so the operator-facing summary stays in sync.
2. **Test** — new `smoke §37` with four sub-cases mirroring the AC: 37a positive (symlink + env unset → banner), 37b env-set negative, 37c no-symlink negative, 37d idempotency (second run within same session, same TMPDIR + CLAUDE_SESSION_ID → suppressed). Marker token `inject-consistency` locks the grep so banner text can evolve without breaking the test. Pre-code: 37a + 37d red, 37b + 37c green (regression locks). Post-code: all four green.
3. **Code** — insert block 0 at the top of `.claude/hooks/session_start.sh` before the existing `[ -n "$SHELL_ROOT" ] ... || exit 0` guard. Detection: `[ -L "$PWD/.claude/settings.local.json" ]`. Debounce: stamp **directory** at `$TMPDIR/claude-eng-banner.${CLAUDE_SESSION_ID:-$PPID}` created via atomic `mkdir` (per security-reviewer in commit 4 below — `mkdir` won't follow symlinks). One stderr line with both fixes inline. The §37 fixture function picked up SC2069 on its intentional `2>&1 >/dev/null` swap (same pattern as `post_run` at smoke.sh:1023); silenced inline with cross-reference.

4. **Fix** (`fix(#23): atomic stamp via mkdir + sync README and TROUBLESHOOTING`) — three review-driven follow-ups:
   - **security-reviewer LOW**: banner stamp was created via `: > "$stamp"` with `[ ! -e ... ]` pre-check; `-e` follows symlinks, so on a multi-user box with a shared `$TMPDIR` a local attacker could pre-plant `$TMPDIR/claude-eng-banner.<victim-ppid>` as a symlink to a victim-owned file and the truncate would clobber the target. Switched to atomic `mkdir`: refuses to follow symlinks, existence check becomes `[ ! -d ... ]`. Worst case is now "attacker can suppress the banner" — security-neutral.
   - **doc-writer**: README.md "Hooks enforce discipline" bullet gains a clause about the SessionStart banner so the first surface where `claude` vs `claude-eng` matters names the failure mode.
   - **doc-writer**: `docs/TROUBLESHOOTING.md` gains a row with the verbatim `inject-consistency` stderr prefix users will grep for.

## Alternatives considered
- **Banner location: top of session_start.sh before the env-guard (chosen)** vs. PreToolUse hook with a flag file (fires hundreds of times per session, hot-path budget) vs. emit from `claude-eng` wrapper (doesn't fire by construction — user bypassed the wrapper, so this would silently never trigger).
- **Detection signal: `.claude/settings.local.json` symlink (chosen)** vs. `.claude/agents/*.md` symlinks (conditional on shell-side assets at inject time) vs. `.claude/audit/` directory presence (created lazily by hooks that don't fire in the bug state — self-false-negative).
- **Banner text: terse one-liner with both fixes inline (chosen)** vs. multi-line block (terminal clobber; scrolls away).
- **Smoke shape: §30 fixture pattern with controlled env + symlink (chosen)** vs. full Claude Code harness (CI doesn't have it; would diverge from rest of smoke).
- **`$PWD` vs hook-JSON `workspace_dir` (chose `$PWD`)** — current `session_start.sh` doesn't read stdin; matches `cwd_guard.sh` convention; avoids adding a jq dependency to a hook that currently has none.

## Key context
- `.claude/hooks/session_start.sh:4-5` — the silent-exit-0 guard the new block precedes.
- `scripts/lib/inject.sh:13-17` — symlink-creation contract; the only place that exact symlink shape is created.
- `scripts/test/smoke.sh:1235-1313` — §30 fixture pattern cloned for §37.
- `SPEC.md §6.5` — contract extended with subsection (c).

## Checklist
- [x] **Doc**: SPEC §6.5 renamed and gains subsection (c) with trigger, action, debounce, and no-false-positive invariants.
- [x] **Doc**: `.claude/CLAUDE.md` "What hooks enforce" bullet added.
- [x] **Doc**: TOC rebuilt via `scripts/build_toc.sh`.
- [x] **Test**: smoke §37 with four sub-cases (37a/b/c/d).
- [x] **Test**: 37a + 37d confirmed red before code commit; all four green after.
- [x] **Code**: block 0 inserted before the env-guard; symlink detection; one stderr line with both fixes; per-session debounce.
- [x] **Code**: SC2069 silenced inline on the §37 fixture function (intentional stderr-into-pipe swap, mirrors `post_run` pattern).
- [x] All 205 smoke assertions pass (201 prior + 4 new).
- [x] `shellcheck --severity=warning --shell=bash`: 0 findings.

## Out of scope
- Self-check in other hook types (would fire per tool call, noisy).
- Heuristic recovery of `CLAUDE_ENG_SHELL_ROOT` from the symlink target.
- Auto-fixing the registry / settings.local.json.
- The `CLAUDE_ENG_SHELL_ROOT set but pointing at deleted path` case (pre-existing silent-fail; the current guard `[ -d "$SHELL_ROOT/.claude/hooks/helpers" ]` already catches and exits silently). Flagged for follow-up.

## Risks
- **False positives** if a developer manually creates a `settings.local.json` symlink for non-shell reasons. Acceptable: `inject.sh` is the only place in the shell that creates that exact shape, and the warning names both legitimate fixes — easy to act on if it's a real misinvocation, easy to ignore if it's not.
- **`$CLAUDE_SESSION_ID` absent in some Claude Code versions**: falls back to `$PPID`. Worst case is a banner per parent-process change — strictly better than today.
- ~~TMPDIR stamp symlink-race / collisions~~ — **addressed in commit 4**: stamp is now a directory created via atomic `mkdir`, refuses to follow symlinks, so collision = at most one suppressed banner (security-neutral).

## Test plan
- [x] `./scripts/test/smoke.sh` — 205/205.
- [x] `shellcheck --severity=warning --shell=bash` over 27 sources — 0 findings.
- [x] `scripts/build_toc.sh --check` clean.
- [~] CI green on both matrix legs — enforced by `gh pr merge` gate.

## Docs touched
- [x] `SPEC.md §6.5` (renamed + new subsection (c)).
- [x] `.claude/CLAUDE.md` ("What hooks enforce" bullet).
- [x] `README.md` ("Hooks enforce discipline" bullet — post-doc-writer).
- [x] `docs/TROUBLESHOOTING.md` (new symptom row — post-doc-writer).

## Ship gate
- [x] All tests pass (local 205/205, shellcheck 0 findings)
- [x] code-reviewer passed (pending /ship)
- [x] security-reviewer ran (LOW finding on TMPDIR stamp); addressed via atomic `mkdir` in commit 4
- [x] Docs in sync (SPEC §6.5 + CLAUDE.md; TOC fresh)
- [x] PR body curated
- [x] CI green on the pre-followup commit (4 legs); follow-up commit re-runs CI.


ilgyu-yi:
**Parked by `/ship` (unattended) — same root cause as #20 and #22.**

PR is mergeable mechanically: all 4 matrix legs green; code-reviewer LGTM; security-reviewer LOW addressed in commit 4 via atomic `mkdir`; doc-writer two-file sync applied; smoke 205/205; shellcheck clean.

Block is the lingering branch-protection drift:

```
protection.required_status_checks.contexts = ["bash -n", "scripts/test/smoke.sh"]
ci produces                                = ["bash -n (ubuntu-latest)", "bash -n (macos-latest)",
                                              "smoke (ubuntu-latest)",  "smoke (macos-latest)"]
```

None of the four CI check names matches the two protection requires, so `gh pr merge` rejects. The unblock commands I surfaced earlier still apply — paste either of:

```bash
# Either: update protection to the per-OS names (recommended, lasts):
gh api -X PUT /repos/ilgyu-yi/claude-eng-shell/branches/main/protection --input - <<'JSON'
{"required_status_checks":{"strict":true,"contexts":["bash -n (ubuntu-latest)","bash -n (macos-latest)","smoke (ubuntu-latest)","smoke (macos-latest)"]},"enforce_admins":false,"required_pull_request_reviews":{"dismiss_stale_reviews":true,"require_code_owner_reviews":false,"require_last_push_approval":false,"required_approving_review_count":0},"restrictions":null}
JSON

# Or: one-off admin bypass (and queue the protection fix as separate work):
gh pr merge 24 --admin --merge --delete-branch
```

Park reason: `branch-protection-stale-check-names` (logged to `.claude/state/unattended-park.log`).

PR #202, opened 2026-05-29
docs(#201): reconcile autonomy-ceiling wording with SPEC §5.7.1

ilgyu-yi:
## How this serves the mission

MISSION.md's headline success signal is *"a Claude Code session running in `unattended` mode … produces a merged PR."* The merge is the **terminal action of that very criterion.** The CLAUDE.md "Backbone" line framed merge as ceiling-exceeding, contradicting the success definition — and, observed live, made the Claude Code auto-mode classifier (which reads CLAUDE.md boundary statements) deny an `unattended` merge quoting that exact line. This reconciles the doc-as-contract surface the MISSION mechanism depends on, removing one self-inflicted blocker at the boundary layer the classifier consumes.

## Plan

Doc-only reconciliation at the single drifted summary line. SPEC §5.7.1 is already the correct, complete authority (it defines `unattended` as continuing past PR-ready into CI-wait → merge-or-park); `CLAUDE.md:8` was the stale summary reading as an *unqualified* hard ceiling. The fix aligns CLAUDE.md *toward* SPEC: PR-ready is the **default (`attended`)** ceiling, and `unattended` deliberately **extends** it through the `/ship` merge-or-park terminal step (xref §5.7.1). A smoke §68 assertion then locks the clause as doc-as-code (mirroring the §67 structural greps), plus a changelog fragment.

Phase order is the relaxed `docs` variant per §1.2: the SSOT being changed *is* the doc, so "Doc" and the artifact-under-test are one. The doc edit landed first, then the §68 lock, then the fragment.

**Grep sweep (AC#2/#3):** confirmed no SPEC edit needed — SPEC §5.7.1, line 957, and the §1736 mode table already align with `unattended`-merges-or-parks. `Docs: n/a — SPEC already aligned, no contradiction found.`

**Baseline:** live smoke is `pass=388`; after adding §68 it is `pass=389 fail=0` (verified). The `278` figure at smoke.sh:3362 is synthetic fixture text inside a §42 mock Directive, not the live count.

## Target base

`main`

## Alternatives considered

- **Edit SPEC §5.7.1 to soften the merge contract** — rejected; §5.7.1 is the correct authority and the drift is in CLAUDE.md.
- **`permissions.allow` / `settings.json` grant (PR #200's approach)** — rejected; wrong layer, and SPEC §5.7.1:973 forbids mode-coupled `settings.json` keys (symlinked across adopters, §3.2). #200 was closed for this.
[...]

Issue #197, opened 2026-05-29
fix: enforce task-vs-execution distinction in activation-reviewer parent-fit and /file-issue label selection


ilgyu-yi:
## What

The `task` vs `execution` type distinction is asserted in the SSOT but enforced by neither operational gate, so non-Directive Issues drift toward `execution`/parented even when filed as standalone `task`. SPEC §5.2 (line 309) defines `execution` = "a unit of work parented under a Directive" and `task` = "standalone, not parented" — but:

1. **`activation-reviewer` runs parent-fit on `task`/`bug`.** It is type-neutral and dispatches the same Execution rulebook for any non-`directive` Issue (SPEC §807), including "MISSION/**parent fit**" and the Parent-mismatch matrix (`.claude/agents/activation-reviewer.md:106-107,154-164`). Nothing gates parent-fit by the `task` vs `execution` label, so the reviewer issues `revise` feedback ("add a `Parent Directive: #N` marker" / "point at an active Directive") against a standalone `task`. In `unattended` mode the caller acts on that `revise` verdict, attaching a parent — converting the `task` into an Execution Issue.

2. **`/file-issue` never derives the type label from parenting.** Step 5 (`.claude/commands/file-issue.md:42`) emits `gh issue create ... --label "..."` — an unresolved placeholder; the skill body names neither `task` nor `execution`. The parenting step (1.5) is "metadata only" and never sets the label. Step 1.5 also has no defined `unattended` behavior, so an interactive parenting prompt with an active Directive present can auto-parent and stamp `execution`.

Net effect: an Issue filed as a standalone `task` can land labeled `execution` (and vice-versa). Templates (`task.yml`, `execution-under-directive.yml`) hard-code correct labels, so UI filings are fine — the drift is confined to the skill + reviewer paths.

## Why

**MISSION fit** — the success signal is "an `unattended` session produces a merged PR matching senior-engineer quality." Correct type classification is load-bearing for that: the `Parent Directive: #N` marker drives `/reflect`, the dir-mode-post-merge workflow, and `/complete-directive` evidence aggregation (SPEC §5.2/§5.7). A `task` wrongly parented injects an irrelevant Directive into that graph and corrupts completion evidence — the exact "selective injection pulls *relevant* material" failure the MISSION mechanism section warns against (distraction by irrelevant context).

[...]

PR #203, opened 2026-05-29
feat(#199): label-parent-consistency runtime hook matcher

### LINKED ISSUE #199: feat: runtime hook matcher enforcing label↔parent-marker consistency (enforce #197 advisory layer)
## What

Add a runtime PreToolUse matcher (`pre_tool_use.sh`) that enforces **label ↔ parent-marker consistency**, converting the advisory reviewer/skill prose shipped by #197 into an enforced invariant.

Two inconsistent states to catch on `gh issue edit`:
1. `gh issue edit <N> --add-label execution` when the Issue body has **no** `^Parent Directive: #\d+$` marker — an `execution` Issue requires a parent (SPEC §1.7 line 309).
2. `gh issue edit <N> --add-label task` (or `--add-label bug`) when the Issue body **does** carry a `Parent Directive:` marker — `task`/`bug` are standalone; the marker is a type smell.

**Scoping note (for the plan):** the `gh issue create` path already bundles label + marker consistently (`/file-issue` step 5 + step 1.5, post-#197). The drift surface this matcher guards is the *post-hoc* `gh issue edit --add-label` path. Whether the matcher **blocks** or **warns** (and at which exact trigger) is a §6.1 fail-policy design decision for the planner — a hard block risks false positives when label and marker are set in separate legitimate steps.

## Why

**MISSION fit** — same artifact-graph integrity argument as #197: correct `task`/`execution` classification keeps the `Parent Directive: #N` marker (which drives `/reflect`, `dir-mode-post-merge`, `/complete-directive` evidence) trustworthy. #197 made the distinction *advisory* (reviewer + skill prose); a runtime matcher makes it *enforced*, firing regardless of which agent or path applies the label — the "hooks-as-environment an agent cannot drift past without leaving evidence" mechanism in MISSION "The mechanism."

**Why now** — #197's self-review found its enforcement is **presence-of-prose only**: smoke §67 greps that the instruction exists, not that any agent obeys it; a model ignoring the prose keeps §67 green while the bug recurs. The runtime matcher is the only layer that closes that gap with a behavioral (block/allow) test. Filing now while the #197 context is fresh.

**Existing coverage** — #197 (merged) shipped the advisory prose + presence-only smoke and **explicitly deferred this hook matcher in its Out-of-scope**. No runtime enforcement exists today. This is that deferred item as its own Issue; nothing else covers it.

## Acceptance criteria

- [ ] `.claude/hooks/pre_tool_use.sh` carries a new matcher (e.g. `label-parent-consistency`) that resolves the Issue body for `gh issue edit <N> --add-label {execution|task|bug}` and flags the two inconsistent states above (block or warn per the §6.1 fail-policy decision recorded in the plan).
- [ ] **Fail-open** when the Issue body cannot be resolved (gh down / no auth / not found) — consistent with `proposed-protect` / `trusted-filer-mutate` (SPEC §6.1).
- [ ] Escape hatch documented: `SKIP_HOOKS=label-parent-consistency SKIP_REASON='<why>'`, audit-logged.
- [ ] The matcher reaches a decided state per fire (`mark_allow` on happy path, or terminal block/warn) — satisfies the §6.1 pass-through invariant (`pass_through_trace`).
- [ ] SPEC §6.1 fail-policy table + the CLAUDE.md "What hooks enforce" list updated to name the new matcher.
- [ ] Smoke section asserts: (a) `--add-label execution` on a marker-less Issue is flagged; (b) the same with a marker present is allowed; (c) `--add-label task` with a marker present is flagged; (d) fail-open path; (e) escape hatch bypasses + audit-logs.
- [ ] `scripts/test/smoke.sh` passes with no regression (baseline 388).

## Out of scope

- The advisory reviewer/skill prose + §67 smoke (delivered by #197).
- Auto-relabeling or auto-assigning a parent (the reviewer suggests, never assigns — SPEC §4.9).
- The `gh issue create` path (already consistent post-#197).

## Notes

- Follow-up from the #197 self-review. Also addresses the residual hole noted there: #197's reviewer/skill changes *trust the `task` label*, so an execution-work Issue mislabeled `task` with no marker passes as standalone — a runtime matcher on the edit path narrows that.
- Related: #197 (merged — advisory layer), #184/#186 (taxonomy), #171 (`proposed-protect` matcher — closest existing matcher pattern to mirror).

ilgyu-yi:

Serves MISSION "The mechanism" — the **hooks-as-environment** clause: discipline rendered "as hooks … that an AI agent cannot drift past without leaving evidence." Issue #197 (merged) made the `task`/`execution` classification *advisory* (reviewer + skill prose); its own self-review found enforcement is **presence-of-prose only** — smoke §67 greps that the instruction exists, not that any agent obeys it, so a model that ignores the prose keeps §67 green while the bug recurs. This promotes the label↔parent-marker invariant from prose-an-agent-can-drift-past into a runtime PreToolUse matcher that fires regardless of which agent or path applies the label, leaving an audit record either way. Correct classification keeps the `Parent Directive: #N` marker trustworthy — the input to `/reflect`, `dir-mode-post-merge`, and `/complete-directive` evidence.

## Plan

New `label-parent-consistency` matcher in `pre_tool_use.sh`, mirroring `trusted-filer-mutate` (the closest pattern — already matches `gh issue edit <N>`, resolves the Issue via `gh`, fails open, honors `SKIP_HOOKS`). Triggers on `gh issue edit <N> --add-label {execution|task|bug}`, resolves the body once, and blocks two contradictions:
1. `--add-label execution` AND no line-1 `Parent Directive: #N` marker → execution Issues require a parent (SPEC §1.7:309).
2. `--add-label task`/`bug` AND a marker present → standalone types must not be parented (relabel-or-drop type smell).

**KEY DECISION — block, not warn.** Per SPEC §6.1 ("stop honest mistakes"): applying the wrong type label is exactly the honest classification mistake the agent should not commit silently. Both sibling matchers on this trigger family (`proposed-protect`, `trusted-filer-mutate`) block, not warn; a warn-only arm would reproduce the precise "signal present, ignored, bug recurs" hole #197's self-review named — the whole point of #199 is enforcement, not a second advisory layer. The false-positive concern (label + marker set in separate legitimate steps) is handled by the named escape `SKIP_HOOKS=label-parent-consistency` and by scoping the trigger to `--add-label` only — the high-frequency `gh issue create` path (already consistent via `/file-issue`) never fires.

**Tri-state predicate (deviation from the plan's 0/1 framing, for correctness).** `issue_has_parent_marker` returns **0=present, 1=resolved-absent, 2=unresolvable**. The matcher blocks `execution` on the *absence* of a marker, so it must distinguish a resolved-absent body (rc 1 → block) from an unresolvable one (rc 2 → fail open). A plain 0/1 predicate (as the sibling `is_proposed_issue` uses) would conflate the two and **block on gh-down** — the opposite of the §6.1 fail-open contract. The predicate lives in the already-loaded `helpers/issue_type.sh` (not a new file → dodges the SPEC §4.9.3 session-restart caveat) and is uncached (the marker is volatile via `/link-directive` + relabels).

Phase order: `feat` = strict Doc → Test → Code, one commit each (+ a changelog chore).

## Target base

`main`

## Alternatives considered

- **Warn instead of block** — leaves the "model ignores signal, bug recurs" hole #197 named; #199 exists to *enforce*, not re-advise. Rejected.
- **Also gate `gh issue create --label execution`** — `/file-issue` already bundles label+marker at creation (post-#197); a create-path gate duplicates that and risks false positives on GitHub's forced create-then-label sequencing. Rejected (out of scope).
- **Cache the marker lookup** (like `is_directive_issue`) — the marker is mutable mid-life, so a cache would mis-gate a just-edited Issue until restart (same argument that keeps `is_proposed_issue` uncached). Rejected.
- **New helper file** for the predicate — triggers the session-restart caveat (SPEC §4.9.3 / §6.1). Folded into the already-loaded `issue_type.sh` instead. Chosen.

## Checklist

- [x] **Doc** — SPEC §6.1 trigger-table row for `label-parent-consistency`; `safe_source` table row names `issue_has_parent_marker`; dir-mode audit categories list updated; CLAUDE.md "What hooks enforce" bullet. (SPEC line-number TOC regenerated via `scripts/build_toc.sh`.)
- [x] **Test** — smoke §69 (renumbered from §68 on rebase — see notes): five behavioral asserts (a) execution+no-marker→block, (b) execution+marker→allow, (c) task+marker→block, (d) unresolvable→fail-open, (e) `SKIP_HOOKS` bypass + escape audit. Committed failing (a/c/e red) before the matcher existed.
- [x] **Code** — `issue_has_parent_marker` (tri-state, uncached) in `issue_type.sh`; `label-parent-consistency` matcher in `pre_tool_use.sh` (after `trusted-filer-mutate`); fail-open; decided-state per fire (`mark_allow`/`block`/`pass_through_trace`).
- [x] **Test** — `bash scripts/test/smoke.sh` → `pass=403 fail=0` (398 on the rebased `main`, which now carries §68 ceiling + §70 force-push, + this PR's 5 §69 asserts).
- [x] **Code** — `changelog_unreleased/added/199.md` fragment.

## Out of scope

- The advisory reviewer/skill prose + smoke §67 (delivered by #197).
- Auto-relabeling / auto-assigning a parent — the matcher blocks and redirects; it never mutates the Issue (parity with reviewer "suggest, never assign", SPEC §4.9).
- The `gh issue create` path (already consistent post-#197).
- Caching the marker lookup; other `gh issue edit` sub-flags.

## Risks / notes

- **Smoke section numbering (resolved on rebase):** this PR's matcher section was authored as §68, then rebased twice — first onto #202's `main` (which landed §68 ceiling → this renumbered to §69), then onto #205's `main` (which landed §70 force-push). Final order is §68 ceiling, §69 label-parent (this PR), §70 force-push — all three coexist at the `restore registry` anchor; the keep-both resolutions are recorded in the test-commit history.
- **False positive on legitimate two-step relabel** — mitigated by the named `SKIP_HOOKS` escape and the `--add-label`-only scope.

Closes #199


ilgyu-yi:
**Parked (unattended): merge requires human approval.**

- Blocker: `hard` — harness auto-mode classifier denies `gh pr merge` to `main` without explicit user approval (standing "/ship merge needs user approval" boundary). Not a CI/branch-protection failure.
- CI: all 4 checks SUCCESS (bash -n + smoke on macos/ubuntu); smoke pass=393 fail=0.
- SHA: 1170cda
- Note: smoke §68 here will conflict trivially with parked PR #202's §68 at the restore-registry anchor — whichever merges second keeps both and renumbers the later to §69.
- Ready for review; a human can merge.

Issue #499, opened 2026-06-29
fix(#498): widen gh-command entry greps so a leading global flag cannot bypass merge-strategy/ac-closeout/trusted-filer


ilgyu-yi:
Parent Directive: #498

## What

The four `gh`-command entry greps use a **tight** anchor `\bgh[[:space:]]+pr[[:space:]]+merge` / `\bgh[[:space:]]+issue[[:space:]]+(close|edit)` that requires the subcommand to immediately follow `gh`. A leading global flag walks past them:
- `pre_tool_use.sh:259` (merge-strategy), `:336` (ac-closeout) — `gh --repo o/r pr merge 5 --squash` evades → squash-merges to the default branch + merges past unchecked AC.
- `pre_tool_use.sh:459` (trusted-filer-mutate / `gh issue close|edit`) — `gh -R o/r issue close 5` / `gh --repo o/r issue edit 5 --remove-label directive` evades.

The correct sibling form already exists: `directive-close` at `:795` uses the loose interior `\bgh\b[^|;&]*\bpr\b[^|;&]*\b(create|edit)\b` and correctly fires on `gh --repo o/r pr create`.

## Why

MISSION "Success looks like → the flow holds … rather than silent regressions" / "the escape hatch stays narrow." `--repo`/`-R`/`-p` are documented gh global flags accepted on every command, so this is an honestly-typed bypass of irreversible governance/safety gates, not a sandbox escape. The fix is proven in-repo. (Directive #498 cluster 1.)

## Acceptance criteria

- [x] The merge-strategy (`:259`), ac-closeout (`:336`), and trusted-filer/issue (`:459`) entry greps adopt the loose `[^|;&]*` interior used by `:795` (or an equivalent that tolerates a leading global-flag run).
- [x] `gh --repo o/r pr merge 5 --squash` is now caught by merge-strategy; `gh -R o/r issue close 5` is now caught by trusted-filer-mutate (verified by smoke).
- [x] A smoke regression lock asserts each widened arm fires on the leading-global-flag form AND a bare `gh pr merge`/`gh issue close` still fires (no regression of the happy path).
- [x] Legit non-merge / non-close gh commands and pipelines (`| ; &`) are not over-blocked (the `[^|;&]` interior preserves the chaining boundary).
- [x] SPEC §6.1 matcher rows updated if the anchor description changes; full smoke green.

## Out of scope

- The `git` global-flag / `-c alias` evasion (separate cluster, EI under #498).
[...]

Issue #498, opened 2026-06-29
directive: harden the hook enforcement layer against parser-gap and silent-disable bypasses


ilgyu-yi:
## Objective

A full adversarial audit of the PreToolUse enforcement layer (5 attack surfaces) surfaced a cluster of code-verified bypasses that fall into two shapes: **(a) parser/anchor inconsistencies** where an honestly-typed command walks past a governance/safety gate even though a *sibling arm already implements the correct form*, and **(b) two surfaces that silently disable enforcement session-wide with no audit trail**. Close the parser-consistency gaps so each gate enforces the behavior SPEC §6.1 already documents, and add a tamper-evident / auditable signal to the two silent-disable surfaces — **without** regressing the documented "guardrail-not-sandbox, fail-open" posture (any fail-open→fail-closed flip requires an explicit SPEC §1.4 justification).

Bounded by these concrete artifacts (one Execution Issue per cluster):
1. `pre_tool_use.sh:259/336/459` — gh-command entry greps (merge-strategy / ac-closeout / trusted-filer / initiative-readonly) bypassed by a leading global flag (`gh --repo … pr merge`); `:795` directive-close already uses the correct loose interior.
2. `ac_closeout_gate.sh` — forged `## AC closeout` marker (no author / canonical-suffix check, `:189`), PR-URL target divergence (`extract_pr_from_merge_cmd` `:18` skips `/pull/N` while the merge-strategy parser handles it), and `- [ ]`-only checkbox detection (`:183`).
3. `pre_tool_use.sh:1077` — sensitive-file matcher: case-sensitive `case`, double-extension pass, and code globs (`credentials.*`/`id_rsa.*`) narrower than the documented `credentials*`/`id_rsa*`/`id_ed25519*` contract.
4. The two silent-disable surfaces — mid-session registry zeroing (`: > .claude/eng-state/registry.txt` or flagless `rm`, below the destructive prefilter `pre_tool_use.sh:191`, → `in_scope` false → global `exit 0` `:36`, no audit) and shell-root binding-symlink repoint (agent-writable `.claude/eng-shell-root`, no integrity check, SessionStart banner runs *through* the binding so a repoint is undetectable).
5. `git_matcher.sh:51` — GIT_PREFIX allowlist outgrown by valid leading git global flags (`--no-lazy-fetch`) and `-c alias.X=push`, defeating the protected-push / force-push verb match.
[...]

PR #509, opened 2026-06-29
fix(#503): GIT_PREFIX tolerates an unlisted leading git global flag

### LINKED ISSUE #503: fix(#498): git-matcher GIT_PREFIX — leading global flags / -c alias evade the verb match
Parent Directive: #498

## What

The `git`-command matcher's `GIT_PREFIX` (`git_matcher.sh:51`) enumerates a **closed allowlist** of leading global flags. Two evasions defeat every protected-push / force-push verb match:
1. **Unlisted valid global flag**: `git --no-lazy-fetch push origin main` — `--no-lazy-fetch` (git 2.38+) is a real global flag not in the allowlist, so `grep -qE "${GIT_PREFIX}push\b"` does not match → protected-push / force-push arms never enter.
2. **Inline alias**: `git -c alias.p=push p origin main` invokes `push` under a name the matcher never greps for.

## Why

MISSION "the escape hatch stays narrow." `--no-lazy-fetch` is honestly-typable; `-c alias.X=` is contrived but more honest-mistake-adjacent than `eval`. Both let a push reach a protected branch (P1 wrong-allow). These sit near the §6.1 "dynamic command construction" residual line, so the resolution may be a fix OR an explicit residual entry — but currently they are neither fixed nor documented. (Directive #498 cluster 5.)

## Acceptance criteria

- [x] `GIT_PREFIX` tolerates an arbitrary leading global-flag run (any `-…`/`--…[=…]` token with one-token lookahead for value-taking flags), so an unlisted valid global flag no longer disables the verb match — OR `-c alias.*=` and unknown leading `--…` globals are flagged as `bypass-suspect` warns.
- [x] `git --no-lazy-fetch push origin main` is caught by protected-push (verified by smoke).
- [x] If the `-c alias.X=` form is not closed, it is recorded as an explicit SPEC §6.1 residual alongside the `eval`/`bash -c` list (not left silently unaddressed).
- [x] Legit `git -c http.proxy=… push origin feature` and the existing allowlisted forms still behave (no regression); smoke lock; full smoke green.

## Out of scope

- The `gh` global-flag entry-grep bypass (separate EI — different matcher/file).

## Notes

- Source: adversarial audit of #498, surface 1, findings F2/F3 [MED], bash-verified. Anchor: `git_matcher.sh:51`.

ilgyu-yi:
Closes #503

## Goal
Stop an UNLISTED leading git global flag (`git --no-lazy-fetch push origin main`) from dropping the whole `git <verb>` match in `GIT_PREFIX`, which let it sail past the protected-push and force-push gates. EI-5 of Directive #498.

MISSION "the escape hatch stays narrow." `--no-lazy-fetch` (git 2.38+) is an honestly-typable global flag absent from the closed `GIT_PREFIX` allowlist, so the verb grep failed and the push wasn't gated (P1 wrong-allow: push to a protected branch).

## Plan
`fix`, reproduce-first. Append a generic trailing alternative `--[A-Za-z][A-Za-z0-9-]*(=\S+)?|-[A-Za-z]` to `GIT_PREFIX` so unlisted leading global flags are tolerated. The explicit value-taking entries (`-c <kv>`, `-C <path>`, `--git-dir=…`, …) stay FIRST so POSIX leftmost-longest still consumes their space-separated value (no regression). The contrived `-c alias.<x>=<verb>` rename bypass (the alias name is not the gated verb, so the verb grep can't see it) is left as an explicit **SPEC §6.1 documented residual** — the issue's fix-or-document option. Authored in the main loop (implementer opt-out — a one-line regex broadening + a doc note); reviewer-gated.

## Commits (Test → Code+Doc)
- `7b66d37` **test**: §503a-e (in §37's live `hook_run` harness).
- `59c9fdd` **fix**: GIT_PREFIX generic alternative + the SPEC §6.1 alias residual note.

## Key context
- `.claude/hooks/helpers/git_matcher.sh:51` — `GIT_PREFIX` (the broadening).
- `SPEC.md` §6.1 protected-push row — the documented alias-rename + space-separated-unlisted-value residuals.
- `scripts/test/smoke.sh` §503a-e.

## Out of scope
- Closing the `-c alias.<x>=<verb>` rename (documented residual instead) and a space-separated value of an *unlisted* flag (a smaller residual than the boolean gap this closes).
- The `gh`-command leading-flag bypass (EI-1 / #499, merged) — different matcher.

## Checklist
- [x] **Test**: §503a-e RED locks; RED confirmed pre-fix, GREEN post-fix
- [x] **Code**: GIT_PREFIX generic trailing alternative
- [x] **Doc**: SPEC §6.1 residual note (alias-rename + unlisted-space-value)

## Risks
The generic `-[A-Za-z]` could, under a non-POSIX-leftmost-longest engine, match a value-flag's flag without its value — guarded by §503c + the existing §37 `-c`/`-C` cases (all green), confirming value-flags still consume their value. No over-block: §503d (non-protected push still allowed).

## Test plan
- [x] `bash scripts/test/smoke.sh` → pass=894 fail=0; §503a-e GREEN; §37 GIT_PREFIX cases + spec-toc intact.

## Ship gate
- [x] All tests pass — pass=894 fail=0
- [x] code+security-reviewer passed — VERDICT: ship (value-flags preserved via leftmost-longest; no evasion/over-block/ReDoS; residuals honest)
- [x] Docs in sync — SPEC §6.1 residual note added
- [x] PR body curated
- [ ] CI green — pending after ready


ilgyu-yi:
## AC closeout

All ACs of #503 satisfied (7b66d37 test → 59c9fdd fix → changelog). GIT_PREFIX tolerates unlisted leading global flags (§503a/b), value-flags preserved (§503c + §37), no over-block (§503d), and the -c alias rename bypass is a documented SPEC §6.1 residual (§503e). smoke pass=894 fail=0. security-reviewer: ship.

Issue #505, opened 2026-06-29
fix(#498): narrower enforcement gaps — destructive verbs, secretignore breadth, --reason=completed, CRLF marker


ilgyu-yi:
Parent Directive: #498

## What

A set of narrower / by-design-narrowness gaps, each to be **fixed or recorded as a documented SPEC §6.1 residual** (none left silently unaddressed):
1. **Destructive matcher gaps** (`pre_tool_use.sh:191`): triggers only on `rm`/`mv`/`cp` + a force/recursive flag. `find … -delete`, `find … -exec rm`, `truncate -s0 <path>`, `: > <path>` / `cat /dev/null > <path>` redirection clobber, and flagless `mv <in-scope> <out-of-scope-dest>` all sail past the out-of-registry check.
2. **`.shellsecretignore` over-broad default** (`.shellsecretignore:21-43`): whitelists `*.md` at any depth + whole `docs/`,`tests/`,`test/`,`examples/` trees from secret scanning → a real key pasted into a runbook/README commits cleanly. Mechanism is solid (two-commit anti-self-bypass verified); the shipped default breadth is the issue.
3. **`--reason=completed` equals-form false-block** (`pre_tool_use.sh:563`): `tf_completed` matches only `--reason[[:space:]]+completed`; `--reason=completed` (gh-accepted) leaves a legit trusted-filer completed-close blocked → escape-hatch pressure (fail-safe direction, but inconsistent with the matcher's other `[=[:space:]]` selectors).
4. **CRLF parent-marker mis-resolve** (`issue_type.sh:211`): `^Parent Directive: #[0-9]+$` fails on a trailing `\r` → a parented issue can get `--add-label task`/`bug` (consistency bypass).

## Why

MISSION "the escape hatch stays narrow" + "rather than silent regressions." These are lower-severity but each is either a real out-of-registry mutation path, a credential-leak surface, or a consistency/usability miss that pushes toward `SKIP_HOOKS`. The Directive's signal 6 requires each be fixed or explicitly documented as a residual. (Directive #498 cluster 7.)

## Acceptance criteria

- [x] Destructive: extend the dest check to flagless `mv`/`cp` (≥2 operands → check final operand); decide (fix or documented residual) on `find -delete`/`-exec rm`, `truncate`, and `>`/`>>` redirection to out-of-scope paths.
[...]

PR #512, opened 2026-06-29
feat(#502): SessionStart registry-zeroed detector + binding residual doc

### LINKED ISSUE #502: fix(#498): guard or document the two silent session-wide enforcement-disable surfaces
Parent Directive: #498

## What

Two surfaces silently disable enforcement **session-wide with no audit trail** — qualitatively worse than the documented fail-open surfaces (which all leave a warn/skip record):
1. **Mid-session registry zeroing**: `: > .claude/eng-state/registry.txt` (truncate; not an `rm`/`mv`/`cp` verb) or `rm .claude/eng-state/registry.txt` (no force/recursive flag → below the destructive prefilter `pre_tool_use.sh:191`) → empty registry → `cwd_guard.sh:26` `return 1` → `pre_tool_use.sh:36` `in_scope || exit 0` → the **entire hook exits**, with no audit record. Subsequent secret / protected-branch / force-push / sensitive-file are all off.
2. **Shell-root binding repoint**: when `CLAUDE_ENG_SHELL_ROOT` is unset the hook self-locates *through* the agent-writable `.claude/eng-shell-root` symlink with no integrity check; repointing it at a neutered tree disables all enforcement next session, and the SessionStart banner — invoked *through* the binding — cannot detect it (traceless).

## Why

MISSION "the escape hatch stays narrow — audit-log queries show bypasses cluster in legitimate cases" and "hard blockers surface as audit-logged states rather than silent regressions." Both fail-opens are by-design (SPEC §3.2.2/§773), but their **reachability by an innocuous, un-gated command** plus the **absence of any audit trail** turns a transparency posture into a traceless session-wide kill-switch. The bar this EI must hold: no enforcement-disabling action stays both innocuous-command-reachable AND traceless. (Directive #498 cluster 4.)

## Acceptance criteria

- [x] **Decision first (SPEC §1.4):** decide per surface whether to add a guard or accept it as a documented residual — recorded in this EI's PR Plan with rationale, since flipping a documented fail-open is constraint-gated. (No code change presumed before the decision.)
- [x] If guarded — registry: the per-project registry path is protected from Edit/Write/truncate/`rm`/`mv` (mirror the sensitive-file basename arm), and/or armed-state is cached at SessionStart so a mid-session disappearance fails-closed or emits a loud, path-independent audit/warn (not silent).
- [x] If guarded — binding: a tamper-evident assertion on the resolved shell-root before sourcing, and/or an out-of-band SessionStart detector that does not run *through* the binding.
- [x] If accepted as residual — SPEC §6.1/§6.5(c) explicitly documents each as the known un-audited disable path, with the guardrail-not-sandbox rationale.
- [x] Whichever path: no longer is there an enforcement-disable that is both single-innocuous-command-reachable and produces zero audit signal. Smoke lock or doc-presence lock as appropriate; full smoke green.

## Out of scope

- Turning hooks into a sandbox (Directive #498 Non-goal 1) — this is tamper-*evidence*, not prevention of a determined `bash -c`.
- The `safe_source` helper-miss fail-open (tracked as a separate consideration; it at least warns).

## Notes

- Source: adversarial audit of #498, surface 2 (H1 registry, re-verified in code) + surface 4 (binding repoint). Anchors: `pre_tool_use.sh:36/191`, `cwd_guard.sh:14/26`, `session_start.sh`, `scripts/lib/inject.sh`.

ilgyu-yi:
Closes #502

## Goal
Make the two silent session-wide enforcement-disable surfaces from the #498 audit no longer **both innocuous-command-reachable AND traceless**. EI-4 of Directive #498 — the **decide-first (SPEC §1.4)** cluster.

## Decision (SPEC §1.4, recorded before code)
- **Registry zeroing → GUARD by DETECTION.** A present-but-empty per-project registry silently disables all enforcement (the global `in_scope` gate fails open on an empty registry — by design). `session_start.sh` now detects the **present-but-EMPTY** state (the "armed-then-disarmed" signal; an *absent* registry is the normal transparent unregistered case and does NOT fire) and emits a once-per-session banner + `audit_log warn registry-zeroed`. **This is NOT a fail-open→fail-closed flip** (Directive #498 Constraint 1): the empty-registry transparency is the deliberate guardrail-not-sandbox posture (§6.1/§773) and is unchanged — the detector only makes the disarmed state **observable**.
- **Binding-symlink repoint → DOCUMENT as accepted residual.** Repointing `.claude/eng-shell-root` at a neutered tree disables the session and is **structurally undetectable from within** (this detector, like the whole SessionStart hook, is reached *through* the binding). The launch-time `bin/claude-eng` check (§3.5) catches a missing/dangling binding but not a repoint-to-valid-but-neutered tree. Recorded in SPEC §6.5(c) as the documented "anyone who can repoint the binding can bash around the hook" residual.

## Plan (`feat`, reproduce-first)
`session_start.sh`: after the helper sources, detect the empty registry → banner (FIRST, the user-visible AC5 signal) + a **subshell-guarded** `audit_log` (so any audit misbehavior under `set -u` can never itself break the hook — this was caught during testing). SPEC §6.5(c) documents the detector + the binding residual + the §1.4 decision. TOC regenerated for the line shift.

## Commits (Test → Code+Doc)
- `855381a` **test**: §502a-d.
- `58f7dd6` **feat**: the detector + SPEC §6.5(c) doc + TOC regen.

## Key context
- `.claude/hooks/session_start.sh` — the registry-zeroed detector (banner-first, subshell-guarded audit).
- `SPEC.md` §6.5(c) — detector + binding-repoint residual + §1.4 decision; TOC regenerated.
- `scripts/test/smoke.sh` §502a-d (mirrors the §40 banner harness).

## Out of scope
- Real-time BLOCKING of the registry-zeroing write (the `: >`/`truncate`/redirect vector is the documented #505 destructive residual) — detection is at the next SessionStart, not a real-time block. Fully blocking every shell vector contradicts the guardrail-not-sandbox model.
- Closing the binding-repoint (structurally undetectable from within — documented residual).

## Risks
The detector must never itself break session_start — guarded by emitting the banner first + running `audit_log` in a subshell (`( … ) || true`), after a test run showed `audit_log` could `set -u`-abort on an unusual env. No fail-open flip → Constraint 1's §1.4 gate is satisfied by the recorded decision, not bypassed.

## Test plan
- [x] `bash scripts/test/smoke.sh` → pass=911 fail=0; §502a-d GREEN; §40 banner + spec-toc/§90c/§101c/§102a intact.

## Ship gate
- [x] All tests pass — pass=912 fail=0
- [x] security-reviewer passed — VERDICT: ship (after the audit-arity refine: persistent record now writes, §502e locks it; no fail-open weakening; residual honest)
- [x] Docs in sync — SPEC §6.5(c) + TOC regenerated
- [x] PR body curated
- [ ] CI green — pending after ready

## Review response
security-review caught a real Medium: the `audit_log warn registry-zeroed "…"` call passed 3 args but the signature is `<event> <category> <decision> <reason>`, so `$4` was unbound → `set -u` abort (swallowed by the subshell guard) → the persistent audit record never wrote (only the banner fired). Fixed in `4b53d10` by adding the `notice` decision arg + a §502e lock that greps the audit aggregate for the record. Re-verified: ship.


ilgyu-yi:
## AC closeout

All ACs of #502 satisfied (855381a test → 58f7dd6 feat → 4b53d10 audit-arg fix → changelog). DECISION (SPEC §1.4): registry-zeroing GUARDED by detection (banner + persistent audit record, §502a/e), NOT a fail-open flip (empty-registry transparency unchanged); binding-repoint DOCUMENTED as accepted residual (structurally undetectable from within, SPEC §6.5(c), §502d). No enforcement-disable remains both innocuous-reachable AND traceless. smoke pass=912 fail=0. security-reviewer: ship (re-verified after the audit-arity refine).

Issue #550, opened 2026-07-06
directive: remediate whole-project review-audit findings (8 clusters)


ilgyu-yi:
## Objective
Remediate the findings from the full-strength whole-project review audit — 35 adversarially-verified findings, re-validated against current `main` (`a70ab1b`) — as **8 file-coherent Execution Issues** (clusters A–H), each landing as a **merged PR** that preserves or **tightens** the guardrail safety posture (never a fail-closed→fail-open regression). Concrete artifact-level boundary: **8 Execution Issues** filed under this Directive (labels `execution`, parented via `Parent Directive: #<this>`), **all closed by merged PRs to `main`**, with `scripts/test/smoke.sh` green on every PR. Clusters (file-ownership grouped to keep PRs conflict-free):
- **A** — `pre_tool_use.sh` + `branch_guard.sh` matcher gaps: unnormalized-path carve-out bypass (`:1157/:1170`), `mv/cp --` evasion (`:192-193`), bundled `-uf` bare-force skip (`:950/:962`), reset/clean message-elision false-block (`:210/:223`), env-operand skip (`:89`), trusted-filer dot-segment (`:638/:664`), case-sensitive `is_protected_branch` (`branch_guard.sh:78`).
- **B** — `secret_scan.sh` missing GitHub token families `ghs_`/`ghu_`/`ghr_` **[MEDIUM]** (`:41-43`) + `ghjig_skip.sh` category validation.
- **C** — `scripts/ac_closeout.sh` `@mention` neutralization (`:59-74`) + idempotency guard matching the gate's strict `(resolved by PR #N)` shape (`:42` vs `ac_closeout_gate.sh:216/:220`) to fix the merge-deadlock (worsened by #500).
- **D** — shell robustness: `stop.sh` throttle sanitize, `conventional_commit.sh:231` locale, `hookrt.sh:137` bare `$GHJIG_ROOT`, `session_start.sh` banner legacy-registry blind spot.
- **E** — CI robustness: mirror Parent grep CRLF/`$`-anchor **[MEDIUM]** (`:118`), mirror `gh view` fail-soft, `check-changelog.yml` diff-fail conflation + line-independent fragment grep, `detect_bare_refs_directive.sh` 5-cap.
- **F** — agent contracts (#285): `activation-reviewer.md:199/:202` N=3 escalation ownership, `planner.md:4` working-tree-discipline block + `CLAUDE.md` isolation-list entry.
[...]

PR #561, opened 2026-07-06
fix(#555): close 7 matcher bypass/false-block gaps (cluster A)

### LINKED ISSUE #555: fix pre_tool_use.sh + branch_guard.sh matcher gaps (bypass/false-block)
Parent Directive: #550

## What
7 verified matcher defects, all preserving or TIGHTENING enforcement (no fail-closed→fail-open):
- A1 pre_tool_use.sh:1157 (shell_self_mod) & :1170-1172 (user_global_memory) — carve-outs prefix-match the RAW $target; a ..-laden absolute path bypasses the normalizing path_in_scope gate (:1180). Fix: physically resolve $target (ancestor-walk + pwd -P) before the case prefix matches.
- A2 :210 (reset) & :223 (clean) — arms grep the un-elided command, false-blocking a commit whose -m MESSAGE mentions the command; force-push (:962)/protected-push (:1041) already elide via strip_command_data ... message. Fix: feed the same message-stripped text.
- A3 :192-193 — mv/cp -- <src> <out-of-registry> evades both destructive arms. Fix: skip an optional leading -- before the operand test.
- A4 :950/:962 — bundled short force git push -uf/-fu has no isolated -f\b token → bare bundled force skips the fail-safe block. Fix: match -[A-Za-z]*f[A-Za-z]*\b.
- A5 :89 — check_destructive_args continues on any token equal to rm|mv|cp|sudo|doas|time|env regardless of position → an operand named env skips path_in_scope. Fix: only skip in command position.
- A6 :638-650/:664-677 — trusted-filer URL/--repo parse validates segment COUNT only, not ./.. segments. Fix: reject dot-segments + enforce owner/repo charset.
- A7 branch_guard.sh:78 — is_protected_branch case-sensitive while force-push arm uses grep -qiE → Main/MASTER unrecognized on a case-insensitive FS. Fix: add -i for parity.

## Why
Serves remediation Directive #550 (MISSION.md:18 enforcement doctrine / :16 high-asymmetry safety redundancy) — this cluster closes verified gaps in the shell's own guardrail layer. Files: pre_tool_use.sh, branch_guard.sh, scripts/test/smoke.sh, SPEC.md/CLAUDE.md (if contract changes).

## Acceptance criteria
- [ ] A1–A7 each fixed with the matcher preserving/tightening its block-path (no coverage lost).
- [ ] smoke assertions added that FAIL before / PASS after for: mv -- out-of-registry (blocked), git push -uf bare force (blocked), operand named env (path checked), ..-path carve-out (blocked), reset/clean -m message no longer false-blocks, case-insensitive protected-branch recognized.
- [ ] scripts/test/smoke.sh green.
- [ ] Any documented-contract change reflected in SPEC §6.1 / CLAUDE.md same PR.

## Out of scope
- Other clusters' findings (each is its own Execution Issue under #550).
- The 6 positively-verified non-defects and the 2 design-questions (/discuss-tier) — per #550 non-goals.
- Any fail-closed→fail-open regression; any lossy/rephrasing-only edit; new features.

## Notes
- Parent: Directive #550. Cluster A. Priority P1. Findings re-validated against main a70ab1b.

ilgyu-yi:
Closes #555

## Goal
Cluster A of Directive #550: close seven verified matcher defects in the shell's own guardrail layer (`pre_tool_use.sh`, `branch_guard.sh`). Every fix **preserves or TIGHTENS** its block-path — no fail-closed→fail-open, no dropped block (SPEC §6.0 P1: negative gates made *more precise*). fix/reproduce-first, so Doc→Test→Code is relaxed; commit graph is docs → test → fix anyway.

## Rework (parked-PR #561 → re-pushed)
The prior push was parked: CI smoke failed on assertion **555a1** (`$HOME/.claude` carve-out bypass not blocked) on **both** runners while passing locally. Root cause: the first A1 fix resolved `$target` **physically** (`[ -d "$_rt_anc" ] && cd … && pwd -P`), which only collapses `..` when the ancestor directory **exists**. `~/.claude` exists on the dev box (local pass) but is **absent** on CI runners, so the resolve fell back to the raw `..`-path and the carve-out matched → bypass re-opened → CI fail. **Fix now normalizes `..`/`.` LEXICALLY** (a pure-string segment collapse, no `pwd -P`/`realpath`/`[ -d ]`/filesystem access) — a security matcher must neutralize a `..`-escape even when the target (or its carve-out ancestor) does not exist, since an Edit/Write routinely creates a not-yet-existing path. A2–A7 are unchanged from the prior push (they passed). A new smoke pin (**555a1-home-absent**, `HOME` → a temp dir with no `.claude`) reproduces the exact CI condition and now blocks.

## What changed (per finding)
- **A1** `pre_tool_use.sh` Edit/Write carve-outs: **lexically normalize** `$target` (make absolute vs cwd, then a pure-string `.`/`..` segment collapse — no filesystem access) **before** the `$SHELL_ROOT`/`$HOME/.claude` prefix match, so a `..`-laden absolute path (`$GHJIG_ROOT/../../etc/passwd`) can no longer borrow a carve-out to skip the out-of-scope gate — and the neutralization holds even when the carve-out ancestor is absent (CI condition).
- **A2** reset/clean arms: use `strip_command_data … message` so a `git reset --hard`/`git clean -f` named only inside a `-m`/`-F` commit **message** is data, not a command, and no longer false-blocks; a real invocation still blocks (the strip never removes a genuine verb).
- **A3** flagless `mv`/`cp` entry arm: skip an optional leading `--` end-of-options marker before the operand test, so `mv -- <src> <out-of-registry>` still enters `check_destructive_args` instead of evading both arms.
- **A4** force-push arm: match a **bundled short cluster** (`-uf`/`-fu`) via a token-start-anchored `(^|space)-[A-Za-z]*f[A-Za-z]*` pattern (entry pre-filter + per-segment re-confirm), so a bare bundled force hits the fail-safe; the token-start anchor keeps a branch name containing `-f` (`my-feature`) from being misread (no over-block).
- **A5** `check_destructive_args`: skip the verb/wrapper words (`rm`/`mv`/`cp`/`sudo`/`doas`/`time`/`env`) **only in command position**, so an operand literally named `env`/`time`/etc. is scope-checked instead of skipped.
- **A6** trusted-filer repo parse: validate the resolved `owner/name` (URL- or flag-derived) for `.`/`..` path segments + the GitHub charset before the repo-aware trust/discussion lookups; a rejected value falls back to the current repo (fail-soft, tightening).
- **A7** `branch_guard.sh` `is_protected_branch`: case-insensitive name match (`grep -qiE`), parity with the force-push arm, so a `Main`/`MASTER` checkout on a case-insensitive filesystem is recognized as protected.

## Test
Added FAIL-before / PASS-after smoke assertions (`scripts/test/smoke.sh`) for A1 (both carve-outs + **home-absent CI-condition pin** + no-over-block), A2 (message false-block + real-block regression), A3 (`--` entry + in-scope no-over-block), A4 (`-uf`/`-fu` bare force + branch-name guard), A5 (`env` symlink out-of-scope + wrapper no-over-block), and A7 (`Main`/`MASTER` + unprotected guard). Full suite: **pass=1007 fail=0**.

## Safety
Every change is a block-path tightening or a false-block removal that cannot under-block: A1/A3/A4/A5/A6/A7 close bypasses (more blocks); A2 removes false-blocks via message-value elision that never strips a real command verb (fail-closed to unstripped on parse failure).

## SSOT
SPEC §6.1 matcher rows annotated for A1–A7 (A1 row updated to lexical normalization; no heading change → no TOC rebuild). CLAUDE.md's pointer-level rows stay accurate (no contract-level change). Changelog: `security/555.md`.


ilgyu-yi:
Parked — CI not green after review-approve. Human check needed.

Issue #634, opened 2026-07-27
Decide the enforcement face for review-body egress: the self-review wrapper posts body content verbatim to a public PR comment


ilgyu-yi:
## What

After #633 collapses the self-review producer into the single bare `scripts/ghjig_file_review_post.sh`, that wrapper posts the review body **verbatim** — apart from a marker shape/head bind — to a **public, unretractable** PR review comment, and the first link in the chain is a `Write` to `.claude/ghjig-state/file-review/staging` that is **not** in `permissions.allow` and is adjudicated by the harness per call.

#633 retires the arbitrary-**path**-read entry point (no shell script in the tree can be pointed at an arbitrary file by an allow-covered invocation). It deliberately does **not** address arbitrary-**content** egress, and its Out-of-scope section says so explicitly: *"No sanitization or safety validation of body content is added to the wrapper"*, with the marker parse carved out as *"a staleness bind on a shell-contracted field, not content sanitization."* Adding a content guard there would also breach #633's producer-only invariant (SPEC §5.29 — `/file-review` adds, changes, and removes no gate).

So the residual is real, and #633 changes **what stands between it and a publish**. Before #633 the chain halts at a step that *cannot* be allow-covered wildcard-free — a structural block, independent of any per-call judgment. After #633 the only remaining checkpoint is the `Write` step's per-call harness adjudication, which the shell neither controls nor records. That is not "an unattended run will publish"; it is that the guarantee changes from *structurally blocked* to *dependent on an adjudication outside the shell's control and outside its audit trail*. This Issue exists so that residual is **tracked and decided**, not narrated once in a SPEC sentence and forgotten.

The decision to make is the SPEC §6.0 **enforcement-face** call:

- **(a) Hard negative gate** — a fail-closed egress secret-pattern scan of the body inside the wrapper *before* the POST, reusing the existing `SECRET_PATTERNS` from `.claude/hooks/helpers/secret_scan.sh` (no new pattern list, so §9 stays clean). Failure message names the **pattern ID only**, never the matched text. Recovery: redact in `/file-review` step 6 — the sibling of the existing `@mention`-neutralization step.
- **(b) Positive guidance only** — an explicit redaction obligation in `/file-review` step 6 with no wrapper-side gate, on the §6.0 P3 advisory-first principle.
[...]

Issue #610, opened 2026-07-16
GHES: `gh api user` resolves the default host, not the repo host, so review posting's own-PR guard mismatches (+ file-review URL validation hardcodes github.com)


ilgyu-yi:
## What
On a GitHub Enterprise Server (GHES) target, the review path fails because `gh api` calls resolve gh's DEFAULT host (github.com) rather than the repo's GHES host — `gh api` does not infer the host from the current repo (`--hostname` defaults to `github.com`). This breaks two things, both host-resolution defects:

**(1) The merge-review gate can't read the review, and the review post targets the wrong host.**
- Review FETCH — `ac_closeout_gate.sh:409` runs `gh api "repos/$nwo/pulls/$pr/reviews"`. On GHES this queries github.com for a repo that lives on the enterprise host → 404/empty → the merge-review gate can't see the head-pinned review → fails closed → **merge blocked** ("merge-review hook can't read the review on GHES").
- Review POST — `ghjig_file_review_post.sh:115` runs `gh api "repos/$owner_repo/pulls/$pr_num/reviews"` against the default host too, so posting the head-pinned review targets the wrong host on GHES.

**(2) The acting-identity lookup mismatches, and at one site that is a self-merge bypass (not just a false-negative).**
The own-PR / merger checks compare `me = gh api user` (no repo context → default host) against `pr_author = gh pr view --json author` (repo-scoped → GHES host). On a machine authed to both github.com and a GHES host the logins differ. Most sites then fail CLOSED (a false-negative), but `merge_is_self` (`ac_closeout_gate.sh:108`) returning 1 ("author != merger") is admitted by `pre_tool_use.sh:536` as a "human merge" — so a wrong-host identity there is a covered-form **self-merge admission**, the system's single fail-OPEN site.

Root cause identified by code inspection (`gh api --help` confirms `--hostname` defaults to github.com; a runtime GHES repro needs the user's multi-host environment). Affected `gh api` sites: `ghjig_file_review_post.sh:50` (identity) and `:115` (review POST); `ac_closeout_gate.sh:108`/`:450` (identity) and `:409` (review FETCH); `.claude/commands/file-review.md:10` (github.com URL validation) and `:20` (identity idiom); `.claude/commands/work-on.md:12` (branch-name identity idiom).

## Why now
[...]

Issue #479, opened 2026-06-25
fix: restore a working in-harness escape channel (both SKIP_HOOKS forms are stripped)


ilgyu-yi:
## What
The enforcement escape hatch has **no working form inside the Claude Code Bash tool**: the leading `SKIP_HOOKS=` env-prefix is consumed by the harness as the spawned subprocess's environment, and the trailing `# claude-eng:skip=… reason=…` sentinel is stripped before the command reaches the PreToolUse hook's `tool_input.command`. The parsers (`parse_skip_sentinel`/`parse_env_prefix`) are verified correct in isolation, so this is a command-delivery gap. (#478 corrected the docs to state this reality + an interim real-terminal recourse; THIS issue restores a working in-agent channel.)

A shell-controllable delivery is needed so a sanctioned protected-branch operation (`/release` commit on `release/X.Y.Z`; the §5.0 bootstrap Stage-0 seed) can be escaped from *within* the agent, not only from a separate terminal.

## Why
- **MISSION fit** — SPEC §6.0/§7: the escape hatch is the audited pressure-relief valve of the enforcement model ("always open but audited"; MISSION "the escape hatch stays narrow"). With no in-harness form, the valve is effectively closed in-agent — breaking the in-agent `/release` and `/bootstrap-repo` flows the shell is built around.
- **Why now** — surfaced concretely by the blocked 0.5.0 release commit; #478 documents the gap, this closes it.
- **Coverage** — #478 is docs-only (corrects the false claim, interim workaround); no open issue restores the mechanism.

## Acceptance criteria
- [ ] A shell-controllable escape channel that **survives the Claude Code Bash tool** reaches the PreToolUse hook and is honored by `should_skip` — a file-based skip token under `.claude/state/` that the hook reads (NOT an in-command prefix/comment the harness strips), or an equivalent out-of-command affordance.
- [ ] A sanctioned protected-branch op (the `/release` release-branch commit) succeeds in-harness via the new channel; smoke covers the happy path AND asserts the channel does NOT leave a standing bypass.
- [ ] **Audit + narrowness preserved**: every use is `audit_log`-recorded (parity with today's SKIP_HOOKS audit), and the channel is one-shot / category-scoped — no persistent or un-scoped bypass.
[...]

Issue #544, opened 2026-07-03
worktree-isolated reviewers can review the wrong artifact (base, not PR head) → false-assurance


ilgyu-yi:
## What
Worktree-isolated reviewer subagents (`code-reviewer`, `security-reviewer`, `plan-reviewer`, etc.) can review the **wrong artifact** and still emit `APPROVE`. When dispatched with worktree isolation during `/ship` and `/review`, the worktree is created at the repo's base/default branch (`origin/main`), **not** at the feature-branch HEAD under review.

Observed live in PR #543 (#539 ship): 3 of 4 ship reviewers' worktrees were at base `3325fc2` while the PR head was `c4ed6c3`; the #539 diff was entirely absent from their worktrees. One security reviewer returned "APPROVE — doc-only diff, #539 changes ABSENT" — it certified code it never saw. The N=3 gate only survived because the other reviewers happened to `git fetch` the PR ref on their own initiative; nothing in the dispatch guarantees that.

## Why
MISSION — enforcement / positive-and-negative dual (§6.0) and unattended operating-mode coupling (§1.5): the reviewer gate is the highest-value code-quality-assurance device and, in unattended mode, the **substitute for human review**. A reviewer that silently reviews the wrong artifact and still approves is **worse than no reviewer** — it manufactures false assurance and can let unreviewed code merge, defeating the whole review apparatus.

Why now: surfaced concretely in #543; the exposure is present on every worktree-isolated post-push review, and the fix is cheap.

## Acceptance criteria
- [ ] `/ship` and `/review` reviewer dispatch pins the reviewed artifact to the actual PR head — via one of: (a) reviewers review a specific pushed ref (`gh pr diff <n>` / `git fetch origin <branch>`) as the authoritative source; (b) the caller verifies each worktree's HEAD == PR head before the verdict is trusted; or (c) post-push review shares the correct (clean) working tree instead of an isolated base worktree.
- [ ] A reviewer whose reviewed artifact does not match the PR head is treated as an **invalid vote (fail-closed)**, never counted as an APPROVE.
- [ ] Smoke/verification demonstrates a reviewer dispatched on a feature-branch PR sees the feature diff, not the base.
- [ ] SPEC updated where the reviewer-dispatch + working-tree-isolation contract lives (§1.5, §4.x reviewers, §5.7 ship).

## Out of scope
- Changing the reviewers' rubrics or verdict vocabulary.
[...]

PR #635, opened 2026-07-28
fix(#633): single-wrapper self-review producer — delete the stage writer, retain the head guard

### LINKED ISSUE #633: Unattended /ship parks at the self-review stage step: the writer's argv is not allow-coverable, and with it gone the stage/post split has no rationale left
## What

The `/file-review` self-review producer is a three-step chain (`.claude/commands/file-review.md` steps 6-7):

```
sanitize -> <mktemp tempfile>
.claude/ghjig-root/scripts/ghjig_file_review_stage.sh <tempfile>   <- NOT allow-covered
.claude/ghjig-root/scripts/ghjig_file_review_post.sh                <- allow-covered (#598)
```

#598/#602 removed the auto-mode-classifier block at the **post** step, but the **stage** step is still an uncovered `Bash` invocation of an opaque local script **carrying a variable argv**. Two defects follow:

1. **The classifier gap recursed one step.** SPEC §5.7.1's own argument for the paired producer exception ("you cannot merge without the review it cannot post") now applies one step earlier: an unattended `/ship` can park at `stage`, not `post`. `.claude/settings.json:5` and `.claude/settings.injected.json:5` cover only `ghjig_file_review_post.sh`, so the documented end-to-end path provably parks on its first unattended run. SPEC is **silent** on the stage step's coverage — §5.7.1 (L1438), §5.29 (L1724) and the tree entry (L2295) describe the writer's existence but never its own invocation coverage. A gap in the contract, not a documented decision.

2. **No wildcard-free allow entry is possible, and the wildcard form opens an irreversible wrong-allow.** The argv is a variable tempfile path, so the only matching form is the prefix wildcard `Bash(.claude/ghjig-root/scripts/ghjig_file_review_stage.sh:*)`. Under this shell's own documented premise (SPEC §5.7.1; `scripts/ghjig_file_review_stage.sh:6`) the classifier defers only on a **byte-exact, wildcard-free** match — so the wildcard buys nothing in the auto mode where the gap lives, while breaking the no-wildcard-allow discipline §148b/§148c/§144e lock for its sibling entries. Worse, it would auto-approve an **arbitrary-path-read** primitive: `ghjig_file_review_stage.sh:25-28` gates only on readable/non-empty, and `ghjig_file_review_post.sh:126-140` posts the staged body **verbatim, with zero content validation**. Pointing the writer at a private key and then invoking the bare wrapper publishes that key as a PR review comment. The `sensitive-file` matcher covers only Edit/Write; the out-of-scope destructive matcher covers only the file-removal and move/copy command forms. Today the visible permission prompt (which shows the path) is the only checkpoint — a `:*` entry removes exactly that checkpoint.

**The fix is not "make the writer bare" — it is to collapse the split.** Once the argv is gone the writer has no job left: `ghjig_file_review_stage.sh` exists **only** to translate a variable path into the fixed one the bare wrapper reads. Its two remaining outputs are handled without it:

- `created=` is replaceable by the staging file's **mtime** — the agent is now the writer, so mtime *is* write time.
- `head=` must be **retained, not subsumed** — and the single-step design can bind it more directly. The deleted stamp compared `git rev-parse HEAD` (`stage.sh:38`, **local**) against `headRefOid` (`post.sh:40-43`, **remote**), so it spanned the whole review, not the split's few seconds: a foreign push (concurrent agent, human push, GitHub "Update branch") that advances the PR head after `/file-review` step 1 would otherwise let the wrapper POST `commit_id=<new head>` for a body reviewed at the old head, and since `merge-review` reads `commit_id` off the review object and only `verdict` from the marker (SPEC §5.7.1), that defeats "never merge an unreviewed head" (SPEC L1359). Today `post.sh:121-123` fails closed on it, locked by smoke §148j-headmiss. **Replacement without a stamp:** `/file-review` step 1 (`file-review.md:18`) computes `HEAD_SHA` from `gh pr view --json headRefOid` — the *remote* head — and the body marker (`file-review.md:47`) carries it as `head=`. So the wrapper can compare **the marker's `head=` against the `headRefOid` it resolves at POST time**: the head the review was performed against versus the head the POST pins, directly, across the whole span. A shell-authored belt is added alongside — the wrapper's own `git rev-parse HEAD` must also equal `headRefOid`, which is exactly what the deleted stamp checked. Both fail closed. Like the stamp it replaces, the marker parse is a **staleness** guard, not anti-forge (the marker is agent-authored; anti-forge stays at the `merge-review` gate, which binds `author.login` / `commit_id` off the review object — unchanged from #602's own disclaimer).

Target shape — one script, no stamps, the head guard retained:

```
agent:  sanitize -> Write .claude/ghjig-state/file-review/staging
bare    .claude/ghjig-root/scripts/ghjig_file_review_post.sh
        resolve PR/headRefOid/author (gh pr view, gh repo view --json url)
        own-PR guard (gh api user --hostname $h == pr_author)
        readable + non-empty staging?           else fail-closed, no POST
        mt=<portable mtime>                     else fail-closed, no POST
        slurp body, then unlink staging          TOCTOU foreclosed
        mt <= now  and  now-mt <= 60            else fail-closed, no POST
        marker head= present and == headRefOid  else fail-closed, no POST
        git rev-parse HEAD == headRefOid        else fail-closed, no POST
        POST event=COMMENT commit_id=$headRefOid --field body=@-
```

The `permissions.allow` surface is **unchanged** — the existing exact wildcard-free `Bash(.claude/ghjig-root/scripts/ghjig_file_review_post.sh)`. Net effect: one script and one prospective allow entry **removed**, no shell script in the tree can be pointed at an arbitrary path, and the fail-closed set is **larger** than today's (an absent or malformed marker now blocks a POST that currently succeeds).

**Residual, stated honestly.** This retires the arbitrary-**path**-read entry point. It does **not** retire arbitrary-**content** posting: the wrapper still posts the body verbatim apart from the marker bind, and the first link in the chain becomes a `Write`, which is **not** in `permissions.allow` and is adjudicated by the harness per call. The checkpoint therefore **moves** rather than disappearing. What is durably gained is that no `:*` allow entry ever auto-approves an arbitrary-path read, and the reachable surface shrinks from "any readable file" to "one fixed path the caller must write itself." SPEC must record this residual rather than claim removal.

## Why

MISSION, "Enforcement is the action-side counterpart of this dual" (`MISSION.md:18`) — specifically its cost-asymmetry rule, which the two defects sit on opposite sides of:

- *Positive* enforcement is one step short of the path it exists to lay: a `permissions.allow` exception that "supplies the right next action" strands the actor one command before the goal — the same one-sided regression as "a hard block without naming the right alternative."
- The only allow form that would close it is a **wrong-allow whose cost is irreversible** (a leaked secret, published) — exactly the class MISSION says must stay a hard, narrow, fail-safe-toward-blocking gate rather than a generously laid positive rule. So the resolution cannot be "allow it more broadly"; it must be "narrow the surface until the exact form suffices."

MISSION's "narrowed again if it begins to fight the prescribed flow … improves under use" clause is the migration this performs. Secondarily — by analogy, not by the letter of the §1.9 harness-absorption doctrine — keeping `stage.sh` after the constraint that motivated it is removed would be the same dead ceremony §1.9 warns against retaining.

**Why now**: the producer path has **never executed** — 0 `file-review` records in the audit log — so the change is free today. Once an unattended `/ship` first runs it, the failure surfaces as a park at `stage`, whose obvious-looking fix is the `:*` entry; the exfiltration surface then becomes load-bearing and its removal a breaking change instead of a no-op. #602 (`364bb1e`, 2026-07-14) is still warm and single-sited; deferring means migrating live bound targets through `settings.injected.json` propagation later.

**Existing coverage**: #633 is the only open Issue and there is no open PR; #598/#602/#610 are closed/merged, so a comment there is not actionable. `git log -15` shows nothing on this path since #602/#610.

## Acceptance criteria

- [ ] **SPEC** §5.7.1, §5.29 and the tree entry (L2295) record the single-wrapper producer: the sanitized body is written **by the caller** to the fixed `<esd>/file-review/staging`; the separate writer and the `created=`/`head=` **stamps** are removed; the TTL moves to the staging file's mtime.
- [ ] **SPEC records the head guard as retained, never as subsumed.** It states what the deleted stamp actually compared (local `git rev-parse HEAD` vs remote `headRefOid` — spanning the whole review, not the stage-to-post window), that the replacement is the wrapper's own comparison of the body marker's `head=` **and** its own `git rev-parse HEAD` against the resolved `headRefOid`, and that both remain **staleness** guards, not anti-forge (anti-forge stays at `merge-review`). No text may claim the guard is unnecessary under a single step.
- [ ] **SPEC records the residual and the moved checkpoint**: the wrapper posts body content verbatim apart from the marker bind, so what is retired is the arbitrary-**path**-read entry point, not arbitrary-content posting; the first link is now a `Write` that is **not** in `permissions.allow`. No claim that the content channel is closed.
- [ ] `scripts/ghjig_file_review_stage.sh` is **deleted**, and no reference to it survives in `SPEC.md`, `.claude/commands/file-review.md`, `scripts/`, or `scripts/test/` (grep for the basename returns nothing).
- [ ] `scripts/ghjig_file_review_post.sh` still takes **no positional argument**, resolves `<esd>` via the shared `ghjig_state_dir_cli()`, reads `<esd>/file-review/staging`, and preserves the order **mtime-read -> slurp -> unlink -> validate -> POST** (the unlink stays *before* the POST). It fails closed with **no POST** on every one of: absent / unreadable / empty staging; unresolvable mtime; mtime in the **future**; mtime older than 60s; a body whose `file-review` marker is absent or malformed; marker `head=` != resolved `headRefOid`; unresolvable local `git rev-parse HEAD`; local `git rev-parse HEAD` != resolved `headRefOid`. `event=COMMENT` hardcoding, the own-PR author guard, and the GHES host pin (#610) are unchanged.
- [ ] The mtime read is **portable**: a BSD (`stat -f %m`) and a GNU (`stat -c %Y`) branch, matching the repo's established idiom (`session_start.sh:114-116`, `helpers/status.sh:35-37`). A platform with neither must not silently skip the TTL.
- [ ] **Agent-write path == wrapper-resolved path is made checkable.** `.claude/commands/file-review.md` writes the body to the project-dir-relative `.claude/ghjig-state/file-review/staging` and states the condition under which that literal is the path `ghjig_state_dir_cli()` resolves: with no `GHJIG_STATE_DIR_OVERRIDE`, and given that a Bash-tool subprocess **often runs with `CLAUDE_PROJECT_DIR` unset** (`hookrt.sh:95-97`) so the wrapper commonly lands on the `git rev-parse --show-toplevel` branch, agreement holds iff the project dir equals the git top level. The doc names the fallback for any other case (resolve the path by running the helper). A smoke assertion pins that branch: with `CLAUDE_PROJECT_DIR` **unset**, no override, and cwd at the project root, `ghjig_state_dir_cli` == `<project-root>/.claude/ghjig-state`.
- [ ] The wrapper's absent-staging failure message **names the absolute path it looked in**, so a path divergence is a diagnosable error instead of a silent park.
- [ ] `scripts/test/smoke.d/90-ghes-host.sh` — `s610_stage()` (L178) no longer invokes a stage script with an argv; it writes the staging file directly. §153-1 and §153-5 (`posts == 1`) stay **green**.
- [ ] Smoke §148h / §148h-ord / §148h-fc / §148i-a / §148i-b / §148i-cli / §148i-doc / §148j are rewritten for the single-wrapper shape: locks asserting the writer's existence or its argv form are **removed**, while **§148j-headmiss and §148j-future are retained and retargeted** to the new mechanisms (marker/local-head mismatch → no POST; future mtime → no POST) rather than dropped. The retained behavioral round trip drives an agent-written staging file to exactly **one** POST on a fresh, marker-matching body and **no** POST on each fail-closed input enumerated above. §148i-cli's "defined exactly once in hookrt.sh" survives with the wrapper as the only caller. §148i-doc keeps a **positive** anchor (currently the stage-ref at `70-gates-contentlocks.sh:749`) so it cannot green vacuously once that ref is deleted.
- [ ] §148a-§148c / §148d / §148i-reg / §144a-§144e stay **green**: the allow surface remains exactly the wildcard-free `Bash(.claude/ghjig-root/scripts/ghjig_file_review_post.sh)` alongside the merge entry, and no wildcard or raw-`gh api`-reviews entry is added.
- [ ] **The stage-script allow entry is removed and cannot return.** `.claude/settings.injected.json` currently carries `Bash(.claude/ghjig-root/scripts/ghjig_file_review_stage.sh)` while `.claude/settings.json` does not — a target/dogfood parity break, and a **dead** entry either way: the bare form does not match the live `ghjig_file_review_stage.sh <tempfile>` invocation, and under this change the script ceases to exist. This Issue deletes that entry, restores parity between the two allow lists, and adds a smoke lock asserting **neither** file references `ghjig_file_review_stage.sh` in `permissions.allow` in any form (bare or wildcard). No existing lock catches this drift — §148b/§148c/§148i-reg count only the literal `ghjig_file_review_post.sh`.
- [ ] `scripts/test/smoke.sh` is fully green.
- [ ] A `changelog_unreleased/` fragment is present (or `skip-changelog` applied), satisfying the `check-changelog` CI gate (SPEC §18).

- [ ] **AC5 amendment — every reject arm is audited.** Every fail-closed reject arm in `scripts/ghjig_file_review_post.sh` emits an audit record via a `deny()` helper: `audit_log info file-review rejected "reason=<arm>"`. The call is env-prefixed with `CLAUDE_PROJECT_DIR="${esd%/.claude/ghjig-state}"` so `audit_log`'s `ghjig_state_dir` (**not** `_cli`, per `hookrt.sh:158`) resolves the per-project log — that single prefix is exact across all four `ghjig_state_dir_cli` branches, so the production call must **not** also set `GHJIG_STATE_DIR_OVERRIDE`, which SPEC §3.2.1 designates a test-only seam. The call is suffixed `|| true`, because a failing audit must never convert a fail-closed reject into a POST. The reason text carries the **arm name only, never body content**. `SPEC.md:1734`'s `file-review` decision set gains `rejected`.
- [ ] **AC5 amendment — the fail-closed enumeration is extended.** AC5's set additionally includes: (i) a staging body whose canonical `file-review` marker count is **not exactly 1** (0 or >=2), counted with the **byte-identical** regex `.claude/hooks/helpers/ac_closeout_gate.sh:499-502` uses, so producer and consumer accept sets cannot drift; and (ii) a staging path whose **leaf or `file-review` directory component** is a symlink (`[ -f ] && [ ! -L ]` on the leaf is provably insufficient — a symlinked parent directory passes both leaf checks while resolving elsewhere). The `nlink > 1` hardlink case is recorded as a named Doc residual, deliberately unguarded.

## Out of scope

- Merge gates and `resolve_self_review_policy` / `.claude/state/self-review`. `/file-review` stays **producer-only** (#586/#587 own the gates).
- Body **sanitization** remains the caller's (`/file-review` step 6) responsibility. No sanitization or safety validation of body *content* is added to the wrapper — the marker `head=` parse required above is a staleness bind on a shell-contracted field, not content sanitization, and *recording* the residual is Doc-phase work, not a new guard.
- Adding `Write(...)` to `permissions.allow`, or otherwise changing how the harness adjudicates the `Write` step. Recording that the Write is uncovered is in scope; changing its coverage is not.
- `event=COMMENT` hardcoding, the own-PR author guard, and the GHES host pin (#610) — carried over unchanged, not redesigned.
- Making the head guard anti-forge. The marker is agent-authored and stays a staleness guard; strengthening anti-forge is `merge-review`'s domain (#586).

## Notes

- Origin: a review of whether `ghjig_file_review_stage.sh` should simply be added to `permissions.allow`. First conclusion was "make it argv-free, then add an exact entry"; that was collapsed further once it became clear the writer's only remaining job was the path translation the argv removal eliminates.
- The arbitrary-path-read primitive is bounded (own PR only, current branch only, `event=COMMENT` only, 60s TTL) but the **content is arbitrary and unsanitized**, which is the part that matters for a secret.
- `activation-reviewer` round 1 (`revise`) supplied three gaps now covered: the `90-ghes-host.sh:178` argv harness, the four-branch `ghjig_state_dir_cli()` path-agreement risk, and the overstated "surface removed" claim. Round 2 (`revise`) rejected the "head-bind is subsumed" claim by showing the stamp compared local vs remote head across the whole review span; the head guard is now retained and bound more directly, and the future-mtime arm plus the §148i-doc anti-vacuity anchor were added from the same review. Both rounds' hook check passed — the prescribed Write under `.claude/ghjig-state/` is blocked in neither the dogfood repo nor an onboarded target.
- `CHANGELOG.md:47-48` mentions the writer as append-only history and is deliberately left untouched.
- Related: #598 (wrapper + classifier exception), #602 (`364bb1e`, out-of-band staging — this Issue reduces it), #610 (GHES host pin), #591 (`settings.injected.json` propagation model), #587 (merge-entry exception), #586 (merge-review gate).

ilgyu-yi:
Closes #633

## Goal
Collapse the `/file-review` self-review producer from a three-step chain to two links. `/file-review` writes the sanitized body itself to a fixed staging path; the single, already-allow-covered bare `ghjig_file_review_post.sh` validates and posts it. The stage writer — whose only remaining job was translating a variable tempfile path into the fixed path the bare wrapper reads — is deleted, so no command on the producer path needs an allow entry that cannot be written wildcard-free, and an unattended `/ship` no longer parks at the stage step.

MISSION "Enforcement is the action-side counterpart of this dual" (`MISSION.md:18`), specifically its cost-asymmetry rule. The two defects sit on opposite sides of it: the positive-enforcement path stops one command short of the goal it exists to lay, and the only allow form that would close it (a prefix wildcard on the writer) is a wrong-allow whose cost is irreversible — it auto-approves an arbitrary-path read whose content is then published verbatim as a public PR review. The resolution MISSION prescribes is not "allow it more broadly" but "narrow the surface until the exact form suffices". The "narrowed again if it begins to fight the prescribed flow … improves under use" clause is the migration this performs.

## Plan (reviewed merge of A + 2 challengers; plan-reviewer: refine→merge; activation-reviewer round 3: pass)
- **Delete the stage writer.** No reference to the basename survives in `SPEC.md`, `.claude/commands/file-review.md`, `scripts/`, or `scripts/test/`.
- **`ghjig_file_review_post.sh` stays bare** (no positional arg), resolves `<esd>` via the shared `ghjig_state_dir_cli()`, and reads `<esd>/file-review/staging` (leaf renamed from the stamped file). Order preserved and tightened: **stat-mtime → slurp → re-stat → one-shot unlink → validate → POST** (the unlink stays *before* the POST; the re-stat equality makes the freshness check certify the bytes actually posted).
- **Fail-closed, no POST, on every one of:** absent / irregular / unreadable / empty staging; a symlinked staging **leaf** *or* `file-review` **directory component**; unresolvable or malformed mtime; **future-dated** mtime; mtime older than 60s; an mtime that differs between the pre-slurp and post-slurp `stat`; a marker count other than **exactly 1** under the **byte-identical** regex `ac_closeout_gate.sh` uses; an unreadable or mismatched marker `head=`; unresolvable or mismatched local `git rev-parse HEAD`.
- **Head guard RETAINED, not subsumed.** The retired stamp compared the writer's **local** `git rev-parse HEAD` against the wrapper's **remote** `headRefOid`, so it spanned the whole review, not the split's few seconds. The replacement binds the invariant more directly: the body marker's `head=` (which per `file-review.md:18` *is* step 1's remote head) **and** the wrapper's own local HEAD, both against the resolved `headRefOid`. Both remain **staleness** guards, not anti-forge — anti-forge stays at the `merge-review` gate, which binds `author.login` / `commit_id` off the review object.
- **Portable mtime**: a BSD (`stat -f %m`) and a GNU (`stat -c %Y`) branch, matching `session_start.sh:114-116` / `helpers/status.sh:35-37`. A platform with neither returns non-zero and the caller fails closed — never a silent TTL skip.
- **Audit on every reject arm.** The wrapper had zero `audit_log` calls; the new fail-closed arms would otherwise be invisible blocks — a one-sided regression by MISSION's deferred-positive-face doctrine. Each `deny()` emits `file-review` / `rejected` carrying **the arm name only, never body content**. Two mechanism details are load-bearing: `audit_log` resolves its log via `ghjig_state_dir` (**not** `_cli`) and a Bash-tool subprocess commonly runs with `CLAUDE_PROJECT_DIR` unset, so the call carries an explicit per-project state-dir prefix; and it is `|| true`-guarded inside a subshell, since under `set -euo pipefail` a non-zero return could convert a clean fail-closed refusal into something else.
- **The absent-staging message names the absolute path it looked in**, so a writer/reader path divergence is a diagnosable error instead of a silent park.
- **Allow surface: zero additions.** The surviving entry stays exactly the wildcard-free `Bash(.claude/ghjig-root/scripts/ghjig_file_review_post.sh)`, and a new lock asserts **neither** settings file references the retired writer in `permissions.allow` in **any** form (bare or wildcard) while the two allow lists stay identical.

## Alternatives considered
- **Add the `:*` wildcard allow entry for the stage script** — rejected. The auto-mode classifier defers only on a byte-exact, wildcard-free match, so the wildcard buys nothing in the auto mode where the gap lives, while breaking the no-wildcard-allow discipline §148b/§148c/§144e lock for its sibling entries — and it would auto-approve an arbitrary-path read whose content the wrapper publishes verbatim.
- **Make the writer bare (keep the split, drop the argv)** — rejected as pointless: with the argv gone the writer has no job left, and both its outputs are derivable (`created=` → the staging file's mtime, now that the agent *is* the writer; `head=` → the marker the body already carries).
- **Declare the head-bind subsumed by the single step** — rejected by activation-reviewer round 2 as unsound; the guard is retained and retargeted instead.
- **Wrapper-side secret/content scan of the body** — ruled out of scope by `plan-reviewer` (it would breach the producer-only invariant); the egress-content residual is owned by follow-up **#634**, which the SPEC residual sentence back-references.

## Key context
- `scripts/ghjig_file_review_post.sh` — the single producer surface: `slurp→unlink→validate→post`, own-PR guard + GHES host pin (#610), hardcoded `event=COMMENT`.
- `.claude/hooks/hookrt.sh` — `ghjig_state_dir_cli()` (four-branch resolution, single definition, `:95-97` on the often-unset `CLAUDE_PROJECT_DIR`), `ghjig_state_dir` / `audit_log`.
- `.claude/hooks/helpers/ac_closeout_gate.sh` — the canonical marker regex the wrapper shares byte-identically.
- `scripts/test/smoke.d/70-gates-contentlocks.sh` §148h/§148i/§148j; `scripts/test/smoke.d/90-ghes-host.sh` `s610_stage()`.

## Checklist
- [x] **Doc** (`d13cc32`): SPEC §5.7.1 (one-link producer + mtime TTL + re-stat equality + head guard RETAINED + two-sided security ledger), §5.29 (marker contract as the wrapper accept set; counts align in shape/regex but not scope; `rejected` added to the file-review audit set), the §5.29 tree entry, `.claude/commands/file-review.md`, changelog fragment. ToC regenerated.
- [x] **Test** (`89b2578`): §148h / §148h-bare / §148h-ord / §148h-fc / §148i-del / §148i-set / §148i-cli / §148i-clibranch / §148i-marker / §148i-audit / §148i-doc rewritten for the single-wrapper shape; §148j behavioral round trip over 12 inputs; §148j-headmiss and §148j-future **retained and retargeted**; `90-ghes-host.sh` `s610_stage()` writes the staging file directly. Landed at **1187 green / 13 red**, every red attributable to the unwritten Code phase.
- [x] **Code** (`0e70269`): stage writer deleted; wrapper read/validate path rewritten (staging leaf, portable mtime, re-stat, symlink guards on leaf + directory component, marker count, both head arms, audited `deny()`).

## Out of scope
- Wrapper-side body **content** validation / secret scanning — the egress-content residual is #634's.
- Any change to the merge gates (`merge-review`, `ac-closeout`, `push-parity`) or to `resolve_self_review_policy`.
- The own-PR author guard, the GHES host pin (#610), and the hardcoded `event=COMMENT` — unchanged.
- The direct `gh api …/reviews` path for another author's PR / a non-current own PR — unchanged.
- The `nlink > 1` hardlink case and the ancestor-symlink case *above* the `file-review` component — disclosed as named Doc residuals, deliberately not guarded.

## Risks
- **The producer path has never executed** (0 `file-review` records in the audit log), so the behavioral smoke round trip is the coverage. It drives an agent-written staging file to exactly **one** POST on a fresh marker-matching body and **no** POST on each fail-closed input; the arms were additionally exercised by hand end to end (each reject writes its per-project audit record with no body content).
- **Anti-vacuity**: §148i-doc's only positive anchor was the stage-ref this change deletes — it now anchors on the staging-path literal plus the wrapper path.
- **Writer/reader path agreement** is a runtime property of `ghjig_state_dir_cli`'s four branches; §148i-clibranch pins the live Bash-tool branch, the doc names the condition and the fallback, and the wrapper's message names the path it looked in — but a worktree/override session can still diverge.
- The staging **leaf rename** is absorbed by no existing lock, so it had to be applied consistently; §148h now pins the new leaf *and* the absence of the old one.

## Deviations from the Plan
- **Audit env prefix.** The Plan named a `GHJIG_ROOT` prefix, which would land the record on the legacy shared path. The wrapper instead prefixes the per-project state dir it resolved via `ghjig_state_dir_cli`, canonical per SPEC §3.2.2; the AC requires an explicit prefix, not which one. §148j-audit asserts the per-project landing.
- **Settings deletion → anti-return lock.** The dead `settings.injected.json` stage entry the post-activation scope note targeted was never committed, so both allow lists were already clean. The AC's substance lands as §148i-set (no reference in any form + list parity) rather than as a deletion.

- **The AC's `CLAUDE_PROJECT_DIR` prefix is implemented; an earlier deviation was reverted.** `post.sh:60` briefly used `GHJIG_STATE_DIR_OVERRIDE`, which SPEC §3.2.1/§3.2.2 designate a test-only seam and which #633's AC5 amendment explicitly forbade. A comment on #633 defended that with a measurement; **that comment is retracted** — the measurement was taken with the override unset, a state branch 1 cannot be in, since the override is where `esd` came from and `ghjig_state_dir` consults it first. Re-measured across every resolution branch, both forms return the identical directory, so the AC's strip is exact and the deviation had no justification. `aca95e6` implements the AC's form and re-pins smoke §148i-audit, which had accepted **either** form — the latitude that let an AC and its implementation disagree with no red test.

## The security fix this PR introduced and then closed

**This PR introduced a silent fail-open and fixed it.** It is the most decision-relevant fact about the diff and it belongs in the merge record, not only in the commit log.

`b55ea1e` — **restore the octal mtime arm.** The base validated the staging mtime as `''|0*|*[!0-9]*`; an earlier commit here dropped the `0*` arm, reasoning that `stat` never emits a leading zero. That reasoning is wrong about the *failure mode*, not just the input: with a leading-zero epoch the TTL's `$(( now - mt1 ))` raises `value too great for base`, and an arithmetic-expansion failure does **not** fail the `[ … ] || deny` test — it **skips the whole list**, so the freshness check is never evaluated and execution reaches the POST. A security vote reproduced it: a body labelled `STALE-BODY-THAT-SHOULD-NOT-POST` published. That contradicts the SPEC promise that there is never a silent TTL skip.

Two earlier reviewers had cleared the same removal by reasoning — one that `stat` emits no leading zeros, the other that octal interpretation could only shrink the value toward `stale`. Both were wrong, and only the pass that **executed** it found the hole.

`bce129f` — **record that the fall-through is script-file-specific.** Run as a script file, the arithmetic error prints and control continues past the TTL to the POST (exit 0); the same statements under `bash -c` **abort** (exit 1). So a `bash -c` probe shows the *safe* behaviour and would invite deleting the arm a second time — the same wrong conclusion, now reachable by a test that does not model the wrapper's real invocation mode. The rationale is in the code so the next reader does not re-derive it.

Verified in both directions rather than only the one the fix addresses: `0178659606`, `00` and `0` all deny `mtime-malformed`; fresh epochs at 0s, 30s and 59s still proceed; a genuinely old epoch still denies `stale`. Two independent security votes reproduced both directions in script-file mode, and one confirmed the malformed `case` sits *above* the arithmetic, so the error is not merely survived — it is unreachable.

**Known gap, filed rather than fixed here:** the restored arm has **no regression lock**. `git grep` over `scripts/test` finds an octal lock only for the sibling case in `escape.sh` (§125-10, #479). So a future refactor could delete this arm a third time with all 1213 assertions green, and a `bash -c` probe would agree it was safe. Two votes independently flagged this and one executed the remedy — a `stat` shim in §148j's existing shim dir asserting `posts == 0`, which discriminates correctly (head denies; the `0*`-removed variant posts). It is a test-only addition that changes no security surface, so it is deferred rather than taking a fifth voting round — and the owner is **#647**, which carries the lock, a paired arm so the lock cannot pass by blocking everything, the missing SPEC/command enumerations, and two further items the producer pass found: `.claude/hooks/hookrt.sh:96`/`:102` still describe the deleted writer, and **the same fall-through class is still live on the adjacent operand** — `now=$(date +%s)` is unvalidated, and a leading-zero `now` published an hour-stale body in a reviewer's execution. That line is unchanged from the base, so it is inherited rather than introduced here, but one operand is guarded and its twin is not.

An earlier version of this section asserted the follow-up was already filed when it was not. The producer review caught it, and the correction matters beyond bookkeeping: the entire justification for deferring the lock is that someone owns it, so an unfiled "filed" is exactly the §6.0 P4 shape this repo refuses elsewhere — guidance with no gate behind it.

## Test plan
`scripts/test/smoke.sh` **1213 pass / 0 fail** (post-rebase onto `main`, which brought PR #641's smoke §156); `./scripts/lint.sh` clean over 79 files. §148j drives the real wrapper under the gh shim: one POST on a fresh marker/head-matching body; **no** POST on absent / empty / stale / future-dated / mtime-changed-during-slurp / zero-marker / two-marker / marker-head-mismatch / local-head-mismatch / symlinked-leaf / symlinked-directory-component. §153-1 and §153-5 (`posts == 1`) in `90-ghes-host.sh` are green with `s610_stage()` writing the staging file directly. §148a-§148c / §148d / §148i-reg / §144a-§144e stay green. §103c (SPEC §8 `scripts/` listing vs disk), red since the Doc commit, closes with the deletion.

## Review round

`code-reviewer` returned **`ship after fix`** at `0e70269` (head-pin blind-compared and matched), with no blocker. It verified the head-staleness guarantee by **execution** rather than reading — a `gh` shim and a throwaway repo, 13 inputs through the real wrapper — and found the replacement's span equal-or-wider than the retired stamp's (the marker is written at step 1, before the review, where the stamp read local HEAD only after it) with a strictly larger fail-closed set. It also independently verified the rebase's ToC conflict resolution: the removed and added ToC label lists are identical at 73 rows each, so only line numbers moved.

Four items, all doc/observability class, closed in `6ae4ec8`:

1. **The marker's `head=` was documented as "redundant human/grep convenience"** in both SPEC §5.29 and `.claude/commands/file-review.md`, while this PR makes it arm 1 of the head-staleness guard. The command contract is what the agent reads *while authoring the marker*, so calling the field redundant invited re-deriving it at write time — the one input that silently vacates that arm. Both sites now require it to be step 1's privately computed `HEAD_SHA`, never re-derived.
2. **"Emitted on every fail-closed reject arm" was over-claimed** — true only of the staging-validation arms; the preflight and identity arms refuse before the state dir is resolved, so they carry no audit line. Narrowed in SPEC and in the command contract. The remedy that would have extended coverage instead was left alone because the reviewer measured it as a silent no-op at those sites.
3. **"Every reject path also drops the file" was over-claimed** — false for the arms evaluated before the unlink. Narrowed.
4. **§6.0 P4: the new blocks named their condition but not the recovery.** Six `deny` messages now carry the positive next action. Because appending text to those messages could have broken the smoke regex that matches them, the suite was re-run — §148h-fc passes and the suite is 1213/0.

**This PR is a `merge-security-surface` diff under SPEC §4.11**, on the reviewer's explicit and unhedged determination: it rewrites the read-and-validate path of untrusted external input at an IO boundary that was previously regular-by-construction, it sits inside an identity-based authorization decision, its publish channel is public and unretractable, and — decisively — the diff itself converts a structural block into an out-of-band per-call adjudication the shell neither controls nor audits, which is what #634 exists to track. The merge therefore takes an **N=3 independent `security-reviewer` majority (2 of 3)** rather than a single verdict; those votes ran at `6ae4ec8`, were invalidated by the head advance, and have been re-run at each subsequent head. Four voting rounds in total — see the security-fix entry below for why.

## Docs touched
`SPEC.md` §5.7.1, §5.29 (+ tree entry, ToC regenerated); `.claude/commands/file-review.md`; `changelog_unreleased/fixed/633.md`.

## Ship gate
- [x] code-reviewer clean (`ship after fix`, all four items closed in `6ae4ec8`) · [x] smoke green (1213/0) · [x] lint clean (79 files) · [x] doc sync · [x] CI green (smoke ×2, bash -n ×2, mutation kill-rate, fragment-gate) · [x] changelog fragment present


ilgyu-yi (commented):
Automated code review materialized by `/file-review` (SPEC §5.29). Verdict: **approve**.

`code-reviewer` returned `ship after fix` at this commit with its `reviewed-head` independently derived and blind-compared by the caller (SPEC §4.5, #544); both required items were **body/Issue-only**, so the head did not advance and this review is pinned to the reviewed artifact. They are closed: the deferral of the restored mtime guard's regression lock now has a real owner in #647, and the body records the correction that an earlier version asserted that follow-up was already filed when it was not.

This diff is a `merge-security-surface` change under SPEC §4.11 and carries a **unanimous 3/3 independent security majority** at this head, with no High or Medium finding. The tier ran four times across the cycle; two of those rounds were consumed by defects the reviews caught, and both are recorded in the PR body rather than only in the log.

The most consequential of those: this PR introduced a silent fail-open and then closed it. Dropping the `0*` alternative from the staging-mtime validation let the TTL arithmetic raise "value too great for base", and an arithmetic-expansion failure skips the whole `[ … ] || deny` list rather than failing it — so the freshness check was never evaluated and execution reached the POST. A security vote reproduced it by execution; the arm is restored with its rationale in the code, including that the fall-through is script-file-specific and a `bash -c` probe shows the safe answer.

Verified by the reviewers rather than accepted: the restored guard denies a leading-zero mtime and does not over-block fresh epochs; no sibling of that defect class remains reachable; posted bytes equal validated bytes and `commit_id` is the single value both head arms validated against; the test-only state-dir seam is gone from production code and the AC's prefix form resolves identically across all four state-dir branches; both settings files carry the same two wildcard-free allow entries with zero additions.

Known and owned rather than hidden: the content channel is open by design (#634), two filesystem residuals are disclosed in SPEC, and #647 carries the missing regression lock plus a live twin of the same fall-through class on the adjacent unvalidated operand.

Issue #639, opened 2026-08-13
directive: attributable reject count so a rejected method must change


ilgyu-yi:
## Objective

Give **repeated rejection a per-change attributable count**, so a method that has been rejected repeatedly is forced to change rather than re-attempted.

The observed failure, reported from onboarded target repos: on a large change a review gate rejected the work **8 consecutive times**, and each round the agent re-attempted the *same* method with a slightly different patch. Repeated rejection carried no method-level information to it. This is negative enforcement with no door — the one-sided regression MISSION names — and it is the costliest half of the drift problem, because a wrong first approach becomes an eight-round cost rather than a one-round cost.

**The substrate gap**, measured while designing #636 and the reason the ladder could not be built there:

1. `code-reviewer` blocks on ten axes (§4.5) and **no external record carries the reason class** of a block. The #361 reviewer-reject reason-class trail (§6.1) covers exactly `issue-reviewer`, `plan-reviewer`, and `activation-reviewer`; `code-reviewer` is absent from it.
2. On the authed user's **own** PR, GitHub rejects a self change-request, so `/file-review` (§5.29) posts a comment-kind review carrying a block marker. This shell's PRs are self-authored and self-reviewed — the premise of the #584 unattended self-merge path — so a change-request **event count is structurally zero** in the dominant population.
3. A raw count is wrong in a third way: for a §4.11 high-asymmetry decision, **three independent reviewer invocations occur at one head** with a fixed denominator of three. Three same-head blocks there are round **1**, not round 3 — and this Directive's own first Execution PR merges an ADR, which is exactly that case.

**Selecting the attribution substrate is this Directive's primary work.**

Artifact-level boundary — done when all of the following have merged:

1. **One ADR** (`docs/ADRs/`) recording the substrate decision with the alternatives and a rejection reason for each. Candidates, at minimum:
   (i) extend `/file-review`'s materialized marker with a reason class (§5.29, `.claude/commands/file-review.md`);
   (ii) extend the #361 reason-class trail (§6.1) to `code-reviewer` — whose coupling must be weighed: `scripts/promotion_candidates.sh` selects on a warn/reject event pair and aggregates by category and class at a threshold of three, so adding `code-reviewer` would inject reviewer blocks into an *advisory-to-hook-promotion* report where a block is not an ignored advisory;
[...]

Issue #638, opened 2026-08-13
directive: canonical naming on durable artifacts, no invented aliases


ilgyu-yi:
## Objective

Require **canonical naming on durable artifacts**: referents are named by their real identifiers — SPEC section numbers, command names, agent names, file paths, function and variable names — and never by a short form or ad-hoc alias coined for the convenience of the document being written.

The observed failure: while authoring an Issue, a Plan, or a PR body, the agent coins a private vocabulary — labelling tiers, alternatives, or failure modes with single letters and digits — instead of naming things by what they are.

**A verified, merged instance in this repo.** The term `blast radius` is used in the generic *reach* sense at `SPEC.md:1076`, `SPEC.md:1440`, and `SPEC.md:1790`, while `.claude/hooks/helpers/blast_radius.sh` defines it as §4.11's **wrong-approve asymmetry** classifier over a closed decision set. Two concepts under one term in one SSOT means a sweep on either returns the other. This collision is in merged code today, recoverable by any engineer with a `git grep`, and it is the corpus this Directive's reader is measured against.

**Why this is a correctness rule, not a style preference.** A durable artifact is a **future search target**. Every retrieval mechanism the shell has — #636's term sweep, `scripts/spec_drift_candidates.sh`'s path-token grep, `/recall`'s pointer search, an engineer's `git grep` — reaches only what is named canonically. An invented alias is a **pre-broken fan-out target**: a site no sweep can find, created deliberately at authoring time. It also severs the artifact from its subject, so a reader cannot verify the claim and a later change cannot locate the reference. The cost falls entirely on future readers and future sweeps, which is why the authoring agent does not feel it.

Artifact-level boundary — done when all of the following have merged:

1. **One new SPEC section** stating three parts:

   (a) **The rule.** Durable artifacts — Issue bodies, PR bodies, Plans, commit messages, changelog fragments, code comments, and shell-authored documentation — name every referent by its canonical identifier. For commit messages the obligation binds the **body**, not the subject, which §10.2 caps at 72 codepoints.

[...]

Issue #637, opened 2026-08-13
directive: subtraction-first authoring so history lives only in git


ilgyu-yi:
## Objective

Make authored prose stop being this repo's dominant defect source: a doctrine that (i) fires the author from carrying facts, (ii) scopes truth-obligation to each artifact's consumption window, (iii) moves testable contracts out of prose into code, (iv) lets facts cross agent/time seams only as artifacts, and (v) makes deletion the default repair — so review rounds converge instead of minting new prose defects at the rate they close old ones.

**Revision 2026-08-25.** This revision broadens the Directive from the subtraction operation alone (prior body: archived as a comment below) to the full authoring doctrine, and raises Priority to P0 — the maintainer's decision of 2026-08-25, taken after the review threads of PR #677, PR #706 and PR #712 were re-read as one record. The subtraction rules survive unchanged as rules (a)–(d) inside item 1.

**Diagnosis.** Re-scoring those review threads by *when each defect was born* shows the dominant class is **born-false** — sentences already false at authoring (a census contradicting the derivation it cites; an example the corpus does not contain; a file asserted changed that the diff does not touch — each instance in the cited PRs' review comments) — not **aged-false** (true prose gone stale). Three mechanisms, all structural for an LLM author: generation and verification are one act, so plausible specificity is produced with the confidence of measured fact; defects concentrate at **transcription boundaries**, wherever a fact is hand-copied between surfaces; and a fact absent from context is **reconstructed rather than re-measured** — durable bodies are authored at session end, when context is noisiest.

**The seam law.** A *seam* is any boundary a fact must cross between contexts — agent↔agent, or now↔earlier within one session. Facts cross either as **artifacts** (a commit, a diff, command output, a verbatim-copied block: bytes, lossless) or as **narration** (a regenerated description — where confabulation enters). A seam is exactly as dangerous as the narration crossing it. Issue #711 is the static form: one contract restated as prose on both sides of a seam, drifting independently.

**Why P0.** The shell's anti-rot machinery (SPEC §1.10's evidence discipline) demands high-precision fact prose — exactly the form confabulation takes; optimizing against rot raised claim volume, and the added claims were born false. Meanwhile the repair reflex — rewriting every stale or wrong sentence — mints new unverified prose at the rate review closes it, which is why rounds did not converge.

Two verified instances of the subtraction face, re-verified at `45d497b`:

[...]

Issue #636, opened 2026-08-12
directive: change-reach-gated SSOT-change protocol to prevent doc drift


ilgyu-yi:
## Objective

Add a **change-reach-gated protocol for conducting an SSOT-affecting change**, decided at **plan time**, so doc drift is prevented while the change is authored rather than detected after it lands.

The observed **fan-out failure**: on a large SSOT change the agent greps the *new* vocabulary, patches those hits, and ships — leaving sites that used the *retired* vocabulary, stated the concept without naming it, or referenced it structurally (`§` cross-refs, Table-of-contents rows, table rows, filenames), plus coupled test and code sites. Measured instance (#172, retiring the `ship`/`refine`/`block` verdict vocabulary at `activation-reviewer`): the first-pass commit `d39f2fd` left **12** `refine` occurrences in `.claude/agents/activation-reviewer.md` — the file it was editing — plus occurrences in `SPEC.md` and 7 untouched files; the follow-up `15f35ad` had to sweep six more files including `scripts/onboard_target.sh` and `scripts/test/smoke.sh`. The failure point is an **under-enumerated candidate-term set**, not grep itself.

**What can and cannot be externalized (measured, not assumed).** Deriving the retired-vocabulary set mechanically from a diff does not work. Four variants were designed and all four died against the #172 fixture: an agent-recorded set always passes; raw removed-line tokens always fail; removed-minus-globally-present is a tautology (the residue is defined as disjoint from the set searched); and removed-minus-added returns 11 candidates of which 9 survive in touched paths and **all 9 are noise** (`since`, `filing`, `have`, `edited`, …) while the one true positive, `refine`, is **absent** — because the same commit added the supersession narration "these replaced the former `ship` / `refine` / `block`", so `refine` appears on an added line and cancels out. Retirement is also **scoped**: `refine` legitimately survives at `issue-reviewer` and `plan-reviewer`, whose grammar it remains (§4.7, §4.8), so no token-global presence test can express what was retired *where*.

Therefore this Directive externalizes the term set's **commitment time**, not its content. The declared scope is a **design fact** — *where is this term retired?* — not a **discovery fact** — *where does it occur?* That is why it is answerable at plan time, before the fan-out is known, and why plan-time declaration does not ask the author for knowledge they cannot have. The machine constrains only when the declaration was fixed and that it never shrinks; the semantic content comes from the author and the reviewer.

[...]

PR #641, opened 2026-08-13
docs(#640): SPEC §1.3.1 SSOT change sweep protocol

### LINKED ISSUE #640: SPEC states no protocol for conducting an SSOT-affecting change
Parent Directive: #636

## What

`SPEC.md` carries no section stating how an SSOT-affecting change should be conducted. The consequence is measured: at `d39f2fd` (#172) the first-pass commit retiring the `activation-reviewer` verdict vocabulary left **12** occurrences of the retired token in `.claude/agents/activation-reviewer.md` — the file it was editing — and the follow-up `15f35ad` had to sweep six further files including `scripts/onboard_target.sh` and `scripts/test/smoke.sh`.

This Issue writes the doctrine section only. It is boundary item 1 of Directive #636 and the Doc-phase prerequisite for the other two Execution Issues, which implement against the contract this section states.

The section must state three parts, per #636 item 1:

- **The candidate-term-set construction rule** — retired vocabulary, replacement vocabulary, paraphrases, and structural anchors (`§` cross-references, Table-of-contents rows, table rows, filenames).
- **Two tiers** — the term-sweep tier (default) and the read-through tier (affected sections read contiguously end to end), with the tier boundary fixed here.
- **Plan-time declaration and tier selection**, path-triggered — including the declaration's trailer form and scope grammar, the sub-threshold path, declaration immutability with its binding point at first push, monotonicity over `(term, scope)` pairs, and the `inconclusive` disposition. The scope unit is a heading anchor resolved by a generic markdown heading parse, with `scripts/build_toc.sh`'s numbered mapping as the SPEC-specific fast path; `§N` is sugar for the corresponding heading anchor rather than a second grammar.

The section must also state the **design-fact vs discovery-fact** distinction that justifies plan-time declaration: the declared scope answers *where is this term retired?*, which is a design fact known at plan time, not *where does it occur?*, which is a discovery fact that is not.

## Why

Serves Directive #636, which serves MISSION "The mechanism" — the selective-injection half. The tier decides how much SSOT to pull into context by measurement rather than guess, at plan time. Narrowing alone starves the agent, and grep-only patching of a wide change is that starvation.

**Why now:** #636 is Active and its other two Execution Issues implement against this contract, so nothing else in the Directive can start until this section exists. **Existing coverage:** none — `scripts/spec_drift_candidates.sh` (#462) and `/reconcile-spec` (#464) detect code-ahead drift, the opposite direction, and are blind to doc-ahead incomplete fan-out by construction.

## Acceptance criteria

- [ ] A new numbered SPEC section states the candidate-term-set construction rule, the two tiers with a fixed tier boundary, and plan-time path-triggered declaration.
- [ ] The section states the declaration's trailer form, its scope grammar including the `!` negation form, and that `§N` is sugar for a heading anchor rather than a separate grammar.
- [ ] The section states the immutability contract: binding at first push, widening permitted as a later trailer, narrowing forbidden, monotonicity defined pairwise over `(term, scope)` with containment, and an `inconclusive` disposition that exits non-zero and is consumed by `/complete-directive` rather than by `code-reviewer`.
- [ ] The section states the design-fact vs discovery-fact distinction.
- [ ] The section states the §1.9 relationship: §1.9 governs whether a mechanism is retired; this section governs how a change's fan-out is conducted. No contradiction is introduced.
- [ ] `scripts/build_toc.sh` is re-run and the Table of contents includes the new section.
- [ ] `docs/*.md` carry a thin pointer only, per §9 reference-not-restate; no detailed contract content is duplicated outside SPEC.
- [ ] The §1.9 harness-overlap classification gains a row for this mechanism, since coverage there is parity-guarded by smoke §116.
- [ ] Smoke asserts the new section exists and that the §1.9 parity guard passes.

## Out of scope

The reader `scripts/ssot_change_sweep.sh`, the `resolve_ssot_paths` helper, onboard seeding, and the carrier-file edits to `planner.md` / `doc-writer.md` / `code-reviewer.md`. Those are the other two Execution Issues of #636. This Issue writes doctrine only.

## Notes

Directive #636 records the four instrument designs that a measurement against `d39f2fd` discarded; the doctrine section must not reintroduce mechanical derivation of the retired-vocabulary set. Sibling Directives #637, #638, and #639 serialize their shared-carrier edits behind #636's carrier Execution Issue, not behind this one.

ilgyu-yi:
## Plan

Writes the doctrine section for Directive #636 boundary item 1 — the Doc-phase prerequisite for #636's other two Execution Issues. Doctrine only; the reader script, the per-project path resolver, onboard seeding, and the carrier-file edits are #636 Execution Issues 2 and 3.

**Contest record.** Base Plan A (planner) was beaten on two different decisions by two mutually-blind challengers, and the merged plan is what landed. `plan-reviewer` independently verified every empirical claim and corrected both challengers.

- **Placement — challenger B2 (cost/effort) dominates A.** A proposed a new `### 1.10` appended after §1.9. Smoke §116's parity window runs `### 1.9` → `## 2.`, so a `### 1.10` sits *inside* the counting window and a posture-token table row there inflates the derived count; A then spent an entire Phase C plus a red decoy test undoing a hazard its own placement created. `#### 1.3.1` sits before §1.9, outside the window. Phase C is consequently **empty**. Verified in the live tree at each commit: expected=64 rows=64 after Doc 1, expected=65 rows=65 after Doc 2.
- **Tier boundary — challenger B1 (correctness) dominates A.** A keyed the tier on an author-asserted *direction* (subtractive vs additive). That breaks #636's reader interface, which requires the tier to be a function of the three reported counts, and it is vacuously measured: an additive change has no pre-existing hits, so counts read zero regardless of true breadth. The landed rule is a single predicate on measured reach, with direction moved into the term-set construction rule.
- **Struck from B1.** Its GitHub ordinal anchor de-duplication rested on a false premise — #636's fixture declares a *path* scope, so no anchor resolution occurs — and the remedy contradicted #636's "anchors are semantic ids, not positions".
- **Corrected by the reviewer.** The section separates the **wide candidate-term set** (sweep input: retired + replacement vocabulary, paraphrases, structural anchors) from the **narrow declared retirement set** (the trailer, consumed by the survivor check). Conflating them produced a false claim in an earlier draft, that #636's success signal is unsatisfiable for an additive PR — the null-declaration clause is the third option and resolves it.

## Checklist

- [x] **Doc** — `#### 1.3.1 Conducting an SSOT-affecting change` under §1.3; §1.3 forward pointer; ToC regenerated in the same commit (§1.3).
- [x] **Doc** — §1.8 lever row + §1.9 `keep-as-policy` posture row + prose count, atomically (smoke §116 derives its expected count from the §1.8 family, so splitting the pair would leave a red commit between them).
- [x] **Doc** — `docs/ENGINEERING_FLOW.md` thin pointer (§9 reference-not-restate) + `changelog_unreleased/added/640.md`.
- [x] **Test** — smoke **§156** in `scripts/test/smoke.d/70-gates-contentlocks.sh`: 13 arms. Eleven content locks over a fence-aware, awk-bounded `#### 1.3.1` window, gated behind a **non-empty-window count guard** so a collapsed window fails loud instead of green-ing vacuous locks; one parity arm (§156j) asserting the new §1.8 lever row and §1.9 posture row land in the shapes §116 counts — the second parity arm was dropped as a duplicate of §116's own identity (see Review round). Full run `pass=1203 fail=0`.
  - **Locks proven to bite.** The cluster was run against three perturbed *copies* of SPEC plus an unperturbed control (the real file untouched): renaming `SSOT-sweep-tier:` fails 2 arms; un-backticking the posture token fails 2 — §116 and §156j are the only two arms that read a backticked posture token at head; and renaming the heading so the window collapses fails the count guard with `non-blank lines=0` while the eleven content locks correctly do **not** run — the empty-window anti-pattern the harness header names, converted into a loud failure.
- [x] **Code** — n/a. Doctrine-only Issue; the assertion surface is smoke. Precedent: `14425a4` (#528) carries the same shape, followed by a green-at-Doc `test(#528)` commit.

## Review round

`code-reviewer` returned **ship after fix** at `a2f9ab4` (head-pin blind-compared and matched). Four findings, all closed in `f5af42c`:

1. **The empty declaration had no grammar production.** The section made absent-vs-empty load-bearing while the fenced grammar had no token for empty — and this PR's own commits then invented `SSOT-sweep: none`, a form the later reader would have had nothing to parse. The doctrine's first application demonstrated its own gap. Now stated as a production.
2. **`resolved SSOT path set` was undefined anywhere in SPEC.** A reader would have resolved it against §1.3's "Target SSOT structure" list, which omits `.claude/agents|commands|templates/*.md` — exactly the carrier files #636 Execution Issue 3 edits. Until this was fixed, the PR body's own claim that the doctrine is "normative for a human author even with no reader" did not hold.
3. **`never be narrowed` dropped the parent's re-anchor carve-out**, making the common case — rename a heading, re-anchor the declaration — a spec-level violation. This was the one place the section contradicted #636 rather than abbreviating it. Carve-out restored.
4. **The tier boundary was unlocked by the tests** — the single decision a challenger won over the base plan, and the thing AC1 calls "a fixed tier boundary". A revert to the direction-keyed form would have left all eleven arms green. Now locked by §156l, with §156m locking the monotonicity direction and §156n the empty-declaration token.

Also dropped **§156k**: it re-derived §116's predicate verbatim, and §116 already owns that identity and runs in the same suite. **Correction to an earlier version of this body:** it claimed the duplicate "had already drifted … so it would have failed for an unrelated cause". The reviewer measured both derivations — `ls .claude/agents/*.md | wc -l` and `find … -type f -name '*.md' | grep -c .` both return 11 — so the divergence is **latent** (it would appear only with a symlink or a directory named `*.md`), not active, and §156k would have passed. The removal stands on the duplication alone. §156j owns the row-shape information §116 cannot see.

Full suite at `f5af42c`: `pass=1203 fail=0`.

## Review round 2

`code-reviewer` returned **ship after fix** at `f5af42c` (head-pin blind-compared and matched), verified all four earlier fixes as real and correctly done, and confirmed `mergeable: MERGEABLE`. Two mechanical items were required; both are closed in `4c23a93`.

- **F1 — the three new arms sat outside the count guard they were indented into.** The `fi` closing the guard sat between §156i and them, so the section header's claim that a collapsed window cannot produce fake passes was false for three of thirteen arms. Not a false-green risk — an empty window fails all three greps loud — but the indentation misrepresented control flow. Guard close moved after §156n, which also restores arm emission order.
- **F2 — stale provenance count** in the guard comment. Partially closed, and the residue is recorded here rather than reported as done: the comment was moved `18 → 21`, and the window then grew to **22** when this same commit added a prose line to §1.3.1. So `:1897` still reads `21` against a measured `22` — the `22` reached the commit message and this body, not the code. The remaining line is an acceptance criterion of #643 ("the `:1897` provenance count is re-measured rather than re-typed"), so it is standing, not closed. This is the third consecutive attempt on that one line, each replacing a stale number with another stale number.

Two advisories were also taken, and the first one broke the build in an instructive way. The reviewer asked for the fence's alternation to be marked, since that passage exists so a machine can read it. The first attempt put a shell comment *inside* the trailer fence — where a line-oriented window parser cannot distinguish it from a depth-1 heading. It truncated this section's own smoke window from 21 lines to 13 and reddened eight content arms. So: the alternation is now stated in prose, **the window extractor is fence-aware**, and the guard floor is raised from 10 to 18 — because 13 lines cleared a floor of 10 and §156a still reported the locks load-bearing over a cut window. The observed truncation is recorded in the extractor's comment.

The second advisory is taken too: the difference from §1.3's "Target SSOT structure" list is now stated in **both** directions rather than as an omission only.

Full suite at `4c23a93`: `pass=1203 fail=0`, `build_toc.sh --check` exit 0.

## Target base

`main`

## Out of scope, surfaced rather than performed

- The smoke §116 `## 2.` window terminator — a latent bug this PR provably cannot trigger, because it took the `#### 1.3.1` placement instead of a `### 1.10`. Filed as **#644**. An earlier version of this body claimed it was already filed when it was not; that claim was false and is corrected here.
- **A gap in #636 for `/revise-directive`:** the parent's success signal says `--survivors` returns a positive `survivors=0` for each Execution PR. An additive-only PR has no retirement to declare, so the parent should record the **null declaration as the third option** alongside declaring and omitting. This PR states the clause in doctrine; it does not edit the parent.
- The stale §1.9 prose count `(§6.1 — 17)` against a machine count of 20. Real, unguarded (§116 counts table rows, never prose), and not claimed here.
- **A third gap in #636 for `/revise-directive`**, surfaced by the reviewer: under §1.3.1's boundary, the term-sweep tier's selection domain is *exactly* the sub-threshold path (`¬(anchors>1 ∨ non-SSOT touched)` ≡ `one anchor ∧ no non-SSOT`), so term-sweep is never planner-selected and read-through is the operational default in this repo, where nearly every PR touches smoke. §1.3.1 is transparent about this ("in the sense that matters"), but #636's constraint sentence "the term-sweep tier stays default" holds only under that re-reading and should say so.
- Upgrading §116 from count parity to name-level identity parity — today a lever row and a posture row for *different* mechanisms cancel out.
- `.claude/CLAUDE.md` surfacing: 11995 bytes against a 12000-byte cap, so it needs a trim that pays for it. Belongs with #636 Execution Issue 3.

## Risks

- **Doctrine ahead of instrument.** SPEC states a contract whose reader does not exist yet. §5.27 classifies this as `spec-ahead` → track, no edit, and `spec_drift_candidates.sh` is code-ahead-only so it cannot false-fire. The doctrine is normative for a human author even with no reader.
- **Section-number collision.** Sibling Directives #637, #638, and #639 each plan a SPEC section. Depth-4 under §1.3 reduces the collision surface; a later section renumbering is itself an SSOT-affecting change and runs this protocol.
- **Self-application.** This PR is an SSOT-affecting change under the doctrine it writes. Its commits carry hand-authored `SSOT-sweep` trailers — necessarily hand-authored, since the reader is Execution Issue 2 — declaring an empty retirement set and the read-through tier.

Closes #640


ilgyu-yi (commented):
Automated code review materialized by `/file-review` (SPEC §5.29). Verdict: **approve**.

`code-reviewer` returned `ship` at this commit, with its `reviewed-head` independently derived and blind-compared by the caller against a privately held head (SPEC §4.5, #544).

Verified by the reviewer rather than accepted from the PR body: all 13 smoke §156 arms pass at head, and the new locks were shown to bite against three perturbed SPEC copies — reverting the tier boundary to a direction-keyed form fails only §156l, deleting the `SSOT-sweep: none` token fails only §156n, and renumbering the heading collapses the window so §156a fails loud while the eleven content arms correctly do not run. §116 parity balances (levers 6 + agents 11 + cmds 28 + hooks 20 = 65 = 65). `build_toc.sh --check` exits 0. shellcheck at the gate severity is clean. All nine acceptance criteria of #640 hold against the section text.

Both deferrals were judged legitimate on reproduction: #643 window over-extension was reproduced (363 non-blank lines) and then shown to produce zero hits for all eleven lock patterns over the swallowed range, so it weakens scoping without a live false pass; #644 terminator needs a future `### 1.10`, which this PR provably avoided.

Findings are advisory only: a provenance comment reading 21 where the measured count is 22 (already an acceptance criterion of #643), an over-general justification sentence that remains correct under both readings of the per-project path set, floor slack, and a commit-trailer ordering cosmetic.

Issue #648, opened 2026-08-13
the documented reviewer-reject audit idiom uses a bare relative helper path, so it fails open from the project root


ilgyu-yi:
## What

Four command files document the categorized reviewer-reject audit emission as:

    . ".claude/ghjig-root/.claude/hooks/hookrt.sh"
    safe_source helpers/reviewer_audit.sh reviewer-reject
    reviewer_reject_audit <tier> <reason-class> <id>

`safe_source` (hookrt.sh:230) takes a path and tests `[ -f "$helper_path" ]`. A bare `helpers/reviewer_audit.sh` resolves against the caller's cwd, so it only exists when cwd is `.claude/hooks/`. The main assistant runs these commands from the project root, where the test fails, `safe_source` returns 1, `reviewer_reject_audit` is never defined, and the intended categorized record is never written. Worse, the fail-open path writes a `helper-missing` warn record naming a helper that is present on disk.

Documented sites (all bare-relative): `.claude/commands/work-on.md:34`, `.claude/commands/file-issue.md:53`, `.claude/commands/activate.md:26`, `.claude/commands/complete-directive.md:37`.

The repo already uses the correct form elsewhere: hooks call `safe_source "$SHELL_ROOT/.claude/hooks/helpers/<name>.sh" <cat>` (pre_tool_use.sh:24-29), and `complete-directive.md:31` / `ship.md:18` source `blast_radius.sh` by its `.claude/ghjig-root/...` path. Only the `reviewer_audit.sh` idiom is abbreviated, and it is abbreviated in four places.

**Live corroboration, in this Issue's own thread.** The round-1 activation review of this Issue reproduced the defect during the very batch run that dispatched it: `safe_source helpers/reviewer_audit.sh reviewer-reject` from the project root failed open, `reviewer_reject_audit` was undefined, and the two `revise` verdicts in that batch (#650, #634) produced no categorized reject record at all — only the generic `helper-missing` warn (https://github.com/ilgyu-yi/ghjig-claude/issues/648#issuecomment-5300575926, "Independently corroborated during this same batch run"). The mechanism needs no corpus statistics: every command-context invocation of the documented idiom from the project root fails open, deterministically, until the call sites are fixed.

## Why

The harm is three separate things, only one of which is the lost record.

[...]

Issue #642, opened 2026-08-13
smoke false-positives when a worktree-isolated subagent is running


ilgyu-yi:
## What

Smoke assertions that `grep -r` over the **whole** `$SHELL_ROOT/.claude` tree also descend into `.claude/worktrees/agent-*/`, which are full repo copies created by worktree-isolated subagents. Any such assertion that counts *files* therefore double-counts while a subagent is live.

Observed during `/ship` on PR #641:

```
✗ 148i-cli: ghjig_state_dir_cli must be a single hookrt.sh definition, not re-defined in
  writer/wrapper (count=2 hookrt=1 wr_redef=0 wrap_redef=0) (#602)
smoke: pass=1200 fail=1
```

The two "definitions" were:

```
.claude/hooks/hookrt.sh
.claude/worktrees/agent-a67947742994b21c1/.claude/hooks/hookrt.sh
```

The second is the worktree of the `code-reviewer` subagent that `/ship` had spawned moments earlier — i.e. **the ship gate's own review step made the ship gate's own test step fail.** The same suite reported `pass=1201 fail=0` minutes before, with no code change in between; the only difference was whether a worktree-isolated agent happened to be alive.

The offending pattern is at `scripts/test/smoke.d/70-gates-contentlocks.sh:728`:

```sh
s148i_defs=$(grep -rlE '^[[:space:]]*ghjig_state_dir_cli[[:space:]]*\(\)' "$SHELL_ROOT/.claude" "$SHELL_ROOT/scripts" 2>/dev/null | sort -u)
```

Scope today is **one** assertion. Three other `grep -r` call sites over `$SHELL_ROOT/.claude` target subdirectories (`.claude/commands/`, `.claude/agents`) and so do not descend into `.claude/worktrees/`: `60-escape-identity.sh:2348`, `50-perproject-recall.sh:1585`, `40-targets-force-push.sh:351`. The pattern is the hazard, not just this instance.

`.claude/worktrees/` is correctly excluded from git (`.git/info/exclude:11`), so there is no commit risk — this is purely a test-harness scoping defect.

## Why

A gate whose verdict depends on whether an unrelated subagent is running is not a gate. `/ship` step 4 stops on a failing suite, so this can halt an otherwise-clean ship, and — worse in the other direction — it trains the operator to read a red smoke as noise.

[...]

PR #649, opened 2026-08-13
feat(#645): route review findings through a finding-judge layer

### LINKED ISSUE #645: review findings reach the author with no triage layer, unlike plan candidates
## What

The shell has a judge for plans and none for reviews.

- **Plan side:** `planner` produces Plan A, two mutually-blind `plan-challenger`s produce rivals on distinct axes, and **`plan-reviewer` judges the contest** before anything is authored.
- **Code side:** `code-reviewer` produces findings and they go **straight to the author**. Nothing verifies a finding is real, nothing verifies a proposed remedy survives, and nothing ranks severity.

Measured on PR #641 (5 review rounds, Execution Issue #640 under Directive #636):

| | judge present | outcome |
|---|---|---|
| Plan contest | `plan-reviewer` | **3 of 3 false items killed** — a challenger's simulated counts were off (66/65, not 65/64) and conditional; a second challenger's premise was true but its consequence false (the fixture declares a *path* scope, so no anchor resolution occurs); that same challenger's proposed remedy contradicted the parent Directive |
| Review rounds | none | a reviewer's own prescription — *"an inline `# one of` … would make the fence self-contained"* — **broke the build**: a `# `-prefixed line inside a fence is indistinguishable from a depth-1 heading to the §156 window extractor, truncating the window from 22 to 13 lines and reddening 8 arms. The same review had quoted the terminator it violated |

Two further consequences of the missing layer, both observed on the same PR:

- **No severity triage.** Items graded "advisory" in round 1 (fence metasyntax, the bidirectional §1.3 statement) resurfaced in round 2 because "advisory" was read as "later".
- **No independence on the fix.** The round-3 reviewer judged fixes the round-2 reviewer had prescribed — grading its own authorship, since both were the same resumed instance.

## Why

`code-reviewer` has drifted adversarial (the caller prompts it that way) **without** the layer that makes an adversarial stance safe. An adversarial role's success condition is "find something", so it carries a false-positive bias by construction; on the plan side that bias is absorbed by `plan-reviewer`, and on the code side it lands in the artifact.

The second, larger reason: **review-response changes are the only phase of the flow with no manifest.** The Plan governs Issue → first artifact. Review findings arrive *after* plan approval and generate changes no plan bounds. On PR #641 the split is clean — the four planned commits (`f3ab53a`, `397d81e`, `808a1d6`, `a2f9ab4`) produced **zero** overshoot defects; both review-response commits (`f5af42c`, `4c23a93`) produced all of them. A judge's output list *is* the missing manifest for that phase.

**Why now:** the pattern was diagnosed with measurements on #641 and the operator reports the same shape recurring across three separate repositories.

**MISSION fit:** MISSION's *Plan-as-manifest* narrowing lever (§1.8) covers the authoring phase and stops at plan approval; this extends the same lever to the review-response phase. It also serves the enforcement dual — a finding is a negative signal, and a judged, justified fix list is its positive face.

**Existing coverage:** none. `/review` and `/ship` invoke `code-reviewer` and hand its prose to the author. `/replan-check` compares the working diff against the Plan, which catches a *divergence* from a plan but says nothing about a change in territory the Plan never covered — the region where every observed defect lived. §4.11's N-way reviewer tier is a redundancy mechanism for high-asymmetry approvals, explicitly *blast-radius-gated, never a default*, and it multiplies reviewers rather than judging them.

## Acceptance criteria

- [ ] A new subagent, `finding-judge` (`.claude/agents/finding-judge.md`), positioned to `code-reviewer` exactly as `plan-reviewer` is to `plan-challenger`: it receives the reviewer's findings **and the source**, and emits a structured list. It **judges, it does not author the fix** — preserving the judge/author separation `issue-reviewer.md:62` already states for its own tier.
- [ ] Per finding, its output carries: **verdict** (`confirmed` / `refuted`) **with the command or measurement it ran**; **action** (`fix-now` / `defer-to-issue` / `drop`); and, for `fix-now`, whether the reviewer's proposed remedy **survives**.
- [ ] **Anti-swing contract** — the role exists partly to stop the author from moving a defect to the opposite end of the axis it failed on. Every `fix-now` item states the **axis**, the **target position** on it, and a **justification**, under three rules:
  - **A justification may not consist solely of the direction of the previous failure.** "It failed the other way" is not a position. An item whose only justification is directional is marked `unjustified — needs measurement`, and the judge measures rather than passing it through.
  - **A bound on a continuous quantity states both endpoints**, or states explicitly why one end is unbounded and why that is safe. (This one rule would have caught the §156 guard's floor-without-ceiling at list-authoring time rather than two rounds later — see #643.)
  - **An extreme position is permitted when it is justified independently** — by a §6.0 P1 cost-asymmetry, a definitional absolute, or a measurement. The rule is not "never go to the end of the axis": `--survivors`' fail-open → fail-closed move on #641 was correct and was justified by cost-asymmetry. What is forbidden is choosing a position *because of the last failure's direction*.
- [ ] **Swing detection across rounds** — the judge receives the previous round's (finding, remedy, justification) triple and flags a new remedy that occupies the opposite end of the same axis as the previous failure.
- [ ] **The list is durable, and written before the fix** — posted as a PR comment (or the PR body's round record) rather than living only in the caller's context, because it is the pair a later reviewer needs to detect a swing. Today the round record is written *after* the fact, which makes it a report; written first it is a plan.
- [ ] **Independence** — the judge is never the instance that produced the findings, and never judges the execution of a remedy it proposed itself.
- [ ] `/review` (§5.6) and `/ship` (§5.7 step 1) route the reviewer output through the judge before the author acts.
- [ ] SPEC gains a §4.x section for the agent, and — because smoke §116's parity is count-derived over `.claude/agents/*.md` — a **§1.9 subagent posture row**, or §116 goes red on the twelfth agent file. (`levers 6 + agents 11 + cmds 28 + hooks 20 = 65` today.)
- [ ] Smoke asserts the judge's output shape, with a **two-sided** pair per assertion: a fixture finding whose remedy is directionally-justified must be marked `unjustified`, and one whose remedy carries an independent justification must pass.

## Out of scope

- Adding reviewers. The measured bottleneck is not coverage: across five sequential `code-reviewer` rounds on #641, round 1 found four pre-existing defects and round 3 found **zero** pre-existing ones — a single pass is close to complete on a static artifact. Multiplying reviewers also contradicts §4.11's standing "never a default". This Issue imports the **judge** half of the plan-side pattern, not the multi-generator half.
- Changing `code-reviewer`'s verdict grammar or its Check list.
- Making the judge a merge gate. It is an authoring aid; §5.29 / §6.1 own the merge gates.
- Requiring `code-reviewer` to validate its own prescriptions. Self-validation is an assertion; the whole point of the separate layer is that it is an independent check. Requiring the reviewer to *label* an unvalidated prescription is a possible complement, filed separately if wanted.

## Notes

Naming: `finding-judge` rather than `review-judge` because it operates per **finding**, which is the unit its output is keyed to; `plan-reviewer` is likewise named for the artifact it judges.

Filed with the `issue-review` gate bypassed and audit-logged: the design is derived from measurements taken on PR #641 (commit-level defect distribution, the plan-contest false-item count, the fence-comment reproduction) with the rationale triad stated above.

ilgyu-yi:
Closes #645

## Why

`code-reviewer` produces findings and they go straight to the author. Nothing verifies a finding is real, nothing verifies a proposed remedy survives, and nothing ranks severity. The plan side has had that layer for a while — `planner` produces Plan A, two mutually-blind `plan-challenger`s produce rivals, and `plan-reviewer` judges the contest. The code side had no judge.

Measured on PR #641: the plan contest killed 3 of 3 false items, while the unjudged review rounds let a reviewer's own prescription break the build. Two further consequences on the same PR: items graded "advisory" in round 1 resurfaced in round 2 because "advisory" read as "later"; and the round-3 reviewer graded fixes the round-2 reviewer had prescribed, since both were the same resumed instance.

The larger reason: **review-response changes are the only phase of the flow with no manifest.** The Plan governs Issue → first artifact and stops at plan approval. On #641 the four planned commits produced zero overshoot defects; both review-response commits produced all of them. A judged, durable fix list is that missing manifest.

MISSION fit: extends the §1.8 Plan-as-manifest narrowing lever to the review-response phase, and serves the enforcement dual — a finding is a negative signal, a judged and justified fix list is its positive face.

## What shipped

A twelfth subagent, `finding-judge`, sitting between `code-reviewer` and the author. `.claude/agents/finding-judge.md` is the contract SSOT; SPEC §4.13 references it. It judges findings and never authors the fix, emits no verdict token, and gates nothing.

Per finding it emits: `verdict` (`confirmed`/`refuted`) with the command **and its invocation mode** actually run; `action` (`fix-now`/`defer-to-issue`/`drop`); whether the reviewer's remedy survives; and for `fix-now` the axis, target position, justification and swing status. A refuted finding is retained and **never convertible to `defer-to-issue`** — filing an issue for a false positive launders it into durable memory.

The judged list is posted to the PR **before any fix is authored**, one comment per round with a canonical marker. The judge fetches the prior round itself, under a trusted-author filter; the caller never hand-carries it.

### `/ship` step 1 was restructured

`ship.md` read "Invoke `code-reviewer`. Blocker → stop." A judge placed after that stop never runs on a `block` — the largest finding set, and precisely the case #641's evidence is about. Step 1 now invokes and blind-compares, step 1.5 judges every verdict, and *then* the blocker-stop is acted on.

**`code-reviewer`'s verdict token is untouched: a `block` still stops the ship.** What changes is only what the author receives at the stop — a judged, durable list instead of raw prose. No merge gate is added or altered.

## How the plan was reached

Two contest rounds. Round 1: base plan A, challengers on `cost/effort` (reasoned concession) and `robustness/risk` (dominated), plus an independent planner artifact. `refine`. Round 2: merged plan A′, challengers on `correctness` and `simplicity/maintainability` — **both dominated on different axes and their deltas composed**. `refine`, carrying the merge authored here. The Issue was parked at the plan-approval gate on the second consecutive `refine` and unparked by operator approval of the `/ship` step-1 restructure.

Two decisions the contest reversed:

- **The agent file lands in the Doc commit, not Code.** §116 is an *exact equality* over on-disk agent files, so agent-file-alone gives `66 ≠ 65` and row-alone gives `65 ≠ 66` — the file and its §1.9 posture row are inseparable. The consequence is stated rather than hidden: the contract locks are green-by-construction here (the sanctioned `70-gates-contentlocks.sh:2066` pattern), and the Phase-B RED comes from the routing surface. No doc prose was parked in Phase C to manufacture a fake RED.
- **No helper script.** An earlier draft added `scripts/finding_triage.sh` on a §9 one-home argument, but the design is "read in the judge, write in the caller" — the hardened read path has exactly one consumer, so the argument does not reach its conclusion. `activate.md` and `activation-reviewer.md` do the identical marker-counting as prose with no script. Dropping it also deleted the §103c coupling, the `gh` stub, and three would-be smoke pairs.

## Review rounds

Three `code-reviewer` rounds, each head-pinned and blind-compared, each judged before any fix was authored. The judged lists are PR comments (rounds [1](https://github.com/ilgyu-yi/ghjig-claude/pull/649#issuecomment-5278420898), [2](https://github.com/ilgyu-yi/ghjig-claude/pull/649#issuecomment-5278839019), [3](https://github.com/ilgyu-yi/ghjig-claude/pull/649#issuecomment-5288252554)); `fix-now` counts 6 / 5 / 6, all applied.

What the judge did that a bare reviewer pass would not have:

- **Ruled a reviewer's own remedy dead.** Round 1 offered "either a positive conjunct or a negative verb blacklist"; the judge measured both and found the blacklist walks past a paraphrased inversion. Only the positive conjunct landed.
- **Escalated an item the reviewer marked advisory.** Round 1's fixture-marker finding was "no fix required"; the judge reproduced three mutants that each left every arm green — one reading a *forged* fixture — and ruled `fix-now`, because an arm that greens on a forgery does not assert the output shape AC 9 names.
- **Narrowed findings against the reviewer.** Round 3's no-PR finding was reduced from "self-contradicts" to "under-specified", from "smoke defect" to "not a smoke defect", and noted the same shape is pre-existing in `code-reviewer.md`.
- **Corrected the reviewer's own transcript.** Round 3 found one claimed mutation as-written actually reds (a line-wrap artifact) and reproduced it correctly, validating the claim while fixing the record.

Round 1's single `defer-to-issue` was executed as #650. #648 was filed separately during the same session.

## A correction the review rounds forced

§157x — the arm locking `ship.md`'s non-conversion sentence — was hardened in four consecutive rounds, each against a new inversion a reviewer constructed on request. That hardening was a **category error**, and the final commit says so instead of continuing it.

Like every content lock in §157, this arm is a **drift guard**: it fires when a later edit deletes the sentence or changes its meaning. An author deliberately writing the inverse is not in scope and cannot be — a presence lock proves a sentence is present and can never prove no neighbouring sentence contradicts it. Measured: appending a contradicting section to the contract file leaves 157e green — an arm no round of this hardening touched (157e and 157x land together in the Test commit; both are new here). The threat model is now stated first in the comment so the next reader does not repeat the hardening.

The arm's behaviour is unchanged and verified against the threat model it actually has: sentence deleted → red, meaning reworded → red, cosmetic emphasis move → green.

## Commits

Doc → Test → Code, strict (`feat` on an external contract). The branch was rebuilt from nine commits to three: six remediation commits, most of them corrections of authoring errors rather than findings against the design, were folded into the phase each belonged to. What those rounds *found* is preserved — in the judged-list PR comments, and at the site where it has a live function (§157x's threat model). A PR comment records the mapping and the pre-rebase heads the judged lists still pin.

- `5e3b2f3` **Doc** — contract SSOT + SPEC §4.13 + §1.9 posture row + agent-roster repair
- `58b43ee` **Test** — smoke §157, three two-sided anti-swing pairs, routing arms (RED), `finding-judge` into the §77 and §92 loops
- `9aa6e85` **Code** — `/review` step 3.5 and `/ship` step 1.5 routing

The arc is verified rather than asserted: the two routing files under `.claude/commands/` are touched by exactly one commit, and at the Test commit they are byte-identical to `main`, so all three routing arms are genuinely unsatisfied there (measured: `pass=23 fail=3`, red on `157v`/`157w`/`157x` only).

## Roster repair, not increment

The sweep found `implementer` missing from five roster surfaces — the SPEC §1.5 table, the §1.7 artifact-tier table, `README.md:90`, `README.ko.md:90`, and `docs/SUBAGENTS.md` (no entry at all) — while `SPEC.md:339` still read "all ten" at 11 agents. Every *smoke-guarded* surface stayed correct; every *unguarded prose* roster drifted. Also repaired: `planner` missing from SPEC's isolation roster, `plan-challenger` missing from the §1.7 Execution row, a `7.5–7.7` drift at two sites, and two "reviewers emit a VERDICT" claims that a verdict-less judge falsifies.

`.claude/CLAUDE.md` was at 11995 of a hard 12000-byte ceiling (§104b), so the Doc commit carries an offsetting trim of its restated-from-SPEC isolation paragraph.

## Checklist

- [x] `finding-judge` subagent, positioned to `code-reviewer` as `plan-reviewer` is to `plan-challenger`; judges, never authors (AC 1)
- [x] Per-finding record: verdict + the command run, action, remedy-survival, plus a mode-aware evidence grammar (AC 2)
- [x] All three anti-swing rules, with a closed both-ends axis menu so "target position" and "opposite end" are expressible (AC 3)
- [x] Cross-round swing detection, with `prior-round:` tri-state so `swing: none` cannot be vacuous (AC 4)
- [x] The list is durable and written before the fix — a PR comment per round, caller posts, judge reads (AC 5)
- [x] Independence: structural on limb (b), honestly recorded as caller-side `dispatch:` on limb (a) (AC 6)
- [x] `/review` and `/ship` route through the judge before the author acts (AC 7)
- [x] SPEC §4.13 + the §1.9 posture row; §116 parity 65 → 66 (AC 8)
- [x] Smoke asserts the output shape with a two-sided pair per assertion, including anti-swing rule 3 (AC 9)

## Test plan

- [x] Full suite **pass=1239 fail=0**, with no worktree-isolated subagent live and no agent scratch archive present (#642 contamination avoided — one such archive was found and traced during this PR; see the note on #642)
- [x] `scripts/lint.sh` clean; `scripts/build_toc.sh --check` exit 0
- [x] §116 re-derived post-edit: `levers 6 + agents 12 + cmds 28 + hooks 20 = 66`, rows `66`
- [x] Intended Phase-B failure confirmed before Phase C: `pass=1236 fail=3`, red on the routing arms only, and proven *reachable* rather than permanently red
- [x] Every judged `fix-now` re-measured against the post-fix artifact, in the invocation mode the suite actually uses (sourced, not `bash -c`)

## Docs touched

- [x] `SPEC.md` — new §4.13; §1.5, §1.7, §1.8, §1.9, §2 diagram, §5.6, §5.7, §8 tree, §16, §17, ToC regenerated in-commit
- [x] `.claude/agents/finding-judge.md` (new, contract SSOT)
- [x] `.claude/commands/review.md`, `.claude/commands/ship.md`
- [x] `.claude/CLAUDE.md`, `README.md`, `README.ko.md`, `docs/SUBAGENTS.md`
- [x] `changelog_unreleased/added/645.md`

## Out of scope

- Routing `/ship` step 2 (`security-reviewer`) and the §4.11 combined finding set — the AC names `/review` and `/ship` step 1. Follow-up.
- A roster-parity smoke arm over `README.md` / `README.ko.md` / `docs/SUBAGENTS.md`. §34 *is* a README guard but iterates a hardcoded name list, so it is a floor that cannot detect a new omission; the follow-up should make it roster-derived. Follow-up.
- The evidence grammar's `mode=` enumeration (`script-file` / `bash -c` / `function-sourced` / `hook-fired` / `CI`) has no token for a **prose-artifact read**, which is what most records in a docs-heavy review are. All three judged rounds on this PR hit it and each wrote `mode=head-blob-read` with the gap disclosed. It is not an AC failure — AC 2 says "the command **or measurement** it ran", and unlike the axis menu the mode list is never declared closed — but the contract's "a wrong-mode measurement is not a measurement" is what makes the workaround load-bearing rather than cosmetic. Recorded here rather than fixed: adding a token would move the head for a gap with no measured harm, and it was never a judged finding (each round filed it as a judge's own note, which the contract forbids converting into a finding).
- Raising the judge's prior-round lookback past one round. The AC says "the previous round's triple", singular. A deeper window would serve a different purpose than swing detection — detecting that rounds on one site have stopped buying anything — and is a separate change.
- Adding reviewers or any N-way multiplication; changing `code-reviewer`'s verdict grammar or Check list; making the judge a merge gate; requiring `code-reviewer` to validate its own prescriptions.


ilgyu-yi:
## Finding triage (round 1)


Produced by a fresh, worktree-isolated `finding-judge` invocation. `dispatch:` the findings were produced by a `code-reviewer` invocation; the judgment by a different invocation with no shared context. Posted **before any fix is authored** (SPEC §4.13 durable-before-fix).

reviewer verdict: `ship after fix` — untouched. The judge emits no `ship`/`approve`/`block` token and enters no §4.11 tally.

```
reviewed-head: 02bd2c5b3f84b2e95c4e02080701d26be607144c
```

Prior-round read run by the judge itself: trusted-author comment filter returned 0 bodies, 0 canonical markers, so round = 1 + 0. Every `swing:` below is therefore `not-evaluated` or `n/a` — a `swing: none` at round 1 would be the vacuous pass the contract forbids.

---

## fix-now (6)

| # | site | axis | position | remedy survives | justification basis |
|---|---|---|---|---|---|
| F1 | `ship.md#step-1.5` | `fail-open↔fail-closed` | high | yes | definitional absolute inherited from the step-1 blind-compare contract + §6.0 P1 |
| F2 | `70-gates-contentlocks.sh#§157x` | `implicit↔explicit` | high | **positive conjunct only** | measurement (3 mutants) |
| F3 | `finding-judge.md#Fail-open` | none — discrete | n/a | yes | measurement: anchor is unique (count=1) |
| A | `70-gates-contentlocks.sh#§157m` | `fewer↔more` | low | yes | both endpoints bound at exactly 1, each measured |
| C | `SPEC.md#§4.13` | `implicit↔explicit` | high | **no** | author picks the end; both close the false claim |
| I3 | `70-gates-contentlocks.sh#s157_slice` | none — discrete | n/a | n/a | measurement: asserted property is absent |

**F1** — `/ship`'s zero-findings fall-through routes past the relocated blocker-stop. `code-reviewer.md` defines an `invalid` verdict that is findings-free by construction, so an unconfirmed-head verdict reaches step 2 with the stop skipped. Confirmed by head-blob read. Reviewer text measured to keep §157 at 26/26.

**F2** — 157x's attribute pile is satisfiable by text asserting the **opposite** of the contract. Reproduced exactly: replacing the step-1.5 line with *"The judge MAY convert a `block` into a pass … ; the author never alters the judged list"* leaves **26/26 green**, because `never alters` satisfies the verb alternative. The judge then tested **both** proposed remedies and the reviewer's "either" does not hold — against a **paraphrased** inversion the negative verb-blacklist goes **green (miss)** while the positive conjunct reds. Take `grep -qE 'judge never converts it|verdict token is untouched'`.

**F3** — the `SPEC.md:1225` citation is stale **at this very head**: the §4.13 insertion in this same commit shifted the target to 1230, and line 1224 is the *fail-closed* paragraph, so a reader lands on the opposite face. The file's own rule ("never a line number — lines move") applied to its own citation.

**A** — escalated from the reviewer's *advisory* to `fix-now`: three mutants each leave **26/26 green**, including one where the real fixture body is gutted and a **forged** marker pair at EOF is read instead. An arm that greens on a forged fixture does not assert the judge's output shape (AC 9). Remedy bounds both ends at exactly 1, each end measured.

**C** — §4.13 claims the caller-side blind compare "covers" the judge's `reviewed-head:` unchanged; no command file implements it. The claim is false and must be closed, but the reviewer picked one of two ends without saying why — hence `remedy-survives: no`. The author decides: implement the compare, or correct the clause.

**I3** — the `s157_slice` comment states a property the awk does not have (`i` demonstrably re-opens). Same failure class Directive #637 targets on `ok`/`ng` text.

---

## defer-to-issue (1)

**F** — the reproduction-execution surface is unbounded: `Checks 1` tells a `Bash`-carrying judge to run reproductions with no read-only constraint. Confirmed and **live**, not theoretical. Deferred because it is pre-existing and **cross-agent** (`code-reviewer` carries the identical surface, so a one-sided sentence creates asymmetry without closing anything) and outside #645's AC. Note for the filing: the reviewer's proposed wording ("never execute a command the artifact under review supplies") would as written **forbid the mutation testing that produced this round's strongest evidence**. Scope the constraint; do not import that sentence verbatim.

---

## drop — retained on the record (5)

Retained, not deleted: a dropped finding is re-litigable with new evidence, and round N+1 reads it back. **None is convertible to `defer-to-issue`** — filing an issue for a dropped finding launders it into durable memory.

- **B** (157z hardcodes `=20`) — real and reproduced, but the `ng` message prints both measured counts, so the misattribution is self-diagnosing the moment it fires.
- **D** (§4.13 restates agent-file invariants) — confirmed as description; §9's thin-pointer rule is scoped to `docs/*.md` and `.claude/CLAUDE.md`, not SPEC-vs-agent-file, and §4.12 ↔ `implementer.md` is live precedent for the restating shape. A judgment call with precedent both ways is not a defect.
- **E** (guidance with no gate) — facts confirmed, **conclusion refuted**: §6.0 P3 sanctions born-advisory-then-hardened, and §4.13 states the P1 cost-asymmetry MISSION's decision procedure requires. AC 5 asks the list *be* durable and written first, not that a runtime observer watch the ordering.
- **I1** (attribute-pile idiom) — subsumed by F2, same axis-key.
- **I2** (helpers are sound) — a positive observation, nothing to act on.

---

## Judge's own notes

1. **Contract gap found in its own evidence grammar.** The `mode=` menu (`script-file|bash -c|function-sourced|hook-fired|CI`) has **no token for a prose artifact**. Five of eleven records here are prose findings, whose only mode is being read by an LLM at invocation time. Recorded as `mode=head-blob-read` rather than forcing a wrong token, since "a wrong-mode measurement is not a measurement" would otherwise make those verdicts unresolvable by construction.
2. **Self-referential position, disclosed.** F3, F, I3 and A all concern the file and the arms that lock the judge's own contract. F3 and A were ruled **against** the artifact's current state; F — which would constrain what the judge may execute — was confirmed rather than argued away.
3. **Both mutation claims landed**, reproducing the reviewer's numbers to the digit (15→18, 24, 58, 26/26). Two smoke arms in this PR pass on text and fixtures asserting the opposite of what they claim to lock. That is the substantive content of this round.

ilgyu-yi:
## Finding triage (round 2)


`dispatch:` findings by a `code-reviewer` invocation; judgment by a different fresh worktree-isolated invocation with no shared context. Posted **before any fix is authored**.

```
reviewed-head: 051488ed5a12d18b9c69f8f3aaf5f0e2d3eff4a0
prior-round: 1@02bd2c5b3f84b2e95c4e02080701d26be607144c
```

Prior round resolved by the judge itself: trusted-author filter returned 1 body, 1 canonical marker, no duplicates. Round = 1 + 1 = 2. **`swing:` is evaluable this round.**

reviewer verdict `ship after fix` — untouched. The judge emits no `ship`/`approve`/`block` token and enters no §4.11 tally.

---

### The swing ruling — and the limitation it exposed

§157x has now been edited three times. Same axis-key, same axis (`implicit↔explicit`), same end (`high`) every time: round 0 **subtracted** a loose alternative, round 1 **added** a positive literal, round 2 **replaces** the pile with a whole-sentence literal. Monotone, no oscillation → **`swing: none`**, legal only because the prior round resolved.

Rounds 0 and 1 were nonetheless *directional* — each was a pile edit justified by the last mutant that got through, and round 1's stated "measurement (3 mutants)" measured a **paraphrase family**, not the discriminator. Round 2 breaks the pattern rather than continuing it, so rule 1 does not fire. **Had round 2 proposed a fifth conjunct, the judge would have marked it `unjustified — needs measurement`.**

> **The anti-swing rule did not fire, and that is the finding about the rule.** Three rounds on one site, all the same direction, so `swing:` stayed `none` every round while the arm stayed broken. Swing detection catches **oscillation**; it does not catch a **monotone ratchet against a tautological floor**. What would have caught this at round 0 is not the swing rule but a **discriminator measurement** — *does each conjunct discriminate?* Conjuncts 1 and 2 have been satisfied by the step-1.5 header line since the arm was written, and no round measured that until now.

---

## fix-now (5)

| # | site | axis | remedy survives | ground |
|---|---|---|---|---|
| F1 | `§157x` | `implicit↔explicit` high | yes | two conjuncts measured tautological; remedy measured against both mutants |
| F2 | `finding-judge.md:38` | none — discrete | **no** | ends are not symmetric; only one is consistent with round 1 |
| F3 | round-1 deferral | none — discrete | yes | an unexecuted item on the list is a counterexample to the contract being shipped |
| F4 | `§157m:2435` | none — discrete | yes | shared-shell sourcing makes an outliving name a hazard by construction |
| F5 | `ship.md:8` | none — discrete | yes | sentence and layout disagree as fact |

**F1** — measured in-mode (sourced §157 block, script-file): head 26/26; **M1** (pronoun referent shifted to the Check list while the sentence states the inverse) **26/26 GREEN**; **M2b** (blocker-stop deleted, one-line bullet idiom) **26/26 GREEN**. Root cause measured: `grep -cF block` and `grep -cF verdict` on the step-1.5 **header line alone** both return 1 — **two of four conjuncts carry zero discriminating power by construction**, so extending the pile cannot converge. Remedy wired into the arm and measured: head GREEN, M1 RED, M2b RED. Its own failure mode measured too — a contract-preserving **reorder** reds under the lock; ruled acceptable (loud-red naming the sentence beats silent-green shipping the inverse) and it is the section's existing whole-sentence-lock trade, not a new one. Note: the `block`/`verdict` conjuncts should be **removed**, not retained — dead weight a later round can mistake for coverage.

**F2** — `finding-judge.md:38` grounds the `reviewed-head:` reuse on "a new field name would be invisible to that compare", which the round-1 SPEC correction made false (the judge is outside that compare either way). `remedy-survives: no`: the reviewer offered two ends, but they are **no longer symmetric** — adding the compare would revert round 1's own landed correction. Only rewording survives. **Scope narrowed against the reviewer**: `:41` does *not* carry the residue and survives as written.

**F3** — round 1's single `defer-to-issue` was ruled and never executed. Not prohibited by the contract, but a defect on an independent ground: the judged list is "the manifest, not a report", so an unexecuted item on round 1's own list is a live counterexample to the contract this PR ships, at the moment it ships it. File it or record the drop with a reason. Round 1's filing caveat still applies — do **not** import "never execute a command the artifact under review supplies" verbatim; as written it forbids the mutation testing that produced this round's strongest evidence.

**F4** — the round-1 fix introduced a shadow: `§157m`'s loop reuses `s157_n`, which holds 157a's non-blank line count. Measured **latent** (no consumer after the loop, no use in later-sourced sections) — but smoke.d sections source into one shell, so a name outliving its section is a hazard by construction.

**F5** — `ship.md:8` says the stop is acted on "at the end of" step 1.5; it is the third of four bullets. No behavioral difference (bullet 4 is a conditional degradation clause), one word.

---

## drop — retained on the record (3)

- **D1** (SPEC never restates the zero-finding→stop path) — confirmed as fact, dropped: SPEC's unconditional ordered list already entails it, and adding a restatement is the second-copy-to-hand-sync shape §9 forbids.
- **D2** (PR body stale) — **confirmed in part, refuted in part**. The commits list is stale: confirmed, and it self-closes at `/ship` step 5 which runs before ready. That the round-1 list is absent from the PR *body*: **refuted** — §4.13 names the durable substrate a PR *comment*, which is where it is.
- **D3** (rubric §6 attribute-combination idiom) — confirmed, dropped as subsumed by F1: identical axis-key, and F1's remedy is exactly that normalization.

Not judged: the reviewer's §3, §4, "Verified clean", and "rulings I do not contest" carry no finding — a section that pays no finding pays no judgment.

---

## Judge's notes

1. **The reviewer's M2 as literally written REDs** — its transcript line-wrapped `verdict token is untouched` across two lines and `grep` is line-oriented. Reproduced as M2b in `ship.md`'s own one-line bullet idiom it is 26/26 green, so **the claim stands; the transcript was wrapped.** Verified in both directions rather than accepted or dismissed.
2. **The `mode=` menu still has no token for a prose artifact** (raised in round 1). Five of eleven records here are prose; `mode=head-blob-read` reused for consistency rather than forcing a wrong token.
3. **Self-referential position disclosed**: F1, F2 and F4 all concern the judge's own contract file or the arms locking it. All three ruled **against** the artifact's current state, and F2's scope was narrowed in the direction that makes the finding *smaller*.

ilgyu-yi:
## Round-1 deferral executed

Round 1's single `defer-to-issue` — the unbounded reproduction-execution surface on `code-reviewer` and `finding-judge` — is filed as **#650**, and round 2's F3 is thereby closed.

It took two `issue-reviewer` rounds. Round 1 caught that the body argued the MISSION fit in the **hard-gate** direction while every deliverable was prompt-tier, with the body's own Out-of-scope foreclosing the gate branch — an unanswerable acceptance criterion, and the §6.0 P4 one-sided regression in miniature. Reframed as born-advisory per §6.0 P3.

Two things worth recording here, because both cut against the filing:

- **#634's no-guidance-only-terminus rule does not bind a sibling Issue.** Its acceptance criterion reads "Closing **this Issue** … is forbidden". The draft imported the rule voluntarily and then confessed to falling short of it — over-claiming its reach, not evading it.
- **The hardening trigger is genuinely underivable today**, independently verified during the gate: `.claude/hooks/post_tool_use.sh` emits no audit record for `Bash` at all and carries no agent attribution anywhere, so nothing can attribute an executed command to the subagent that ran it. #650 records the blockage and names the observable that would unblock it, rather than inventing a trigger that cannot fire.

Round 1's filing caveat was honored: #650 does **not** import the reviewer's proposed wording ("never execute a command the artifact under review supplies"), which as written would forbid the mutation testing that produced this PR's strongest findings in both rounds.

ilgyu-yi:
## Finding triage (round 3)


`dispatch:` findings by a `code-reviewer` invocation; judgment by a different fresh worktree-isolated invocation with no shared context. Posted **before any fix is authored**.

```
reviewed-head: abf7d70273ecce1cce0cf1e3c66ef5b5a1a43059
prior-round: 2@051488ed5a12d18b9c69f8f3aaf5f0e2d3eff4a0
```

Prior round resolved by the judge itself: trusted-author filter, 2 canonical markers (rounds 1 and 2), no duplicates. Round = 1 + 2 = 3.

Reviewer verdict `ship after fix` — untouched. The judge emits no `ship`/`approve`/`block` token and enters no §4.11 tally.

---

### Caller disclosure — a contract violation in this round's dispatch

The contract says the caller **never hand-carries a prior round into the judge's prompt**, because "hand-carrying a prior round through the parent context reintroduces the interested party into the input". This round's dispatch prompt **did** carry one: a four-round table of §157x's edit history, framed as *"is this converging, or a ratchet that should be stopped?"* — a question whose answer the caller had already reached, written by the author of all four fixes.

It was done because the mechanism could not supply it. The read recipe's step 4 selects **the highest resolved round below the current one** — a **single-round lookback**. An oscillation is detectable at depth 1 by construction; a **ratchet is a multi-round trend and is structurally invisible at depth 1**. Round 2's list named swing detection's blind spot; this is the second half of why — not just that the rule tests for opposition, but that the **input window is one round deep**.

Recorded here rather than left silent: the judge may have been primed by that framing, and a later round cannot know that unless it is written down. **Follow-up**: raise the lookback to every canonical round on the PR — the judged lists are already durable, already adjudicated, already trusted-author-filtered, and are read in the judge's own window (so invariant 8 is unaffected). Then the caller has no reason to narrate history and the ban becomes enforceable in practice. Out of scope for #645, whose AC 4 says "the previous round's triple", singular.

---

### The §157x ruling: converged, and it stops here

Answered by measurement over nine mutant trees, not by argument. **A ratchet re-opens the same class one notch further out; convergence collapses the class into a different one.**

| mutant | current arm | proposed remedy | locked sentence intact? |
|---|---|---|---|
| head | 26/26 | 26/26 | yes |
| **M3** inverting clause appended past the literal's mid-sentence end | **26/26 GREEN** (the hole) | **25/26 RED** | no |
| **M-emph** cosmetic `**` move, contract-identical | 25/26 RED (false red) | **26/26 GREEN** | no |
| M-reword contract-preserving reword | 25/26 RED | 25/26 RED | no |
| M4 contradicting sibling bullet | 26/26 GREEN | 26/26 GREEN | **yes** |
| M4′ contradicting clause on another bullet | — | 26/26 GREEN | **yes** |
| M5 contradicting next sentence, same bullet | — | 26/26 GREEN | **yes** |
| **e4** contradicting section in `finding-judge.md`, hitting **157e** — an arm this PR never touched | — | **26/26 GREEN** | **yes** |

Rounds 0→1→2 each re-opened the *same* class: the arm kept greening on text inverting the very sentence it names. Round 3 is the first after which **no surviving mutant touches that sentence at all**. The class collapsed — that is convergence, and it is why the arm stops here, not because it is perfect.

**"Unfixable" was over-stated, and it matters.** Measured: a structural check (`bullets=4` vs `5`) *does* separate head from M4 — so M4 **as disclosed** is closable. M4′ and M5 keep `bullets=4` with the sentence byte-identical, so the accurate statement is **"no *presence* lock can exclude a contradicting neighbour"**, not "unfixable". The judge does **not** propose that check; it rules that the counsel-of-acceptance **survives** but its ground must be written as measured — otherwise round 4 reads "unfixable" and stops looking at the wrong boundary.

**The residual belongs to the idiom, not this arm** — proven by execution on `157e`, untouched by this PR. A round 4 on §157x would be attacking a section-wide property from inside one arm.

**Rule 3 applied to "extend the literal further":** it does **not** continue the four-round pattern. Prior rounds added *conjuncts* justified by the last mutant; this one moves the right endpoint onto the sentence's own terminating period (definitional, the only non-arbitrary endpoint) **and simultaneously relaxes** the arm on a second dimension — `tr -d '*'` flips M-emph from red to green. **Directional-only reasoning cannot produce a relaxation.**

---

## fix-now (6)

| # | site | remedy survives | note |
|---|---|---|---|
| F1 | `§157x` | yes, **incl. the comment rewrite** | right endpoint currently arbitrary (rule 2 violation) |
| F2 | `finding-judge.md` Artifact resolution | yes | **narrowed** against the reviewer, three ways |
| F3 | `SPEC.md:373` §1.7 tier row | yes | diff-introduced; §4.13 says "do not fix this by adding it" |
| F4 | `.claude/CLAUDE.md:57` | yes | 11945 B, 55 B headroom left |
| F5 | `docs/SUBAGENTS.md:5` | yes | fold to a pointer **without** restoring the false generalization |
| I2 | `finding-judge.md` citations | n/a | **escalated** — incomplete execution of round 1's own F3 |

**F1** — the locked literal has a stated left endpoint and an **arbitrary right endpoint** (mid-sentence, no terminator, no statement of why it is unbounded): an anti-swing **rule 2** violation in the arm itself. The remedy's comment rewrite must state four *measured* things: the reword false-red **persists** under the remedy too; the locked literal roughly **doubles** (127→232 bytes), doubling that false-red surface; the M4/M4′/M5 residual is real; and it is **shared with 157e/157p/157u**. One correction to the reviewer: its "both halves are wrong" is itself imprecise — the reword-reds half is **true**; what is false is the *trade* the comment claims.

**F2** — narrowed three ways **against** the reviewer: (1) not a smoke defect (§157j is a text-presence lock and greens either way); (2) "self-contradicts" is too strong — Fail-open is gated on "if you cannot judge", and in `--staged` one *can*, so it is **under-specified**, not contradictory; (3) the identical unconditional shape is **pre-existing** at `code-reviewer.md:16-19`, so fixing only one file leaves the two agents asymmetric — an observation for the author, not a finding the judge may add.

**F3** — independent ground the reviewer did not use: #645's AC asks for a **§1.9 posture row**, which is present, not a §1.7 tier entry. The `:373` addition is required by nothing and contradicted by §4.13. Annotating and removing are both defensible; the reviewer's annotate has a stated ground, so it survives.

**I2** — escalated above the reviewer's "low value" on a ground the reviewer did not have: round 1's **F3** was this exact axis-key, and its remedy landed the no-line-numbers policy sentence while leaving three counterexamples in the same file. So this is **unfinished round-1 work**, not a style nit. The load-bearing half is a factual error: the file cites `activate.md:34` as *counting* markers, but `:34` **posts** one — the counting shape is at `:16`.

---

## drop — retained (1)

- **I1** (rubric criterion 6 idiom) — confirmed, subsumed by F1: identical axis-key, and F1's `tr -d '*'` normalization *is* that idiom.

**Refuted: none.** The reviewer's two prose imprecisions are corrections *within* confirmed findings, not refutations.

---

## Judge's notes

1. The cycle **converged this round and should stop** — two independent measurements: the surviving-mutant class collapsed, and the residual is demonstrably a property of the presence-lock idiom, measured on an arm this PR never touched.
2. **Contract gap, third round running**: the closed `mode=` menu still has no token for a prose artifact. Six of seven records here are prose.
3. **Self-referential position disclosed**: F2 and I2 concern the judge's own contract file, F1 the arm locking the routing into it. All three ruled **against** the artifact; F2 was narrowed to be *smaller*, and I2 escalated *against* the reviewer's severity because the round-1 record — fetched, not received — showed it was unfinished work.

ilgyu-yi:
## History cleaned to Doc → Test → Code

The branch was rebuilt as three phase commits and force-pushed. The old graph carried six remediation commits on top of the three phases; **most of them existed because of authoring errors of mine**, not because a reviewer found something wrong with the design — a lock I wrote too loose, a literal that stopped mid-sentence, a comment sentence that was factually false, a wrong singular. Those are not history worth keeping in git.

| old | now |
|---|---|
| `4568b89` docs · `cac136b` test · `014de62` feat | `3f4c071` docs · `a7837d0` test · `a122cac` feat |
| `02bd2c5` · `051488e` · `abf7d70` · `1956394` · `6b4cfa9` · `e57d1d6` | folded into the phase each belonged to |

**The judged lists above still pin pre-rebase heads** — `head=02bd2c5`, `051488e`, `abf7d70`. Those commits are no longer on the branch. The lists themselves are PR comments and stay readable; what they record — what each round found, and what the judge confirmed, refuted, escalated or dropped, and why — is unaffected. Only the SHA each one names is now unreachable from the branch tip.

**What was deliberately kept, because it has a live function** (Directive #637 (b)): the §157x threat model. That comment records that the arm is a **drift guard**, and that hardening it against invented adversarial text was a category error. That is not supersession narration — it is what stops the next reader from repeating four rounds of the same mistake, so it stays at the site.

**What the arc now shows, verified rather than asserted:**

- The two routing files under `.claude/commands/` are touched by **exactly one** commit, the `feat`. At the `test` commit they are still at `main`'s state, so all three routing arms' predicates are absent — no `1.5.` line in the ship command, no `3.5.` line in the review command, and an empty step-1.5 block. That is the intended Phase-B red, and it comes from the routing surface rather than from doc prose parked in Phase C.
- The agent file and its §1.9 posture row are inseparable — smoke §116 is an exact equality over on-disk agent files — so both land in the `docs` commit and the contract locks are green-by-construction there, the sanctioned `70-gates-contentlocks.sh:2066` pattern.

Final tree at `a122cac`: full suite **pass=1239 fail=0**, lint clean, `build_toc.sh --check` exit 0.

A fresh head-pinned review follows. The merge-review gate reads `commit_id` off the review object, so the rebase correctly invalidates every earlier verdict.

---

*An earlier revision of this comment was posted with `gh pr comment --body "…"`, and two backticked paths in it were executed as command substitutions by the shell before `gh` ever saw them — the reason `/file-review`'s contract requires every body to travel as a written file rather than an inline argument. Corrected here via `--body-file`; the discipline applies to one's own text too.*

ilgyu-yi (commented):
## code-reviewer verdict: ship

Reviewed at the PR head, resolved independently by the reviewer and blind-compared by the caller against a privately held head.

**No commit-requiring findings.** Two body-only findings were raised and are already applied to the PR body; neither moves the head, so this verdict stands at the reviewed commit.

### What was verified by execution, not asserted

Every measurement below ran in the mode the suite actually uses — a driver that sources `smoke.d/_preamble.sh` and then the section under test into the **same** shell, never `bash -c`, because the two disagree on this repo's own record.

- **The Doc → Test → Code arc is real.** The phase split is disjoint: seven doc surfaces in `docs`, three smoke files in `test`, two command files in `feat`. The two routing files are byte-identical to the base at the `test` commit, so the three routing arms are genuinely unsatisfied there — measured `pass=23 fail=3`, red on the routing arms only. At the head: `pass=26 fail=0`.
- **Parity and ceilings.** Smoke sec-116 derives 66 and classifies 66; sec-103a/b, sec-92, sec-77 ("all 9") and sec-91 all green. `shellcheck --severity=warning --shell=bash` on the three changed smoke files: rc 0. `build_toc.sh --check`: rc 0. The always-injected `CLAUDE.md` is 11961 bytes against its 12000-byte ceiling.
- **The locks bite.** Gutting the axis table reds sec-157f; rewording the locked non-conversion sentence reds sec-157x; a cosmetic emphasis move does not. The comment describes what the code does.
- **The judge's read recipe works against this PR.** Running its trusted-author filter verbatim returns the three canonical judged-list markers, and `authorAssociation` is exposed on the comment payload — so the filter cannot silently return zero and disable round derivation.

### Review history

Seven passes. The first three produced judged finding lists posted to this PR before any fix was authored, which is the contract this change ships, applied to itself. Findings per pass ran 6, 5, 6, 4, 3, 2, 1, 0. The last four passes found only authoring slips — a wrong section number, a stale byte figure, a rename applied at one of three sites — and no design defects.

One correction is worth naming because it went the other way: an arm was hardened across four rounds against invented adversarial text, and that was a category error rather than four defects found. A content lock over prose is a drift guard; it can prove a sentence is present and can never prove no neighbouring sentence contradicts it. That residual is a property of the idiom, measured on an arm this change does not modify, and the comment now states the threat model first so the next reader does not repeat the hardening.

### Standing axes

Roster consistency, doc sync, enforcement-style fit, security surface and scope were each checked and are clean. The new layer adds no hook, no matcher and no merge gate; a negative smoke arm asserts that, and it requires the hook file to be present so a missing file cannot green it. All nine acceptance criteria on the linked issue are met, with the parity arithmetic re-derived rather than accepted.

PR #657, opened 2026-08-15
docs(#646): name the second isolation axis: trees, not scratch

### LINKED ISSUE #646: N-way reviewer votes share one scratchpad, so their harnesses can overwrite each other
## What

SPEC §4.11's `merge-security-surface` tier requires **three independent** `security-reviewer` invocations, and §1.5/#285 makes worktree isolation the canonical way to run them. Worktree isolation separates each reviewer's **git tree**. It does not separate the **session scratchpad**, which all subagents share.

Reviewers that verify by execution build harnesses there — shim binaries on `PATH`, throwaway repos, per-case work dirs. When three run concurrently and each picks a conventional relative name (`bin/gh`, `bin/stat`, `work/<case>/`), they collide.

Observed during the round-4 tally on PR #635, reported by one of the three votes:

> the session scratchpad is shared, and a concurrently-running agent overwrote my `gh` shim mid-review (I caught it because the shim errored on a variable it never defined). I re-ran everything from a uniquely-named private directory with the wrapper sha256-pinned to the head blob. If the other two invocations in this tally used the bare scratchpad root, their results may be contaminated.

The detection was **incidental** — the overwriting shim happened to reference an undefined variable and error loudly. A shim that merely behaved differently would have produced a silently wrong result, and a `security-reviewer` whose `gh` shim is not the one it wrote is reporting on something other than what it thinks.

The two sibling votes in that tally did use scratchpad-relative harness paths (`bin/{gh,stat,date}`, `p1/`, `work/<case>/`), so contamination was reachable for them.

## Why

**This defeats the property the tier exists for.** §4.11 replaces a single verdict with an N-way majority precisely so that one reviewer's error does not decide a high-asymmetry merge. That argument assumes the votes are independent. Context independence is enforced (no shared conversation, separate worktrees); **execution-environment independence is not**, and for reviewers that verify by running things, the execution environment is where the verdict is actually produced. Correlated contamination across votes is worse than a single wrong vote, because the majority rule converts it into confidence.

**It also plausibly explains a failure already on the record.** Round 3 of the same tally lost all three votes to an "infrastructure limit" with no verdicts. The reporting vote names shim collision as a plausible mechanism. That is unconfirmed, but it is the kind of failure that is indistinguishable from a resource limit from the outside.

**Why now:** PR #635 is a `merge-security-surface` diff, so it took the N=3 tier four times, i.e. twelve concurrent worktree-isolated reviewers over the cycle. The collision surface is proportional to exactly the reviews the shell treats as most important.

**MISSION fit:** the N-way tier is MISSION's "high-asymmetry safety redundancy" made concrete. Redundancy that is silently correlated is not redundancy — it is one sample reported three times.

**Existing coverage:** none. This is **not** #642, which is a different mechanism at a different location: there, worktree *copies of the repo* under `.claude/worktrees/` are walked by smoke assertions that `grep -r` over `$SHELL_ROOT/.claude`, inflating file counts. That is about the repo copy; this is about the shared temp scratchpad. The two are unrelated apart from both being consequences of running several isolated subagents at once.

## Acceptance criteria

- [ ] Each subagent invocation gets a **distinct scratch directory**, and the contract that it must use it is stated where the caller can be held to it — not left to each reviewer's naming discretion. Any of: a per-agent path the harness exposes, a documented convention keyed to the agent id, or an instruction the dispatcher is required to include.
- [ ] The reviewer contracts that verify by execution (`security-reviewer`, `code-reviewer`) state that a harness must live under that per-invocation path, and that any executable it places on `PATH` must be created there.
- [ ] A reviewer must be able to **detect** contamination rather than rely on catching it by luck: state the minimum check — pin the artifact under test by content hash, and re-verify that hash after the run.
- [ ] SPEC §4.11 records that the tier's independence requirement covers the **execution environment**, not only the conversation context, and names shared scratch as the known way it breaks.
- [ ] A smoke case asserts the dispatch path supplies a distinct scratch path per invocation, with a paired arm asserting two concurrent invocations do not receive the same one.

## Out of scope

Changing worktree isolation, and anything about #642's `grep -r` descent into worktree repo copies. Also out of scope: retroactively re-running PR #635's tally. Its verdict stands on the vote that demonstrably ran clean (private directory, wrapper sha256-pinned to the head blob) plus agreement across all three on the load-bearing executed facts; the contamination risk is recorded in that PR rather than treated as invalidating.

## Notes

Reported by a `security-reviewer` during PR #635's round-4 tally, with the detection mechanism stated. Filed with the `issue-review` gate bypassed and audit-logged: the finding is a reviewer-reported observation with a named detection path and the rationale triad stated above.

Worth stating for whoever picks this up: the caller (me) is the proximate cause. Nothing in the dispatch instructions I wrote told the three invocations to isolate their scratch, and nothing in the reviewer contracts required it. The fix belongs in both places, not only in the reviewers.

ilgyu-yi:
Closes #646

## Why

Worktree isolation separates each subagent's **git tree**. It does not separate the **session scratchpad**, which every subagent of a session shares by path. A reviewer that verifies by execution builds its harness there — shim binaries on `PATH`, throwaway repos, per-case work dirs — and, absent a convention, reaches for conventional relative names.

That matters most exactly where the shell cares most. SPEC §4.11 replaces a single verdict with an N-way majority so that one reviewer's error cannot decide a high-asymmetry merge. That argument assumes the votes are independent. Context independence is enforced, tree independence is enforced, **execution-environment independence is not** — and for a reviewer that verifies by running things, the execution environment is where the verdict is produced. Correlated contamination across the three is worse than one wrong vote, because the majority rule converts it into confidence.

**Measured, in this repo:** across a dozen-plus concurrently-isolated invocations one scratchpad root accumulated 42 bare `.sh` scripts with generic names plus single-letter directories from different agents, with four separate actors independently choosing the name `probe`. That is the precondition and a near-miss; the only observed clobber remains the one reported on PR #635. The plan is against the precondition and claims nothing more.

## Plan

The plan contest ran `planner` → two mutually-blind `plan-challenger`s on distinct axes → `plan-reviewer`. **Both challengers returned `dominates-A`, prescribing opposite mechanisms** — one dropping the helper, one enriching it. `plan-reviewer` judged the contest, carried the simplicity plan, and took exactly two items from the robustness plan. Round-1 verdict `refine` (`scope-bleed`); round-2 verdict `ship` on the carried plan.

## What is deliberately NOT here

Each omission is measured, not preference:

- **No helper script.** The harness exposes no per-subagent scratch path — probed directly: `CLAUDE_CODE_SESSION_ID` keys the *shared* scratchpad path and `CLAUDE_CODE_CHILD_SESSION` is a boolean, not an id. A script therefore cannot auto-discover the root either, so a helper would mint **outside** the scratchpad. `mktemp -d` already yields mode `700` and an absolute path, and its `O_EXCL` retry is the uniqueness guarantee under any design.
- **No caller-side distinctness check across the votes.** It is **inert on the motivating incident**: the colliding paths were `bin/{gh,stat,date}`, `p1/` and `work/<case>/`, which are distinct as strings, so such a check passes on the very collision it exists to catch. Its sensitivity would rest on each reviewer's reporting granularity — the naming discretion the convention exists to remove.
- **No `scratch:` verdict line and no collision → invalid-vote rule.** §6.0 P3 says harden after evidence. Compliance rates have not been observed, and the wrong-block cost is an invalidated vote parking a high-asymmetry merge.
- **No restatement across the five dispatch command files.** The `finding-judge` "Dispatch shape" line that would be the precedent for that ends: *"Do not restate the judge's output schema here — `.claude/agents/finding-judge.md` is the contract SSOT (§9)."*
- **Zero bytes in `.claude/CLAUDE.md`.** It sits at **11961** against a smoke-enforced **12000** (§104b) — 39 bytes of headroom. Its working-tree bullet already points at §1.5.

## Commits

- `269323c` **Doc** — SPEC §1.5 (the second axis), SPEC §4.11 (why it matters to an N-way vote), and the operative recipe in both reviewer contracts.
- `0221157` **Test** — §158a/b/c: extract the `mktemp -d` command *from the contract* and run it K=5 concurrently.
- `b4a0b73` **Code** — changelog fragment. (The substance of this change is a contract, so the contract edit *was* the implementation and Phase A carried it.)

Round 1 of review — the judged list is in the `<!-- finding-judge: round=1 -->` comment:

- `020a3e3` **Doc** — F1 (§4.5/§4.6 pointer dead-end), S-MEDIUM-1(a) (the pin wording oversold its coverage), F6 (evidence channel named), F7 (`<scratch-root>` resolution).
- `0d5dd99` **Test** — F2 (§158d locks *both* contracts), F3 (the stated RED reproduction could not produce a RED).

Round 2 of review — the judged list is in the `<!-- finding-judge: round=2 -->` comment:

- `c3ff57f` **Doc** — F-4/F-3 (**drop** §4.5's five-step restatement rather than sync it: round 1's own fix went stale against the contracts inside this PR, which is the measurement that settles it), F-2 (15 `safe_source` *call sites* over **11 distinct helpers**, 9 unconditional — the earlier figure conflated the two), F9 (bound the closure, not just the file: `git archive` + the §3.2.1 `GHJIG_ROOT_OVERRIDE` seam, **with the registry rider** — `in_scope … || exit 0` returns 0 for every input with no `registry.txt`, so an `rc=0` without a control means the guard never ran).
- `453fe5a` **Test** — F-1 (blocking) + F8.

Round 3 of review — the judged list is in the `<!-- finding-judge: round=3 -->` comment. **It is the first round in which the judge found no false claim in the reviewer's verification prose**, and it re-verified round 2's five landed fixes rather than taking them on trust.

- `fa39d47` **Doc** — F-A (blocking) + F-B.

## Checklist

- [x] **Doc** — SPEC §1.5 + §4.11; `code-reviewer.md` + `security-reviewer.md` carry the recipe (mint under `mktemp -d`, every `PATH` executable there, `command -v` residency assertion, hash-pin + re-verify, fresh dir on re-run). (AC1, AC2, AC3, AC4)
- [x] **Test** — §158a/b/c extract the literal command **from the agent contract** and run it K=5 concurrently (count-guard on exactly K outputs, `sort -u` yields K, each path a `0700` directory), plus §158d locking that **both** execution-verifying contracts carry the recipe. *Round-1 correction (F3): the reproduction originally stated here could not produce a RED — the arm's parent tree contains no §158 at all, so that recipe yields silence, which is the silent-skip shape the case comment argues against. The working reproduction is the hybrid (head's smoke file over the pre-Doc contracts) and is now stated in the case comment.* (AC5)
- [x] **Code** — changelog fragment; lint clean, `build_toc.sh --check` exit 0, full smoke `pass=1255 fail=0` in a quiet window (zero live agent worktrees, so §148i-cli measures cleanly).
- [x] **Round-2 F-1 (blocking) — §158a–c now execute *both* contracts' templates.** §158d locked *presence* in both, but §158a–c read `code-reviewer.md` alone. **Control measured, not assumed:** planting the exact escape §158c exists to catch — `<scratch-root>/../ghjig-escapes-XXXXXXXX` — in `security-reviewer.md` scored **`pass=4 fail=0`** against the previous revision. Fully green on a contract instructing its reviewer to escape the scratch root; the copy that could no longer be **deleted** could still silently **rot**. Post-fix the same mutation reds at `158c[security-reviewer]`, and a deleted section reds at both `158a[security-reviewer]` and `158d`. Within the judge's bounds: still exactly those two files, no glob, no new asserted property.
- [x] **Round-3 F-A (blocking) — the F9 seam recipe did not execute as written.** Two defects, both reproduced: **(a)** nothing under `.claude/state/` is tracked (`git ls-tree -r <head> .claude/state/` → 0 files), so `git archive | tar -x` never creates `$S/.claude/state` and the prescribed `printf … > "$S/.claude/state/registry.txt"` exits **`rc=1`** — the recipe stops at step one; **(b)** `in_scope` keys on `$PWD` (`cwd_guard.sh:5-27`) and the recipe never said to create `$S/work` or run from inside it, so a literal follower lands out-of-scope and collects the **vacuous `rc=0`** the rider was added to prevent. Fixed within the judge's ceiling (`mkdir -p`, `cd "$S/work"`) — no new capability, and the `ghjig_state_dir` precedence chain deliberately **not** restated (that would be a second copy of `hookrt.sh` logic, the §9 problem F-4 closed one round earlier). **Verified by executing the corrected bullet verbatim in a fresh directory: `printf rc=0`, `rcA=2`, `rcB=0`.** The judge made that reproduction the condition for correcting the bullet rather than deleting it, decidable by running two commands rather than by argument.
- [x] **Round-3 F-B — the seam had an unstated environment precondition.** The shipped precondition is stated in both contracts: run the probe **with `CLAUDE_PROJECT_DIR` unset**. Precondition only — the resolution chain is deliberately not restated. *(Round 4 deleted a measurement figure that stood here; see the round-4 judged list.)* Promoted from the reviewer's advisory because the same sentence was already being edited for F-A, and shipping a recipe certified "now runnable" with an unstated precondition would repeat this PR's signature defect one more round.
- [x] **Round-2 F8 — the doc-supplied template is prefix-validated before it is executed.** `cd "$S158_ROOT"` bounds a *relative* template only; `mktemp` ignores cwd for an absolute one, and an absolute template created K directories **outside** the test root with §158c naming them only after the fact. The gate requires the literal `<scratch-root>/` prefix; measured **0** directories created outside. §158c is **retained** — the gate does not replace it, and cannot see the relative-`..` escape §158c catches.
- [~] **Follow-up filed — N/A for this PR; split across three subjects, one of which this PR cannot file.** **Filed: #658** — contract agreement + per-role template distinctness + the symlinked-root gap (round-3 F-C, round-4 F4-C), through the `issue-reviewer` gate on the second pass; the first pass caught a false claim in the draft (`mktemp -d`'s `O_EXCL` means identical templates still cannot collide — measured 10/10 distinct — so the real cost is attribution, not collision) and it was re-grounded before filing. **Commented, not filed: #642** — the §148i-cli worktree collision was already covered there mechanism-and-all, so the coverage check said comment, not duplicate; the new datum is that it reproduced twice on this PR and the second `hookrt.sh` counted was the investigating judge's *own* worktree. **Not filed: scratch GC** — still blocked at `/file-issue`'s escalation gate (three consecutive `refine` verdicts → `block` in unattended mode). The third reviewer dissented from its own verdict's consequence — *"my refine is a single missing field, not a verdict that it should be closed out; if the escalation policy converts this to a block, it discards an Issue I judge sound, and a human should override"* — and that override is a human's to make, not an unattended agent's. Draft: `.claude/state/blocked-issue-drafts/scratch-gc.md`; reasoning in `.claude/state/issue-block.log`. Accumulation is observed, not hypothetical: the scratchpad root grew from **222 entries / 645M** to **269 / 738M** over this PR's review rounds, mostly whole-repo materialisations by reviewers pinning artifacts.

## One addition beyond the round-2 judged list (disclosed for the next reviewer to strike)

The judge flagged, explicitly as an observation it would **not** judge: after F-1, both contracts could carry the **identical** template and per-role distinctness would still be unlocked. **I did not close that** — asserting distinctness would be a new asserted property, which the judge's F-1 ceiling ruled out. It is recorded here rather than silently dropped, so the next reviewer can decide whether it belongs in this PR or the follow-up.

What *was* added beyond the literal fix text, and why: the round-2 F-1 comment initially restated its own control measurement in two places in the same file. That is the exact duplication whose staleness produced F-4 one round earlier, so the measurement now lives once (under `MUTATION-CHECKED`) and the loop-head comment points at it. Comment-only; no behaviour change.

## Deferred by the round-3 judged list

- **F-C — nothing locks *agreement* between the two duplicated contract blocks.** They differ in only 4 of 16 lines, and the F9 bullet is byte-identical in both — which is exactly why F-A had to be fixed in two places. §158d locks presence, not agreement. Deferred because locking agreement is a **new asserted property**, explicitly outside round 2's F-1 ceiling. The judge notes it converges with round 2's own unjudged observation (per-role template distinctness is likewise unlocked) and that **one issue should carry both** — and that the scratch-GC follow-up is the wrong home for it.
- **F4-C — a symlinked scratch root defeats the seam recipe**, because `in_scope` compares `pwd -P` against the registry entry as written. Deferred rather than fixed for a reason that is the design working: the bullet mandates the two-arm control, so this fails **loudly** rather than silently. The `pwd -P` normalisation is a third addition, outside the round-3 ceiling. Joins F-C's issue; figures are in the round-4 judged list.
- **F-D — §158's sibling arm §148i-cli reds whenever an agent worktree is live.** Mechanism reproduced by the judge, whose *own* worktree was the second copy counted: agent worktrees are created under `.claude/worktrees/`, which is **inside** `$SHELL_ROOT/.claude` — one of the two directories that arm greps recursively. SPEC §1.5 makes worktree isolation canonical for exactly these agents, so the two contracts collide by design on every PR reviewed this way. Out of scope here (this PR touches neither `hookrt.sh` nor the arm; the red is caused by the reviewing process, not the diff) and a genuine latent defect. 7 recursive-grep sites reference `SHELL_ROOT` across `smoke.d/`; only §148i-cli was **observed** to red, and that scope is transcribed rather than extrapolated to the other six.

## What this PR has NOT demonstrated (§4.9.3)

**No reviewer in this session has actually run under the amended contract.** Claude Code binds `subagent_type` definitions by scanning `.claude/agents/*.md` **at session start** (SPEC §4.9.3), so every reviewer dispatched here carries the blob as of session start, not as of head. A round-2 `security-reviewer` verified this against itself: its injected prompt matches `269323c:.claude/agents/security-reviewer.md` (`cdf80d8a…`) while head is `cda156dd…`.

Consequence, stated rather than glossed: the reviewers who followed scratch discipline during this PR's own review did so because the **dispatch prompts** told them to, not because the contract did. The contract's effect is therefore **unverified by this PR** and will first be exercised in the next session. The §158 arms prove the *recipe* works (the contract's own command, extracted and run K=5 concurrently); they cannot prove an agent reads and follows it.

This is a known and documented binding, not a defect introduced here — but claiming the contract works because reviewers complied this session would be an unverified inference, and #646 exists because of one of those.

## Deferred by the round-1 judged list

- **S-MEDIUM-1(b)** — the pin covers the artifact, not the 15-helper closure that decides the result. The threat model is real and inside this repo (`implementer` is deliberately not worktree-isolated), but #646's AC3 prescribes the one-file pin verbatim, the gap predates this PR, and the PR narrows it. Deferred *with the measurement attached* so the follow-up starts from evidence.
- **S-LOW-c** — "mint fresh every re-run" has no cleanup, and the shell's own out-of-scope destructive matcher **blocks** a reviewer from removing its own minted directory (`rm -rf <own dir>` → rc=2; in-scope control → rc=0). Folds into the scratch-GC follow-up, which must record the matcher interaction, not only the accumulation.
- **S-INFO (residual)** — no enumerated §4.11 high-asymmetry kind covers "the PR edits the reviewer/gate contract itself". Harmless here (additive, advisory), but a future PR that *deletes* the `reviewed-head` pin falls through the identical hole.

## AC closeout note (recorded rather than ticked clean)

**AC1 as written is satisfied** — it is a three-way disjunction and this PR takes limb 2, "a documented convention keyed to the agent id" (`ghjig-<role>-XXXXXXXX`). What is **not** satisfied is the issue's *Notes* preference for a dispatcher-side copy as well. That omission is deliberate and verified: the `finding-judge` "Dispatch shape" line at `ship.md:12` / `review.md:16` ends *"Do not restate the judge's output schema here — `.claude/agents/finding-judge.md` is the contract SSOT (§9)"*, so restating the recipe at five call sites is the drift that precedent forbids.

**AC5** is met only in its paired-arm limb: the arm exercises the *convention's command*, not a dispatch path. Recorded as a stated deviation rather than a re-litigation of the design.

## Test plan

- [x] `pass=1252 fail=0` at `0d5dd99` in a quiet window (zero agent worktrees at launch and finish); `lint OK: 79 files clean (shellcheck v0.11.0)`; `build_toc.sh --check` exit 0; **CLAUDE.md unchanged at 11961 bytes** against the smoke-enforced 12000 gate.
- [x] §158d mutation-verified in isolation before commit: control (both contracts intact) → GREEN, checked 2 of 2; `security-reviewer.md` section deleted → RED naming `security-reviewer.md=headings:0 security-reviewer.md=templates:0`.

## Out of scope

Worktree isolation itself (it stays canonical; this adds an axis beside it). #642 — a different mechanism at a different location (`grep -r` descending into `.claude/worktrees/` repo copies). Re-running PR #635's tally. The §4.11 tally arithmetic, N, majority threshold, or the `reviewed-head:` protocol. Scratch GC (follow-up). `implementer` — its output is a commit verified downstream by tests and reviewers, so contamination surfaces as a failing test rather than a laundered verdict.


ilgyu-yi:
## Finding triage (round 1)



```
reviewed-head: b4a0b734d2a22711b22c14c59e410e61ab04feb9
prior-round:   none (0 canonical markers under the trusted-author filter)
durable:       PR comment, canonical marker round=1 — posted BEFORE any fix
dispatch:      findings from 1x code-reviewer (`ship after fix`) + 1x security-reviewer
               (`refine`), both heads valid. §4.11's N=3 tier NOT fired, and BOTH
               reviewers independently agreed with that call. Judgment from a
               worktree-isolated finding-judge with no shared context.
```

## Must land before merge — seven

**F1 — both new SPEC paragraphs route the reader to §4.5, which is silent on scratch.**
`SPEC.md#§4.5::pointer-dead-end` · **confirmed** — `grep -n -i "scratch|mktemp|scratchpad"` over the head blob: exactly **2 hits in 2848 lines**, both the new paragraphs (346, 1226). §4.5 read in full (1034-1080): none. §4.6 (1081-1090): none. §1.5 ends "The reviewer-side obligation is in §4.5"; §4.11 says "carried operatively by §4.5". Both pointers dead-end.
Axis `implicit↔explicit`, target **interior**. **Floor** — §4.5 may not stay silent. **Ceiling** — it may not restate the recipe: §4.5.1 sets the in-repo precedent verbatim ("lives in one home … SPEC references it, never restates it"), and the PR's own body invokes the same §9 rule for the five dispatch files. Position: a pointer naming `.claude/agents/{code,security}-reviewer.md` "## Scratch discipline (#646)", **in §4.5 and §4.6 both** — §4.6 because `security-reviewer.md` gained the recipe and §4.6 was not touched.

**F2 — the §158 format lock covers only `code-reviewer.md`.**
`70-gates-contentlocks.sh#§158::lock-coverage` · **confirmed** — `S158_AGENT` is a single path; the only `security-reviewer` occurrence in the arm is inside a comment, never an asserted target. Deleting the entire `## Scratch discipline (#646)` section from `security-reviewer.md` leaves `158a ✓ 158b ✓ 158c ✓`, suite identical to unmutated head.
Axis `narrow↔wide`, target **interior**. **Floor** — two files: #646 AC2 names both by name, and §4.6 calls the security reviewer's approve the highest-cost one. **Ceiling** — exactly those two, **not** a glob over `.claude/agents/*.md`: the other contracts carry no scratch recipe by design, so a glob reds on every future agent that legitimately has none.

**F3 — the RED reproduction stated in the §158 case comment cannot produce a RED.**
`70-gates-contentlocks.sh#§158::stated-reproduction` · **confirmed** — the arm is added by `0221157`, a child of `269323c`, so `269323c~1` (= `fbf87cb`) contains no §158 at all and `smoke.sh | grep 158` there yields no `158a/b/c` line. The comment's own claim ("158a fails with `collected 0 of 5`") is unreachable at the tree it names; the recipe yields **silence** — precisely the silent skip the surrounding comment argues against.
Discrete, no axis. The working reproduction is the **hybrid** (head's smoke file over pre-Doc contracts). Two riders the fix must carry: the recipe writes to `/tmp/predoc-646`, an unqualified shared-scratch path **in a comment arguing for scratch discipline**; and the identical defective claim appears a second time in the **PR body's Test checklist item**, so correcting only the comment leaves the false claim standing in the durable artifact.

**F4 — the PR body is stale against its own head (§1.4 / §10.4).**
`PR#657#body-commits-and-checklist::living-doc-sync` · **confirmed** — `commits[]` returns 3; the body lists only `269323c` and leaves Test / Code / Follow-up unchecked; `updatedAt` post-dates the last commit. §4.5 takes the PR body as a reviewer input for checklist verification, so a stale body mis-primes every downstream reader.
Narrowing the reviewer's wording: the commits are pushed, but "green" is the author's to confirm — tick Test/Code only after full smoke + lint + `build_toc.sh --check` at head. **The Follow-up item must stay unticked**: the scratch-GC issue is drafted but blocked at the `/file-issue` escalation gate and is **not filed**.

**F5 — AC1 satisfied only on the reviewer side. Confirmed, with a correction to the framing.**
`issue#646#AC1::ac-satisfaction-claim` · AC1 is a **three-way disjunction** — "Any of: a per-agent path the harness exposes, **a documented convention keyed to the agent id**, or an instruction the dispatcher is required to include." The PR mints `ghjig-<role>-XXXXXXXX`, i.e. **limb 2 of AC1's own disjunction**. The stronger "both places" language is in the issue's **Notes**, not in AC1.
So the closeout record must be accurate: AC1 **as written** is satisfied; what is unmet is the Notes' preference for a dispatcher-side copy, and the reviewer's own verification that restating at five call sites is the §9 drift the PR correctly refuses. AC5's "the dispatch path supplies" is genuinely met only in its paired-arm limb — the arm exercises the convention's command, not a dispatch path. Both belong in `## AC closeout` as stated deviations, not as a re-litigation of the design.

**F6 (converged with S-MEDIUM-2, independently reached) — §4.11 cites "harden after evidence" with no channel through which evidence can arrive.**
`SPEC.md#§4.11::evidence-channel` · **confirmed** — §6.0 P3's own text names this exact gap as the historical problem ("Promotion pressure once left no audit trace — an ignored advisory was invisible") and closes it via the reviewer-reject reason-class trail — which covers **issue-/plan-/activation-reviewer verdicts only**, not code-/security-reviewer, and records **verdicts, not compliance** with an advisory convention. No existing consumer would ingest a scratch compliance report even if one were emitted.
Axis `implicit↔explicit`, target **interior**. **Floor** — SPEC may not invoke "harden after evidence" while leaving the evidence path unnamed. **Ceiling** — **no new reporting instrument ships** (see the closing note). Position: one clause naming the reachable evidence — a further observed collision incident of the PR #635 kind — and recording that no compliance telemetry is collected, deliberately.

**S-MEDIUM-1(a) — the contract's hash-pin wording oversells what the pin covers.**
`{code,security}-reviewer.md#Scratch-discipline::pin-claim-breadth` · **confirmed, reproduced independently** — real `pre_tool_use.sh` under `GHJIG_ROOT_OVERRIDE` + `GHJIG_STATE_DIR_OVERRIDE` with a seeded registry fixture: run A pristine closure → rc=2 (blocked); **one line appended to a sibling helper** (`should_skip() { return 0; }` in `helpers/escape.sh`, *not* the pinned artifact) → run B same payload → **rc=0, guard silently gone**; pinned sha256 identical across both → **"HASH-PIN RE-VERIFY: MATCH"**. Structurally corroborated: `grep -c 'safe_source "$SHELL_ROOT'` → **15** helpers loaded at fire time, none covered by a one-file pin.
Axis `narrow↔wide`, target **low** — narrow the claim to exactly what was measured to hold, justified by measurement (rule 3). **Floor** — the pin **stays**; it is AC3's literally prescribed check and it does detect artifact mutation, so removing it would regress a written AC. **Ceiling** — it may not be presented as *the* contamination detector.
The caveat must do **two** things: bound the pin's scope, **and** stop the bullet order from implying the pin is the detector. The only field evidence that exists — the security reviewer's own live unprompted incident (its T4 run left `escape.sh` neutered; a later probe silently returned rc 0), plus this judge's reproduction — says the **fresh-directory rule** is what catches this and the pin says MATCH throughout.

**F7 (near-free rider on the S-MEDIUM-1(a) edit) — `<scratch-root>` has no stated resolution rule.**
**confirmed** — running the template verbatim: `mktemp: mkdtemp failed on <scratch-root>/…: No such file or directory`, rc=1, no directory left. Fail-closed and loud. Axis `implicit↔explicit`, target **high**. **Floor** — name where the root comes from (the session scratchpad path as given in the agent's prompt). **Ceiling** — do **not** invent a helper or an env var; the PR already measured that none exists. Half a line, landing in the same two files S-MEDIUM-1(a)'s caveat already opens. The weakest of the seven; may drop to defer without harm if a minimal diff is wanted.

## Deferred — three

- **S-MEDIUM-1(b)** — the pin covers the artifact, not the 15-helper closure. The threat model is real and inside this repo (`implementer` is deliberately not worktree-isolated, so it edits `helpers/*.sh` in the shared root while a reviewer exercises a matcher there, and `mktemp -d` does nothing for contamination outside the scratchpad). But #646's AC3 prescribes the one-file pin verbatim, the gap predates this PR, and the PR narrows it. **Defer with the measurement attached** so the follow-up starts from evidence.
- **S-LOW-c** — "mint fresh every re-run" with no cleanup, **and the shell's own matcher blocks the cleanup**: `rm -rf <own minted dir>` → rc=2 "destructive command points outside registry"; in-scope control `rm -rf ./node_modules` → rc=0, proving the matcher was live and discriminating. Folds into the drafted scratch-GC follow-up, which must record the **matcher interaction**, not only the accumulation.
  **Mode note, recorded because it is the exact trap #633 named**: this judge's first attempt measured it through the live Bash tool and got rc=0 — allowed. That was `matches-artifact-mode=no`. No `registry.txt` exists in this environment, so `in_scope` fails and `pre_tool_use.sh:37` exits 0 for every input — *the guard was not allowing the command, it was not running*. The rc=0 refutes nothing.
- **S-INFO(residual)** — no enumerated §4.11 kind covers "the PR edits the reviewer/gate contract itself". Harmless here (additive, advisory), but a future PR that *deletes* the `reviewed-head` pin or edits the fixed-denominator arithmetic falls through the identical hole.

## Confirmed but dropped — retained on the record

**Nothing was refuted.** Four confirmed-and-dropped:

- **S-MEDIUM-2's remedy** (ship an observe-only `scratch:` line) — diagnosis folded into F6; the instrument does not ship. See below.
- **S-LOW-a** (an absolute template escapes the arm's `cd`) — confirmed by construction; `cd` confines relative templates only and 158c's `pwd -P` residency check reds afterwards. Reachable only via a contract edit that itself reds the arm. Free to pre-check *if* the author is already writing F2's loop; not required before merge.
- **S-LOW-b** (`command -v` residency covers only tools you shim) — confirmed as a limitation, but the contract claims no more than it delivers ("Any executable **you put on** `PATH`"). No misstatement to correct. Same-uid PATH writability is the OS trust model, outside #646.
- **S-INFO** ("rejected as inert" reads unconditional) — the qualifier is already in the same sentence. Absorb only if the F6 edit touches that paragraph anyway.

Advisory idiom notes carry no action: the two-attribute format lock is brittle (5 of 6 cosmetic edits red it) but **fails loudly** through 158a's count-guard naming the extracted template — the mitigating design the reviewer credits; and the bare `wait` is currently safe.

## The two weighed calls

**S-MEDIUM-1 vs scope — split it, and the split is the whole ruling.** The *gap* is out of scope: AC3 prescribes the one-file pin literally, the gap predates the PR, and the PR narrows it. The *wording* is in scope, because **this PR writes the sentence**, and a doc-only PR whose deliverable is doc coherence cannot ship a claim its own reviewer measured false in the direction a reader will read it. A **caveat, not a replacement** — deleting the pin would regress a written AC and discard a check that does detect artifact mutation.

**F6 / S-MEDIUM-2 convergence.** Real on the *diagnosis*, confirmed independently against §6.0 P3's text. But the two reviewers asked for **different things**, and that disaggregation decides it: F6 wants a clause naming what evidence would trigger hardening and where it comes from; S-MEDIUM-2 wants the `scratch:` line to ship. Only F6's survives.

The plan-contest reasoning is still right, and there is a **stronger reason available than the one it gave**. S-MEDIUM-2's argument ("an observe-only line has zero wrong-block cost") answers an objection nobody raised — the rejection was never about cost, it was about **inertness**. P3, the very principle S-MEDIUM-2 invokes, supplies the decisive text: *"An unconsumed log is an inert record, not a positive channel; wiring the consumer is what completes the deferred positive face."* Consumer side checked: the reject-reason trail behind `promotion_candidates.sh` records non-pass verdicts from issue-/plan-/activation-reviewer only — not code-/security-reviewer, and verdicts, not compliance. A `scratch:` line would land in **no aggregator** and be precisely the inert record P3 warns against. Its §4.5.1 precedent is disanalogous: the idiom axis is self-observing because its output *is* the finding; a `scratch:` line is a claim *about* an unobserved action, self-reported by the agent whose compliance is in doubt.

F6's ask survives because it needs no instrument to exist — only honesty that none does. And the clause is writable: the evidence that produced #646 was a spontaneous incident report from PR #635 round 4, so **the reachable channel is a recurrence, not telemetry**. Writing that also gives the plan contest's reasoning a durable home, which it currently lacks.

## Scratch-discipline datum — the PR's own contract, applied to this run

Everything ran under a `mktemp -d` directory minted with the contract's own shape; nothing was written to the scratchpad root. Three firsthand observations: the recipe **works as advertised** (K=5 concurrent → 5/5 collected, 5 unique, all mode 700); resolving `<scratch-root>` **required a guess the contract does not supply** (F7); and **cleanup would have been blocked** in a registered project (S-LOW-c). The scratchpad root above this run holds 222 entries, corroborating the GC follow-up's premise.

ilgyu-yi:
## Finding triage (round 2)



```
prior-round: 1@b4a0b734d2a22711b22c14c59e410e61ab04feb9
durable:     PR comment, canonical marker round=2 — posted BEFORE any fix
dispatch:    code-reviewer (`ship after fix`) + security-reviewer (`ship`), both heads
             valid, §4.11 N=3 not fired. Judgment from a worktree-isolated finding-judge.
verdict tokens: untouched. `ship after fix` stays `ship after fix`.
```

## Must land before merge — five

**F-1 — §158d locks presence, not soundness; 158a–c still read `S158_AGENT` (code-reviewer only).**
`70-gates-contentlocks.sh#§158::lock-coverage` · **confirmed, reproduced** — sed the `security-reviewer.md` template to `<scratch-root>/../ghjig-escapes-XXXXXXXX` → `✓158a ✓158b ✓158c ✓158d`, `pass=1248 fail=4`, **byte-identical to the unmutated control**. The case comment names that exact mutation under `MUTATION-CHECKED` as what 158c exists to catch; it catches it in one file and not the other.
Axis `narrow↔wide`, target **interior**. **Inside round 1's ceiling, on a dimension round 1 did not rule**: that ceiling was about the *file set* ("exactly those two, not a glob"), and looping 158a–c over `$S158_AGENTS` does not widen the file set at all — it moves a **different continuum, depth of check per file (presence → execution)**. Justified by measurement, not direction. **Floor** — execute the extracted template from each of the two named files. **Ceiling** — still exactly those two, no glob, and **no new asserted property** beyond 158a–c's existing three (count, distinctness, residency+mode). swing: **none** (interior → interior).

**F-2 — "15 helpers" is a call-site count, not a helper count.**
`{code,security}-reviewer.md#Scratch-discipline::pin-claim-figure` · **confirmed independently** — 15 `safe_source` call sites over **11 distinct files** (`ac_closeout_gate.sh` ×5); 9 unconditional, the other 6 inside the `gh pr merge` path; `hookrt.sh` sourced directly, not via `safe_source`. Transitive helpers (`branch_guard.sh→git_matcher.sh`, `ac_closeout_gate.sh→secret_scan.sh`) are **already inside the 11**.
Axis `fewer↔more`, target **interior**. **Explicitly not directional**: the security reviewer's "the true closure is LARGER than 11" is a direction, and following it would be rule-1 reasoning. **Floor** — at least **11** distinct on any fire. **Ceiling** — at most **12** on a single fire (9 + `hookrt` + `ac_closeout_gate` + `ship_mode`); **no single fire loads 15**. Round 1's bar governs verbatim: *"a doc-only PR whose deliverable is doc coherence cannot ship a claim its own reviewer measured false in the direction a reader will read it."* The caveat's substance ("none of them the pinned file") is unaffected.

**F-3 — `SPEC.md:1071` says "lives in one home" and then names two files.**
`SPEC.md#§4.5::one-home-phrase` · **confirmed** — the phrase is lifted from `:1077`, where it genuinely names one file per language. Discrete, no axis. Free rider: F-4 forces an edit to this same sentence.

**F-4 — the §4.5 index restates the pin in the unbounded form the contracts just corrected.**
`SPEC.md#§4.5::pointer-dead-end` · **confirmed, and stronger than reported** — the index enumerates "…pin the artifact under test and re-verify after the run…" while `code-reviewer.md:56` now reads "It detects mutation of that file's own bytes, and nothing else." The index also lists the pin **before** the fresh-directory rule, inverting the priority the contract just established.
`remedy-survives: **no**`. Axis `implicit↔explicit`, target **interior**.
**Drop the enumeration; do not sync it.** Syncing keeps a second copy to hand-sync — the drift the finding itself names — and it is **outside round 1's own stated ceiling** on this exact axis-key (round 1 F1: *"it may not restate the recipe"*, citing §4.5.1's "SPEC references it, never restates it"). **The landed F1 fix shipped a five-item restatement and has already gone stale inside one PR. That is the measurement that settles it.**
**Floor** — round 1's F1 floor still binds: §4.5 may not go silent; the pointer naming the two contracts and the exact heading **stays**. **Ceiling** — the recipe enumeration goes.
swing: **none** — a *return inside the round-1 band* (the landed fix overshot toward `explicit`), not a move to the opposite end. An overshoot correction is not a swing, checked on the key rather than assumed.

**F8 — §158 executes a doc-supplied template with no prefix validation.**
`70-gates-contentlocks.sh#§158::template-prefix-validation` · **confirmed, reproduced** — an ABSOLUTE template → `✓158a ✓158b ✗158c ✓158d`, `pass=1247 fail=5`, with **five directories created OUTSIDE the test root** and 158c naming them only *after* the side effect. The subshell's `cd` confines relative templates only; `mktemp` ignores cwd for an absolute one.
Axis `permissive↔strict`, target **high** — an extreme position justified independently on two grounds: (i) **definitional** — the contract's template *must* begin with `<scratch-root>/`, so a prefix reject rejects nothing legitimate; (ii) **§6.0 P1** — a wrong reject reds loudly and names the template (one cheap re-run), a wrong accept executes doc-supplied filesystem writes at an arbitrary absolute path. **Floor** — validate the prefix before execution. **Ceiling** — prefix only, no further shape assertions, and **158c is retained**; the pre-check does not replace it.
swing: **none — this is a promotion on round 1's own stated trigger, not a reversal.** Round 1 held it as S-LOW-a with the condition written out: *"Free to pre-check **if** the author is already writing F2's loop; not required before merge."* F-1's fix is that loop. The condition is now met.

## Sixth — real, and the one safe drop

**F9 — the pin bullet leaves a diagnosis with no prescription.**
`{code,security}-reviewer.md#Scratch-discipline::prescription-gap` · **confirmed** · axis `implicit↔explicit`, target **interior**.
**The new evidence promotes the *sentence*; the *gap* stays deferred.** Round 1's deferral rested on (a) "a new capability, out of scope" and (b) "#646's AC3 prescribes the one-file pin verbatim". F9's measurement — three commands, zero new shell code, over the already-documented §3.2.1 seam — **refutes (a) on evidence**. (b) survives untouched, but (b) was never a bar to *prose*: round 1 itself drew exactly this split for S-MEDIUM-1(a) — *"the gap is out of scope … the wording is in scope, because this PR writes the sentence."* Applying the same split here is consistency, not a flip. **S-MEDIUM-1(b) remains deferred as the capability item.**
**Floor** — name one actionable move for the closure the bullet just declared uncovered. **Ceiling** — at most one clause plus a short recipe; stays **advisory**; ships **no helper and no new capability**; must not become a second recipe home that re-opens the §9 problem F-4 is closing.
**Rider the prescription must carry**: the registry seeding *and* the control run (present → rc=2, removed → rc=0). Without it the recipe prescribes a probe where `in_scope || exit 0` returns 0 for every input — the guard **not running**, not allowing. That is the #633 trap, and prescribing it unguarded would be worse than the missing sentence.
**This is the weakest of the six.** F-2/F-3/F-4 correct things that are *wrong*; F8 closes a behavioural gap in new code; F9 improves *helpfulness*. Its own reviewer calls it non-blocking. If a minimal diff is wanted, drop it to defer without harm.

## Confirmed but dropped — retained on the record

- **The `promotion_candidates.sh` consumer clause.** The reviewer's correction is **factually right** — the jq filters on `.event/.decision/class=` and `group_by([.category,.cls])` with **no category whitelist**, so the restriction is the *emitters'*, not the script's. Dropped anyway: the sentence's subject is *"the trail"*, and "that `promotion_candidates.sh` aggregates" is a restrictive clause identifying which trail, not a claim the script enforces it. Retained **with the measurement** so a future PR leaning on it as a design constraint starts from evidence rather than from the sentence.
- **The AC1 residual** — the tension between AC1's lead clause ("where the caller can be held to it") and `SPEC.md:346` ("the agent applies to itself") is real and non-changing: limbs 1 and 2 are both inherently non-caller-side, so the disjunction governs, and the operative half (*not left to naming discretion*) is satisfied by a literal template.
- **Idiom (advisory).** The sharpest observation: **F-1 is itself a confusion between `S158_AGENT` and `S158_AGENTS`**, two identifiers differing by one character with different scopes. No separate action — under F-1's fix `S158_AGENT` most likely **disappears entirely**, dissolving the hazard at the source. The unquoted `for` word-split is genuinely advisory: a space in `$SHELL_ROOT` produces a wrong path that 158d reports **loudly** as `=ABSENT`, not silently.
- **F10 — the round-2 security reviewer ran the round-1 contract.** Blob OIDs exact (`cda156dd…` head vs `cdf80d8a…` at `269323c`). Confirmed; **no artifact change follows** — the §4.9.3 session-restart caveat operating as documented, and both versions prescribe the pin. Already recorded in the PR body. Not `defer-to-issue`: a known documented caveat, not a defect.

**Nothing was refuted this round.**

## Standing carry-overs from round 1 — still open

- **F5's `## AC closeout` deviations** (AC1 satisfied *as written* via limb 2; the Notes' dispatcher-side preference unmet; AC5 met only in its paired-arm limb) must land as *stated deviations* at closeout.
- **F4's Follow-up checkbox stays unticked** — the scratch-GC issue is drafted, blocked at the `/file-issue` escalation gate, **still not filed**. It must record the **matcher interaction** (S-LOW-c: `rm -rf <own minted dir>` → rc=2), not only the accumulation.

## Observations — not judged records, no action assigned

- **A gap adjacent to F-1 that no reviewer raised.** After F-1's fix, both contracts could carry the **identical** template: 158a–c would pass per file (5 distinct *within* each run) and 158d only counts one heading and one template per file. **Per-role distinctness — the property that stops two different agents colliding, which is the whole point of the convention — would remain unlocked.** Flagged for the author to weigh while writing that block; adding findings is not the judge's to do.
- **Scratch discipline, applied to this run.** Everything under a private `mktemp -d`; nothing at the scratchpad root. Two honest notes: the judge **reused** one tree for the F-1 control and mutation before minting a fresh one — a slip against the contract's own fresh-directory rule, disclosed rather than hidden; and F8's absolute-path escape wrote its 5 directories inside the minted dir **only because the mutation was aimed there** — aimed at the root it would have landed in the root.
- **Pollution datum.** Scratchpad root: **222** at round 1, **228** mid-round-2, **240** now — roughly +18 during this PR's review cycle. The GC follow-up's premise keeps strengthening while the issue stays unfiled.

ilgyu-yi:
## Finding triage (round 3)

```
reviewed-head: 453fe5ae42a5077456198cf0b735c3026b66713b
prior-round:   2@0d5dd990a4840aeb2bcedd678b5c1cff99b306b0
dispatch:      findings produced by a worktree-isolated `code-reviewer` at 453fe5ae;
               judgment by a separate, fresh, worktree-isolated `finding-judge` with
               no shared context. Limb (a) recorded by the caller, per §4.13.
```

**Round 3 is the first round in which the judge found no false claim in the reviewer's verification prose.** Every load-bearing figure was re-measured and held. That changes the weighting: this reviewer is transcribing rather than reasoning, so its one must-fix earns a lower burden of proof than prior rounds'.

### Must land before merge — one

**F-A — the F9 recipe cannot be executed as written.** `{code,security}-reviewer.md#Scratch-discipline::prescription-gap` · **confirmed, both defects reproduced.**

- `git ls-tree -r 453fe5ae --name-only .claude/state/` → **0 tracked files**, so `git archive | tar -x` never creates `$S/.claude/state`, and the prescribed `printf … > "$S/.claude/state/registry.txt"` exits **rc=1**. Defect (a), unconditional, stops the recipe at step one.
- `in_scope` keys on `$PWD` (`cwd_guard.sh:5-27`) and the recipe never says to create `$S/work` or run from inside it → `rc_nocd=0`. Defect (b).
- The corrected form was measured end-to-end: **`rcA=2`** (registry seeded, cwd `$S/work`) / **`rcB=0`** (registry removed).

`action: fix-now` · axis `implicit↔explicit`, target **interior** · `swing: none` — round 2 set this key at the same axis/target; round 3 moves further along the same direction, not to the opposite end.

**Floor** — the recipe must execute as written. **Ceiling** — exactly two additions (`mkdir -p "$S/.claude/state" "$S/work"` before the `printf`, and `cd "$S/work"`), **no** new capability, and **no** restatement of the `ghjig_state_dir` precedence chain — that would be a second copy of `hookrt.sh` logic, the §9 problem F-4 closed one round ago.

Two narrowings of the reviewer's wording, neither a refutation:

1. The reviewer calls defect (b) "the precise false pass the rider was added to prevent." **The prescribed control actually catches it** — with the wrong cwd, arm A returns 0 where the recipe says it must block, a visible failure rather than a silent pass. The silent-pass exposure needs a reviewer who ignores the `rc=1` *and* skips the control. Defect (a) alone is dispositive.
2. It applies to **both files identically** — the two blocks differ in only 4 of 16 lines and the F9 bullet is not among them (`35e9e703e63d…` both).

**No §158 interaction**: 158a/158d extract on the anchored whole-line form, which an inline-backticked `mkdir`/`cd` in a bullet cannot satisfy.

### Fix now, as a free rider on that same sentence — one

**F-B — the seam recipe has an unstated environment precondition.** `…::seam-env-precondition` · **confirmed, measured**: with `CLAUDE_PROJECT_DIR` set (cwd correct, registry seeded) → **`rc_cpd_set=0`**; the seeded registry is bypassed and the guard never runs. Mechanism: `hookrt.sh:90` prefers `$CLAUDE_PROJECT_DIR/.claude/ghjig-state`; `cwd_guard.sh` falls back to the seeded path only if that is absent.

`action: fix-now`. **Floor** — state the precondition ("with `CLAUDE_PROJECT_DIR` unset"). **Ceiling** — the precondition **only**; do not explain the resolution chain. Promoted from advisory because the author is already editing this exact sentence in both files for F-A, and shipping a recipe certified "now runnable" with an unstated precondition would repeat this PR's signature defect one more round.

### Defer to issue — two

**F-C — the duplicated block has nothing locking agreement.** Confirmed (`diff` = 4 differing lines of 16; the F9 bullet byte-identical, which is *why* F-A is duplicated). Locking agreement is a **new asserted property**, explicitly outside round 2's F-1 ceiling. Converges with round 2's own unjudged observation (per-role distinctness unlocked) — **one issue should carry both**; the scratch-GC follow-up is the wrong home.

**F-D — §148i-cli reds in the presence of a live agent worktree.** Confirmed and reproduced: the arm's own grep returns 2 files — `hookrt.sh` and `.claude/worktrees/agent-…/.claude/hooks/hookrt.sh`, the second copy being **the judge invocation's own worktree**. Out of scope here (this PR touches neither `hookrt.sh` nor the arm; the red is caused by the reviewing process, not the diff) **and** a genuine latent defect: SPEC §1.5 makes worktree isolation canonical for these agents while the arm greps `$SHELL_ROOT/.claude` recursively — the two contracts collide by design on every PR reviewed this way. 7 recursive-grep sites reference `SHELL_ROOT` across `smoke.d/`; only §148i-cli was observed to red, and that scope is transcribed rather than extrapolated.

### Confirmed but dropped — retained on the record

- **93-line loop body at column 0** (`70-gates-contentlocks.sh:2938-3031`) — confirmed; dropped as scope creep on a doc PR.
- **`[ "$s158d_n" -eq 2 ]` hardcodes cardinality** (`:3056`) — confirmed; a third contract yields `checked 3 of 2`. Dropped: **fail-safe direction, reds loudly**. `${#S158_ALL_AGENTS[@]}` is free to take if F-C's issue later touches this arm.
- SC2016/SC2031 — info-level noise / false positive. No action.

**Nothing was refuted this round.**

### The one weighed call

**Correct the F9 recipe, do not delete it.** Deletion is the opposite end of this key's axis, and the predecessor PR's constraint (2) — *"delete unmeasured claims rather than correct them"* — is the strongest argument for it. It does not apply: the F9 bullet is not an unmeasured claim reasoned to at a desk, it is a **procedure whose corrected form was measured end-to-end**. Correcting it is transcription.

**The fallback is written into the instruction, so the author is not left to reason:** reproduce the two-arm control (`rcA=2`, `rcB=0`) after editing. **If it does not reproduce, delete the F9 bullet rather than ship a third revision of it.**

### Observations — not judged records, no action assigned

- **Three items now want the unfiled follow-up** (round 1's S-LOW-c scratch-GC + matcher interaction; F-C's duplication/per-role distinctness; F-D's worktree collision). Three different subjects. The carry-over binds: the Follow-up checkbox stays unticked until actually filed, and the `/file-issue` escalation gate is still the blocker.
- **Scratch discipline, applied to the judge's own run** — everything under a private `mktemp -d`; nothing written to the scratchpad root. Honest note, disclosed rather than hidden: it **reused** one tree across three probes instead of minting fresh per re-run, a slip against the contract's own rule. Mitigating — each probe's discriminating result was a different rc from the same tree, and the registry was restored and re-verified between arms.
- Round 2's five landed fixes were re-verified rather than taken on trust; all hold.

ilgyu-yi:
## Finding triage (round 4)

```
reviewed-head: fa39d4754ed7f70c2909cb1c88fab8ec9359fd95
prior-round:   unresolved at judgment time — the round-3 list was posted under a
               non-canonical header (marker present, but the first line read
               "## Judged findings — round 3" rather than the canonical
               "## Finding triage (round N)"). Highest canonical was 2@0d5dd990.
               REPAIRED AFTER judgment: the round-3 comment's header has been
               corrected in place, so canonical count is now 3 and the next
               derivation is 4 — no collision. The judgment below was reached
               WITHOUT that repair and is left exactly as delivered.
dispatch:      findings by a worktree-isolated `code-reviewer` at fa39d475; judgment by a
```

Because the prior round was unresolvable at judgment time, **every `fix-now` record below carries `swing: not-evaluated`, not `swing: none`** — a `swing: none` over an unresolvable prior round is the vacuous pass the rules forbid. The substantive swing analysis is given in prose instead. (Round 3's `head=453fe5ae` was corroborated as an ancestor of the current head; corroboration does not restore canonical shape.)

### The swing question, answered explicitly: **genuine correction, not a swing**

1. **Different site, therefore a different axis key.** Round 3's F-B ruled on the **shipped contract text**; round 4's finding A rules on the **PR body's AC checkbox**. Cross-round matching requires the site half.
2. **Round 3's remedy is measurably still correct and round 4 does not touch it.** The "with `CLAUDE_PROJECT_DIR` unset" direction yields `rc=2` (guard runs) under all four configurations. No removal, no weakening, no demotion is asked.
3. **Round 3's own mechanism sentence already carried the condition.** What it stated loosely was its **headline figure** (`rc_cpd_set=0`, given without its condition). Round 4 corrects the figure in the direction round 3's prose already implied — transcription catching up.
4. **Same direction, not opposite** — both move toward explicit.

**Where round 4 does bite round 3**: round 3's unconditioned figure propagated verbatim into the PR body *and* the commit message. That is a real defect in round 3's summary figure — but correcting a figure is not occupying the opposite end of an axis.

### Verification run

Mechanism read at the head blob: `ghjig_state_dir` prefers `$CLAUDE_PROJECT_DIR/.claude/ghjig-state`; `cwd_guard.sh` carries the back-compat fallback `[ -f "$registry" ] || registry="${GHJIG_ROOT:-}/.claude/state/registry.txt"` — exactly the path the recipe seeds. So the bypass requires the per-project `registry.txt` to **exist**.

| arm | `CLAUDE_PROJECT_DIR` | rc |
|---|---|---|
| A | unset (control) | **2** (blocks, with the out-of-registry message) |
| B | `= $S` (materialised tree) | **2** |
| C | `=` the real repo | **2** |
| D | `=` a dir carrying `.claude/ghjig-state/registry.txt` | **0** |
| G | unset, scratch root minted via the `/tmp` symlink | **0** |

`find /Users/onedgelee/ghjig -name registry.txt` → **0 hits**. The reviewer's table reproduces exactly, arm for arm.

### Judged records

**F4-A — the PR body states the F-B bypass unconditionally; it is conditional on `$CLAUDE_PROJECT_DIR/.claude/ghjig-state/registry.txt` existing.**
`PR#657-body#AC-item-Round-3-F-B::overstated-measurement` · **confirmed** (evidence: the five-arm table above, driven as a script file — the artifact's own mode).

`action: fix-now` · `remedy-survives: **no**` · axis `fewer↔more`, target **interior** · `swing: not-evaluated`.

**The reviewer's "qualify the measurement" does not survive; the claim should be DELETED, not corrected.** Independent ground, not directional: (i) the figure is **load-bearing for nothing** — measured, the shipped contract text is correct under all four configurations, so neither the fix nor its license depends on the rc figure; (ii) qualifying it requires restating the `ghjig_state_dir` → `cwd_guard` fallback chain in prose, which is net-new mechanism text in a document that has now failed prose-accuracy review four rounds running — **correction manufactures the next round's finding**; (iii) SPEC §6.0 P1 cost-asymmetry — a retained wrong claim is durable and is read back by later rounds as fact (**demonstrated in this very PR**), while a deleted decorative figure costs one non-load-bearing detail.

**Floor** — the AC item must still record that F-B was fixed and what the shipped precondition is; do not delete the checkbox. **Ceiling** — delete only the unconditioned measurement clause; add no mechanism prose, no resolution-chain restatement, **no replacement figure**. The reviewer's "contract text needs no change" survives, confirmed by arms A–D.

**F4-B — "Fixed with exactly the two additions the judge's ceiling allows" is checkable and false at the byte level.**
`PR#657-body#AC-item-Round-3-F-A::overstated-scope-claim` · **confirmed** — beyond the two additions the line also gained three rationale clauses and the `rc=2`/`rc=0` annotations. **Substance is inside round 3's ceiling** (no new capability; precedence chain not restated, verified against `hookrt.sh` at head); the word "exactly" is what is false.

`action: fix-now` · `remedy-survives: yes` · axis `fewer↔more`, target **interior** · `swing: not-evaluated`.

Narrowed, not refuted — the reviewer offered "drop or soften"; **only the DROP arm survives.** Softening authors a replacement claim, which is new prose and therefore new failure surface, in a PR whose last three rounds were all prose-accuracy defects. Deleting one word adds zero surface and is the loop-terminating option. It also rides the adjacent lines already being edited for F4-A.

**Floor** — the claim must stop asserting a falsifiable count. **Ceiling** — delete the word; do **not** enumerate what else changed, and do **not** substitute a corrected count — an enumeration is a new checkable claim and re-opens this key.

**F4-C — a symlinked scratch root silently defeats the recipe.**
`.claude/agents/code-reviewer.md#Scratch-discipline::symlinked-root` · **confirmed** — `in_scope` compares `pwd -P` against the registry entry as written; minting `$S` through the `/tmp` → `/private/tmp` symlink gives `rcA=0` where the physical-path run gives `rcA=2`.

`action: defer-to-issue`. Judged rather than passed through unmeasured — it is a real, measured claim and earns a disposition. Defer is right because the prescribed two-arm control makes it **fail loudly** (a reviewer sees `rcA=0` and knows the guard never ran), and the fix (`pwd -P` normalisation) is a third addition outside round 3's ceiling. Belongs with F-C in the deferred issue.

### On the correction-loop risk

The answer is not "stop reviewing prose" — it is **make both remedies deletions**. Both `fix-now` items are net-negative in bytes and add no new assertion. That is the property that terminates the loop: **a round whose fixes only remove checkable claims cannot generate a round-5 finding about the text it wrote, because it wrote none.**

This is also where round 3's reasoning correctly did *not* apply and round 4's does. Round 3 declined deletion because the F9 bullet was a **procedure whose corrected form was measured end-to-end** — correcting it was transcription. That does not transfer: F4-A's target is not a procedure, it is a **decorative figure that over-generalises its own measurement**, and correcting it would require reasoning at a desk about a resolution chain.

### Observations — not judged records, no action assigned

- **A round-numbering collision was scheduled** by the non-canonical round-3 header (canonical count 2 → the next judge would derive 3 and collide). The judge flagged the cheap repair and explicitly did **not** apply it, leaving it to the caller. It has since been applied: the round-3 comment's first line now reads `## Finding triage (round 3)`.
- **The commit message is better than the PR body here** and cannot be edited anyway (already pushed): it carries the conditioning clause, though its own `Measured rc=0` is still unconditioned. The remedy correctly targets the PR body only.
- **Scratch discipline, applied to the judge's own run**: a fresh dir minted for this invocation, not reused across probes; read-only git only. Disclosed slip: the worktree-isolation guard refused a heredoc-written probe script, so the arms ran as separate commands rather than one pinned script — more auditable in effect, but not the single-script form the recipe prescribes.

ilgyu-yi (commented):
**Verdict: approve** — `code-reviewer` returned `ship` at the confirmed head after remediation of its one `ship after fix` finding.

The reviewer verified by execution rather than by reading. The `## Scratch discipline (#646)` recipe was run verbatim from a fresh directory (`printf rc=0`, `rcA=2` with the registry seeded and the correct cwd, `rcB=0` with it removed, and `rc=0` when run from the wrong directory — confirming both round-3 defects are genuinely fixed). The contract's central claim reproduced: appending one line to a *sibling* helper flipped a real guard from BLOCK to ALLOW while the pinned artifact's hash reported MATCH. Every `MUTATION-CHECKED` claim in the §158 case comment red as documented across five fresh-directory mutations, including the escape planted in `security-reviewer.md` that scored fully green before this PR. All counted figures check out (15 `safe_source` call sites over 11 distinct helpers, 1255 total assertions, 11961 bytes of `CLAUDE.md`). Scope is five on-subject files, Doc → Test → Code ordering holds, CI green on both platforms.

The one finding was a durable-record honesty gap, not a code defect: `scripts/ac_closeout.sh` re-emits AC lines verbatim with no deviation capability, so it ticked issue #646's AC1 and AC5 clean and silently overwrote a `finding-judge` instruction from round 1 (carried by round 2) that both must land as **stated deviations**. AC5 as written claims the dispatch path supplies a distinct scratch path; this PR deliberately supplies nothing dispatch-side, so a future reader of #646 alone would have inferred a mechanism that does not exist. Remediated by appending both deviations to that comment without altering the ticks; the reviewer re-fetched the live comment, confirmed the inference is now explicitly foreclosed, and confirmed the §6.1 `ac-closeout` gate header still matches. Its factual audit of the appended text found one misattributed review round, which has since been deleted.

Four advisory idiom notes are recorded and none escalates (SPEC §4.5.1). Follow-ups are filed: #658 (contract agreement, per-role template distinctness, symlinked-root gap) and a comment on #642 (the §148i-cli worktree collision, already covered there). The scratch-GC issue remains deliberately unfiled — blocked at `/file-issue`'s escalation gate, and overriding that gate is a human's call, not an unattended agent's.

Issue #383, opened 2026-06-18
fix: commit-subject extractor mis-detects heredoc-form on a bare <<TAG token in a normal -m body


ilgyu-yi:
## Problem
`extract_commit_subject` (`.claude/hooks/helpers/conventional_commit.sh`) detects the heredoc message form `-m "$(cat <<TAG … TAG)"` by searching the **entire** `-m` value for `<<-?\s*['"]?([A-Za-z_]\w*)` (python path ~line 61; legacy awk path ~line 89). The detector fires on **any** value containing a `<<TAG` token — including ordinary prose in a multi-line `-m` body. So a commit whose line-1 subject is a valid `<type>(#N): …` but whose body merely *mentions* `<<EOF` (common when the change is about shell/heredocs) is mis-parsed: a body line is returned as the "subject", fails the Conventional-Commit format check, and the commit is **blocked**.

Observed live in PR #380: a valid `fix(#375): …` commit was blocked because its body described the bypass-suspect heredoc matcher (literal `<<EOF` tokens). Workaround used at the time: reword the body to avoid the `<<` token. Promoted from discussion #382.

## Rationale
- **(a) MISSION fit**: §1.3 active SSOT + §6.0 enforcement-style. The conventional-commit gate is a hard *block*; a false-positive over-block fires on a **correct** commit, walling in legitimate work (the gate fails safe by over-blocking, but the cost is real — it forces an awkward reword of valid commit prose).
- **(b) why-now**: surfaced live in PR #380; recurs whenever a commit body discusses shell/heredocs — frequent in this repo's own hook work.
- **(c) existing-coverage**: #367 (closed) fixed adjacent extraction shapes (multi-`-m`, embedded-newline single-`-m`, sibling-command heredoc in a compound invocation) but NOT a bare `<<TAG` appearing as **prose inside a normal `-m` body** (no `$(cat <<TAG …)` command substitution). No open issue.

## Acceptance criteria
- [ ] `extract_commit_subject` returns the **line-1 subject** for a single `-m` whose body merely contains `<<TAG` prose and whose line-1 is already a valid Conventional Commit subject — both the python primary and the legacy awk fallback paths.
- [ ] The genuine `-m "$(cat <<TAG … TAG)"` heredoc-substitution form still extracts the heredoc body subject (no regression; #367's cases stay green).
- [ ] A genuinely malformed subject still fails the format check (no security regression — the gate still fails safe).
[...]

PR #659, opened 2026-08-15
fix(#654): fail closed when subject length cannot be measured

### LINKED ISSUE #654: the commit-subject length gate reports valid when its own length computation fails
## What

`.claude/hooks/helpers/conventional_commit.sh` computes a subject length and then gates on it without checking that the computation produced a number:

```
len=$(_codepoint_len "$rest")
if [ "$len" -lt 1 ] || [ "$len" -gt 72 ]; then   # <- both tests error if $len is empty
```

`_codepoint_len` prefers `python3` and falls back to a shell path. If `python3` is **present but broken** — on `PATH`, executable, exits non-zero — the command substitution yields an empty string.

With `len=""` under `set -uo pipefail`, both `[` tests error with "integer expression expected", each returns a non-zero status, the `||` chain therefore evaluates **false**, and the `if` body is skipped. The subject-length gate reports **valid**.

That is the fail-open direction, on a gate.

## Measured

Script-file mode throughout, never `bash -c`.

**Reachability.** Sourced the real `hookrt.sh` + `conventional_commit.sh`, with a `python3` shim early on `PATH` that exits 1:

```
_codepoint_len "hello"  ->  len=[]        (control, real python3 -> len=[5])
```

**Consequence.** Ran the exact `if [ "$len" -lt 1 ] || [ "$len" -gt 72 ]` construct under `set -uo pipefail`:

| `len` | result |
|---|---|
| `5` | accept (control) |
| `99` | reject (control) |
| `""` | both tests error; **the gate accepts** |

Both halves close. This was originally raised as an unverified pointer by a `security-reviewer` on PR #651 and left explicitly unexecuted; `finding-judge` executed both halves and upgraded it from pointer to confirmed.

## Why

**Why now:** #647 is about exactly this shape — an operand reaching a comparison without a guarantee that it is a number — and its own review surfaced this instance in a *different* helper, on a *gate* rather than an advisory. #647 correctly declined to widen into it. Left unfiled it is an unowned deferral, which is the failure #647 was itself filed to correct for #635.

**MISSION fit:** "The mechanism" → the enforcement dual and its cost-asymmetry rule. `MISSION.md` reserves the hard negative face for wrong-allows that are expensive, and the commit-subject gate is one: a malformed subject that passes propagates into durable git history, which is exactly the class the shell exists to keep clean. A gate that silently reports "valid" when its own measurement failed is a wrong-allow with no signal.

**Contrast with the sibling issue on `stop.sh`:** that one is advisory-only and correctly gets a cheap sanitizer. This one is a **gate**, so the cost of being wrong is higher and the fix should fail closed rather than fall back to a permissive default.

**Existing coverage:** none. #647 covers the two TTL clock operands and names three cache-TTL `date +%s` sites out of scope. The `stop.sh` counter is filed separately as an advisory-tier instance.

## Acceptance criteria

- [ ] When `_codepoint_len` cannot produce a length, the subject-length check **fails closed** — it refuses rather than reporting valid. A gate that cannot measure must not approve.
- [ ] The failure names its cause, so an operator with a broken `python3` sees why the commit was refused rather than a bare rejection (SPEC §6.0 P4).
- [ ] A smoke arm covers it, two-sided: with the length computation broken the gate refuses, and with it working a valid subject is still accepted and an over-length subject is still refused.
- [ ] The shell fallback path inside `_codepoint_len` is checked for the same shape — if it can also emit empty, the guard must cover both paths, not just the `python3` one.

## Out of scope

- Rewriting `_codepoint_len`'s measurement strategy or removing the `python3` preference. The defect is the missing guard on its output, not the choice of tool.
- The 72-codepoint limit itself.
- Any other operand in the hook tree. The `date +%s` cache-TTL sites are recorded out of scope on #647 with their reason, and the `stop.sh` counter is filed separately.

ilgyu-yi:
Closes #654

## Target base

`main`

## Why

`check_commit_subject` computes a length and then gates on it without checking that the computation produced a number. When it did not, both `[` tests error, the `||` chain evaluates false, the `if` body is skipped, and **the gate reports the subject valid** — a fail-open on a gate whose wrong-allow lands in durable git history.

Reproduced in script-file mode under `set -uo pipefail` (the artifact's real mode):

| `len` | result |
|---|---|
| `5` | ACCEPT (control) |
| `300` | REJECT (control) |
| `""` | **ACCEPT** — the fail-open |

## Plan

Contest of three candidates; two mutually-blind challengers on `robustness/risk` and `UX`. Both returned **dominates-A**, on independent and complementary grounds, and the `plan-reviewer` shipped A's structural core with seven grafts — improving on both challengers rather than picking one.

**A's core, which survived:** the defect is *not* a `python3`-branch property — measured, stripping `wc`/`tr` from `PATH` makes the shell branch fail identically. It is `check_commit_subject` consuming an unvalidated value at the comparison. So the guard belongs at the **consumption site**, which covers every producer path and all four call sites by construction rather than by four parallel edits.

**What the challengers overturned:**

- **A's central soundness claim was false.** A argued "a positive match cannot itself be defeated by a bad operand". Measured against an A-faithful implementation, `9223372036854775808` and `99999999999999999999999999` are all-digit, pass `case ''|*[!0-9]*`, then error in `[` exactly like the empty string → **ACCEPT**. The true domain of `[ -lt ]` is *`intmax_t`-representable* decimals, not "digit strings". The fail-open survived A's fix.
- **A's fallback newly activates byte-counting on the very host it targets.** `_codepoint_len` selects its branch with `command -v python3` — by *presence*, not success — so on a present-but-broken host the `wc -m` path is unreachable today. A's "not newly introduced" was inaccurate.
- **A's refusal message named a dead escape.** `should_skip` occurs **0 times** in `.githooks/commit-msg` and `.githooks/_lib.sh`, so `SKIP_HOOKS=commit-format` does nothing at the git-hook tier (§6.7 folds it to `--no-verify`); and per §7 the live Bash tool strips the env-prefix, so it is wrong for the agent too.

**Judge's refinements on the challengers themselves:** saturate the *comparison* operand but report the **raw** length (the challenger's line would print `got 999` for a real 5000-codepoint subject); and **relocate** the byte-count discriminator into `_codepoint_len`'s fallback, because at the message site it ignores `len` and would misname a genuine over-length as unmeasurable.

## Checklist

- [x] **Doc** — SPEC: the measurement contract (total function, ladder keyed on **output validity** not exit status, measurability guard, refuse-never-approve with the `intmax_t` domain named); the fail-policy split (*helper missing* stays fail-open, *helper present but unable to measure* is fail-closed); §6.0 P4 recovery is **arm-scoped** — the shared helper emits cause only, each adapter appends its own live recovery; and the #383 line-1-precedence blast radius named rather than left to be discovered. Changelog fragment added.
- [x] **Test** (`66707da`) — arms `16p`–`16t` plus a `16-fixture` setup guard, sited next to `16n`/`16o` and reusing their curated-`PATH` pattern. All five confirmed **red for the right reason** before Code (each carries a distinct rc so a fixture miss cannot masquerade as the intended failure). The test author additionally ran a *satisfiability* check (a throwaway sketch against a scratchpad copy → 6/6 green, so the arms are not impossible) and an *anti-vacuity* mutation (disabling the `python3` shim flips the arms to "shim never ran" rather than greening). That guard caught a real fixture bug during authoring: `ln -sf` had made the shim a symlink, so the heredoc wrote *through* it and the real `wc` answered.
- [x] **Code** (`f99bed4`) — `_codepoint_len` total function with validity-keyed rung advancement + the measurability guard; positive decimal test and saturated-operand/raw-report comparison in `check_commit_subject`; per-adapter recovery text at the three operator-facing surfaces.

### Verification (quiet window, zero live agent worktrees)

`smoke: pass=1261 fail=0` · `lint OK: 79 files clean` · `build_toc.sh --check` rc 0.

The quiet window matters: `§148i-cli` counts `hookrt.sh` copies under `$SHELL_ROOT/.claude`, and agent worktrees live there (#642, parked). With two live it reported `count=3`; with none it reports 1. Measured both ways rather than assumed — and the two worktrees were removed via `git worktree remove` only after confirming each one's HEAD was already an ancestor of this branch, never with `rm -rf`.

- `1f56da5` **Doc** — F1: `docs/TROUBLESHOOTING.md` indexed the sibling message from the same function but not the new one, which is the harder of the two to diagnose unaided. Landed inside the judge's bounds — **floor**: name an actionable *environment* repair, because measured, no surface an operator sees today names one (every offered recovery is escape-the-gate or edit-the-subject, none repairs the cause); **ceiling**: one row at sibling scale, pointing at SPEC §6.1.1 rather than paraphrasing it, so the §9 thin-pointer rule holds.

**One finding was refuted, and is recorded rather than silently dropped.** The reviewer read `[ "${#len}" -gt 3 ] && cmp=999` as a latent `set -e` carrier. The observation is true (it evaluates non-zero on the common path) but the inference is false: measured under `set -euo pipefail` with errextit confirmed live by a positive control, a bare unsuppressed call did **not** abort — bash exempts a command failing inside an `&&` list unless it follows the final `&&`, and `cmp=999` never runs. No entry point in this tier sets `-e` at all. The construct is also house style (15+ identical forms across `.claude/hooks/`, all clean under the pinned shellcheck gate).

A third item — gating the SPEC sentence that forbids `SKIP_HOOKS` in the shared helper — was **dropped**: the invariant holds at head, and the proposed "one-line grep arm" would red on correct code, since `.githooks/commit-msg` deliberately mentions `SKIP_HOOKS` in comments explaining its absence.

### Implementation notes worth review attention

- **Rung 1 invokes `python3` unconditionally** rather than gating on `command -v`. Presence is precisely the signal the contract says not to key on, and an absent interpreter yields empty output, which the same validity test already routes onward — one selection mechanism, not two.
- **`_codepoint_len("")` returns `0`, not a failure.** An empty subject stays an *out-of-range* rejection (`got 0`), not an *unmeasurable* one. The `16s` `empty` fixture's name refers to the ladder emitting nothing, not to empty input; the two paths now produce different messages.
- **The `.githooks/commit-msg` adapter** calls `check_commit_subject` unredirected and does not capture stderr, so its replacement text says "see the reason above" rather than re-stating a cause. Intentional under arm-scoped recovery, but it holds only while no caller swallows that stderr — none does today.

## Test plan

1. `_codepoint_len` never emits a non-decimal on rc 0.
2. It advances rungs on non-decimal output from a **successfully exiting** `python3` (banner-to-stdout case).
3. The fallback refuses on byte-counting-locale + non-ASCII, while still measuring ASCII exactly — so a degraded host keeps committing.
4. `check_commit_subject` refuses on `""` **and** on `9223372036854775808` / `99999999999999999999999999`. Needs `set -uo pipefail` script-file mode to reproduce the original accept.
5. The rejection message carries the **raw** length, not the saturated operand.

## Docs touched

- `SPEC.md` — §6.1.1 measurement contract, §6.1 fail-policy paragraph, §6.0 P4 arm-scoped recovery. The `helpers/conventional_commit.sh` fail-policy row is **byte-unchanged**; the new paragraph records the second axis.
- `changelog_unreleased/fixed/654.md` — category `fixed/` (SPEC §6.1 frames these as guardrail, not security boundary).

## Out of scope

- Rewriting `_codepoint_len`'s measurement strategy or removing the `python3` preference — the defect is the missing guard on its output.
- The 72-codepoint limit itself.
- A `python3` that **hangs** — unhandled by all three candidates, fails toward blocking, named rather than fixed.
- `extract_commit_subject`'s over-extraction on the legacy path — pre-existing, block-leaning, worth its own issue.
- `§148i-cli` reds on a full smoke run whenever an agent worktree is live (#642, parked). Unrelated to this change; measured in a quiet window.


ilgyu-yi:
## Finding triage (round 1)

```
reviewed-head: f99bed4226b556c1c106a9fa676ed6cba22f9a79
prior-round:   none (first round; 0 comments on this PR under the trusted-author filter)
dispatch:      findings by a worktree-isolated `code-reviewer` at f99bed42; judgment by a
```

One finding confirmed, one **refuted**, one dropped. The judge re-ran two of the reviewer's load-bearing measurements rather than accepting the set on trust; both held — the parent `f99bed42^` returns **rc=0** (subject reported VALID) while emitting `[: 9223372036854775808: integer expression expected` twice, and head returns rc=1; and head reports `got 5000`, not `got 999`, confirming the saturation touches only the compared operand.

### F1 — the new operator-facing message has no symptom→fix row · **confirmed** · `fix-now`

`docs/TROUBLESHOOTING.md#symptom-fix-index-table::doc-index-row`

The message is live and reachable (measured: the ladder-exhausted fixture returns rc=1 with the new text; the same fixture at the parent returned rc=0). The sibling message from the *same function* has a row at line 12; this one has none.

**Floor** — the row must name at least one actionable **environment** repair (a working `python3` on `PATH`, or a UTF-8 locale; ASCII-only subject as the stopgap). Measured reason: *no surface an operator actually sees today names one.* The helper emits cause + a SPEC pointer; the PreToolUse adapter appends the `ghjig_skip.sh` token form; the git-hook adapter appends `--no-verify`; `ghjig_commit` appends fix-the-subject. Every offered recovery is **escape the gate** or **edit the subject** — none repairs the environment that caused the refusal. So a row that merely restates the message is below the floor.

**Ceiling** — one row, ≤ ~350 chars (sibling-row scale; the measured band over the 21 content rows is 119–930, median ~603, and the two commit-format siblings sit at 119 and 132). It may **not** restate the ladder ordering, rung semantics, the `_codepoint_len` total-function contract, or the measurability-guard relation. It points at SPEC §6.1.1; it does not paraphrase it. That ceiling is what keeps the row inside SPEC §9's thin-pointer rule.

### F2 — "`[ "${#len}" -gt 3 ] && cmp=999` is a latent `set -e` carrier" · **REFUTED**

`.claude/hooks/helpers/conventional_commit.sh#check_commit_subject::errexit-carrier`

The reviewer's *observation* is true — the and-list does evaluate non-zero on the common path (`${#len}`=2 for a normal subject). Its *inference* is false.

Measured under `set -euo pipefail` (flags reported `ehuB`, so errexit was genuinely live): a bare, unsuppressed `check_commit_subject "feat(#1): a perfectly valid short subject"` **did not abort**. A bare mid-position failing and-list likewise did not abort. **Positive control**: a subsequent bare `false` *did* kill the script (rc=1), proving errexit was active and the survivals are not a dead measurement. Reproduced under both `/bin/bash` and `env bash`.

Bash exempts a command that fails inside an `&&` list from errexit unless it is the command following the final `&&` — here the failing command is `[ … ]`, and `cmp=999` (the exempt-from-exemption position) never runs. So there is no trap for `if … then … fi` to remove.

The finding's supporting fact is also weaker than needed: **no entry point in this tier sets `-e` at all** — `pre_tool_use.sh` and `.githooks/_lib.sh` are both `set -uo pipefail`, and `smoke.sh` has no `set -e`. Checked the four cited call sites plus three smoke call sites the reviewer did not enumerate.

*The one condition under which F2 would become real*, so this is not read as "the construct is unconditionally safe": if a future edit made that and-list the **last** statement of the function, it would return 1 on the common path — a false REJECT. It currently is not. The construct is also house style: 15+ identical `[ … ] && var=…` forms across `.claude/hooks/`, all clean under the pinned shellcheck gate. Changing this one instance would create a local inconsistency without touching the class.

Next round this reads back as `already-refuted (round 1)` and is not re-litigated without new evidence.

### F3 — the §6.0 "no `SKIP_HOOKS` in the shared helper" sentence has no gate · **confirmed** · `drop`

**The reviewer's own decline is upheld**, on three measured grounds. (1) Nothing to fix: the invariant *holds* at head (the helper carries 0 occurrences), so this is a hypothetical future regression, not a defect. (2) The proposed "one-line grep arm" is not one line if correct — `.githooks/commit-msg` deliberately *mentions* `SKIP_HOOKS`/`should_skip` twice in comments explaining their deliberate absence, so a naive `grep -c == 0` reds on correct code. *(This also corrects the reviewer's own phrasing: its "0 times in `.githooks/commit-msg` and `_lib.sh`" is imprecise — `_lib.sh` is 0, `commit-msg` is 2-in-comments. The substance holds: no live `should_skip` call, nothing in the operator-facing string.)* (3) Wrong venue — a fixed-string project-policy grep is what `lint_bash_idioms.sh` exists for; gating one arbitrary §6.0 sentence inside a `fix` PR, when no §6.0 principle is gated today, is scope expansion whose failure mode is low-cost and recoverable.

### Observations — not judged records, no action

- `docs/TROUBLESHOOTING.md` is **not** exhaustive: 21 content rows, with zero rows for `push-parity`, `merge-review`, or the destructive `out-of-scope` matcher. A missing row is therefore a discretionary gap, not a contract violation — **F1 survives on the narrower ground stated above**, not on a completeness claim.
- The reviewer's latin-1 observation is correct and was not re-measured.
- All four of the reviewer's line citations are accurate at head. Its factual accuracy is high; the single failure is the F2 inference.

ilgyu-yi:
## Finding triage (round 2)

```
reviewed-head: 1f56da52350441b39e9ff21818645c7ec7d705d2
prior-round:   1@f99bed4226b556c1c106a9fa676ed6cba22f9a79
dispatch:      TWO independent worktree-isolated `code-reviewer` invocations at this head (one
               fresh, one round 1's reviewer resumed); both emitted `ship`. Judgment by a
verdict token: untouched. Both `ship` verdicts stay `ship`. The judge casts no vote.
```

The judge was dispatched **despite two `ship` verdicts** because the two reviewers measured the same facts and reached opposite conclusions about the row the author had just written. All three findings' facts reproduce. The disagreement resolves **against Reviewer B on one premise, for Reviewer B on the conclusion** — and **F3 is substantially worse than reported**.

### F1 — the row's stopgap clause is unbounded · **confirmed** · `fix-now`

Reviewer A's facts are all correct. Reviewer B's *conclusion* is right; one of its *premises* is not.

B's ground (1) is measured-sound: **there is no operator dead end.** In every reachable configuration the Fix column's primary repair works. B's ground (2) — *"the looseness is what keeps the row inside the ceiling"* — is an **overstatement and is not upheld**: bounding the stopgap to the condition it was measured true under names a *locale* condition, which the Diagnosis already names, and states no ladder ordering, rung semantics, total-function contract, or measurability relation. Recording that premise as sound would license future inaccuracy under a constraint that does not bind.

But the row is closer to correct than Reviewer A allows: the stopgap is **true for both causes the row names**, and false only under a **third, unnamed** cause (ladder fully exhausted). That is an *incompleteness*, not a falsehood — and round 1 already established the table is not exhaustive, so a missing cause is discretionary.

**Reviewer A's remedy does not survive — measured.** The row is **351 codepoints**; round 1's ceiling was **~350**. Zero headroom. A's proposed addition (~44 chars) lands at ~395, a numeric breach — and it repairs the *cosmetic* limb (Diagnosis) while leaving the *actionable* limb (the stopgap instruction) untouched.

**Floor** — the Fix column must contain no sentence false on a host where the symptom is reachable; bound the stopgap to the measured condition or delete it. Round 1's floor is retained: the row must still name an actionable environment repair. **Ceiling** — a **replacement, not an addition**: ≤ ~350 codepoints, budget-neutral or net-negative, word-level, no new row, no contract restatement. **The Diagnosis disjunction is explicitly out of bounds** — cosmetic, measured non-dead-ending, and the limb whose repair would breach the ceiling. Not to be re-litigated.

### F2 — `ghjig_commit`'s printed recovery cannot succeed for the measurement cause · **confirmed** · `fix-now`

The P4 floor *is* met (cause text and the §6.1.1 pointer precede it), so this is not a floor violation — it is the one surface where the printed recovery is provably unachievable for one of its three causes. Bounded: on the byte-counting-locale variant "fix the subject" **does** succeed, so the line is wrong only on the exhausted host.

**Floor** — the recovery must not assert an action that cannot succeed for the measurement cause; a conditional qualifier suffices. **Ceiling** — half a clause, at this call site only; it must **point** at the environment repair, not restate it (round 1 routed that to the doc row; duplicating it is the second copy §9 forbids). Do not touch the shared cause text — its cause-only contract is what makes the three adapters' distinct recoveries possible.

### F3 — exit-status-keyed selection produces a live fail-open · **confirmed, mechanism CORRECTED, blast radius LARGER** · `defer-to-issue`, high priority

**Reviewer A's consequence is real; its mechanism attribution is wrong, and the correction makes this much more serious.** The fail-open is **not** in `extract_commit_subject` and does **not** merely collapse `-m` commits into the intentional `-F` fail-open. It is one step upstream in **`strip_command_data`** (`.claude/hooks/helpers/git_matcher.sh`), whose own header declares **FAIL-CLOSED** — "python3 absent, an unclosed quote, or any parse error prints the cmd UNCHANGED". That contract holds for a `python3` that exits **non-zero** (an exit-127 shim falls closed correctly) and is **violated** by one that exits **0 with wrong stdout**, because the primary is selected by `if out=$(… python3 …)` — exit status, precisely the mechanism this PR's own SPEC change forbids for `_codepoint_len`, in a sibling helper.

Because `strip_command_data` gates **arm entry**, the loss is not one check — the whole `git commit` umbrella arm (protected-branch, commit-format, directive-close) is skipped.

**Measured by the author, with controls, at this head:**

| command | healthy `python3` | banner `python3` (exits 0, wrong stdout) |
|---|---|---|
| `git commit -m "<valid subject>"` on `main` | **rc=2** — protected-branch blocked | **rc=0** — bypassed |
| `gh pr merge --squash` | **rc=2** — merge-review fails closed | **rc=0** — bypassed |

The triggering condition is not hypothetical: the banner-printing `python3` is the exact scenario `_codepoint_len`'s own header names, and smoke arm `16q` pins it. The project has already accepted it as real.

**Deferred on three measured grounds**, not on say-so: `git_matcher.sh` is **byte-identical at base and head** (`45f7072a…`), so it is untouched by this PR; #654's own *Out of scope* reads "Any other operand in the hook tree"; and "valid output" for a command-stripper has no simple predicate like the decimal test — it needs an output-validity protocol. Coverage is thin: `strip_command_data` appears **7 times in `smoke.d/`, all of them comments**, with no assertion exercising the selection mechanism.

**This is filed before #659 merges, not "someday."** It is the most serious thing measured this round and the one item at risk of being lost precisely because it is out of scope. Filing is not a merge gate — the judge emits no verdict token — but a round that spent two cycles on a doc row and let a silent protected-branch bypass go unrecorded would have inverted its own priorities.

- **F1 and F2 are one defect on two surfaces**: both are advice assuming the `wc` rung exists. Fix in the same commit or neither.
- Both reviewers' measurements were accurate wherever they overlapped the judge's. The single factual error across the round is Reviewer A's mechanism attribution in F3 — and it **understated** the defect.
- The banner shim's effect on `strip_command_data` was predicted by neither reviewer; the judge found it only because the case failed *open* where it was expected to fail closed.

ilgyu-yi (commented):
**Verdict: approve** — `code-reviewer` returned `ship` at the confirmed head, with no blocking findings.

The reviewer verified by execution rather than by reading, and ran the control in both directions. At the base commit it reproduced the fail-open in all four shapes — empty, `9223372036854775808`, `99999999999999999999999999`, and banner text each returned **rc=0 (ACCEPT)** with `[: … integer expression expected` on stderr; at head all four refuse. The digit pair is the load-bearing part: a digit-shape test alone would not have closed it, since the true domain of `[ -lt ]` is `intmax_t`-representable decimals rather than "digit strings".

`_codepoint_len`'s totality holds across every fixture (no third outcome observed), the measurability guard fires only where the answer would be wrong (non-ASCII under a byte-counting locale refuses; ASCII still measures exactly, so a degraded host keeps committing), and the saturation reports the **raw** length — `got 500` and `got 9223372036854775808`, never a fictional `999`.

The smoke arms are not vacuous: 6/6 green at head, and 1 pass / 5 fail with the base helper substituted, each red carrying its declared reason rather than a fixture-miss code. Full-suite delta is +6/0.

All three adapters' recoveries were printed at their actual surfaces, not merely read from source — the PreToolUse block naming `scripts/ghjig_skip.sh`, the git-hook tier naming `--no-verify`, and `ghjig_commit` carrying the round-2 qualifier that *points* at the cause text rather than restating the environment repair. Each was checked against a registry-removed control, so an rc=2 is the guard deciding rather than the guard never running.

Two review rounds preceded this one and both are durable on the PR. Round 1 refuted a `set -e` claim with a live positive control and the raiser withdrew it. Round 2 corrected a reviewer's mechanism attribution on a separate fail-open, which *enlarged* its blast radius — that defect lives in a file byte-identical at base and head, is outside this issue's stated scope, and is filed as **#660 (P1)**; the reviewer independently confirmed that framing is honest.

Three advisory idiom notes are recorded and none escalates (SPEC §4.5.1): a four-times-repeated decimal predicate that is defensible because each site takes a different action on the miss; a conditional qualifier that prints on all three causes, where the alternative would reintroduce cause-knowledge at the adapter and defeat the cause/recovery split; and a SPEC sentence about locale selection that reads accurately in context. All are preferences, not defects.

Issue #367, opened 2026-06-17
fix: commit-subject extraction mis-handles multi-m and embedded-newline single-m forms


ilgyu-yi:
## What
`extract_commit_subject` (`.claude/hooks/helpers/conventional_commit.sh`) mis-extracts the subject in three command shapes, all the same root — extraction operates on the wrong slice of the command rather than the commit's own message source:
1. **Single `-m` with embedded newlines** (subject then blank line then body in one quoted value): the greedy sed `.*-m[[:space:]]+"([^"]*)"` runs against the **normalized** command (`pre_tool_use.sh:125` collapses newlines to spaces first), so subject+body flatten into one quoted blob, the whole blob is captured as the "subject", and `_codepoint_len` trips the 1..72 gate (reproduced this session: a multi-line single-`-m` yielded a ~190-359-codepoint "subject" -> false length block).
2. **Canonical multiple `-m`** (a commit with `-m subject -m body`, git's documented multi-paragraph form): greedy `.*-m` matches the **last** `-m`, extracting the **body** as the subject -> false length block (long body) or false format block (non-Conventional-Commits body).
3. **Sibling heredoc in a compound command** (third facet, hit repeatedly this session): the heredoc branch (`conventional_commit.sh:19`, `grep -qE "<<-?..."`) keys on **any** heredoc opener anywhere in the raw command — not the commit's own message arg. So `cat > changelog.md <<EOF ... EOF && git add ... && <commit> -F msg.txt` makes the extractor return the *changelog's* first body line (`- SPEC's ...`) as the subject, rejected as non-Conventional-Commits, even though the `-F`-supplied message is well-formed.

SPEC §6.1.1 specifies subject extraction for only the plain single-line and the `$(cat <<TAG ...)` heredoc forms — silent on all three above.

## Why
MISSION "hooks-as-environment": the conventional-commit gate misfires on standard, documented commit invocations, forcing reword / message-file (`-F`) workarounds (hit this session — a single multi-line `-m` tripped a 359-char subject block; the sibling-heredoc form blocked `-F`-file commits chained after a `cat` heredoc). The 1..72 codepoint rule and the structural regex are correct; the gap is purely *which substring becomes the subject*, and SPEC under-specifies the extraction contract.

## Acceptance criteria
- [ ] A single `-m` with embedded newlines extracts only the **first line** as the subject — gated against 1..72, not the whole blob. Extract from the **raw** command (pre-normalization), mirroring the heredoc walk.
[...]

PR #661, opened 2026-08-15
fix(#660): key interpreter helpers on output validity, not exit status

### LINKED ISSUE #660: strip_command_data keys its python3 primary on exit status, so a banner-printing python3 silently disables the whole git-commit arm
## What

`strip_command_data` (`.claude/hooks/helpers/git_matcher.sh`) selects its `python3` primary by **exit status**:

```sh
command -v python3 >/dev/null 2>&1 || { printf '%s' "$cmd"; return 0; }
if out=$(printf '%s' "$cmd" | python3 -c '…
```

Its own header declares the opposite contract — **fail-closed**: *"python3 absent, an unclosed quote, or any parse error prints the cmd UNCHANGED"*. That contract **holds** for a `python3` that exits non-zero (measured: an exit-127 shim falls closed correctly, returning the command unchanged so the arm still runs). It is **violated** by a `python3` that exits **0 with wrong stdout** — the function returns that wrong stdout, the arm-entry `grep` misses, and **the arm is never entered**.

Because `strip_command_data` gates *arm entry* rather than one check, the loss is the whole `git commit` umbrella arm — protected-branch, commit-format, and directive-close together.

## Measured

Head tree materialised via `git archive`, exercised through the §3.2.1 `GHJIG_ROOT_OVERRIDE` seam with `CLAUDE_PROJECT_DIR` unset, a seeded `registry.txt`, and cwd inside the registered path. Script-file mode. **Controls run both ways**: guard present → blocks; registry removed → `rc=0` for everything, confirming an `rc=0` here is a real allow and not a no-measurement.

The shim is a `python3` that exits 0 after printing a banner to stdout:

```sh
#!/bin/sh
echo "Python 3.12.0 site banner: loaded venv"
exit 0
```

| command | healthy `python3` | banner `python3` |
|---|---|---|
| `git commit -m "fix(#654): a perfectly valid subject"` on branch `main` | **rc=2** — `commit on protected branch (main) blocked` | **rc=0** — bypassed |
| `gh pr merge <n> --squash` | **rc=2** — merge-review fails closed | **rc=0** — bypassed |

Isolated: `strip_command_data "$cmd" heredoc` returns the banner text instead of the command; `extract_commit_subject` is never reached. A `python3` that is **absent** (exit 127) correctly falls closed.

**The triggering condition is not hypothetical.** A `python3` that prints a site/venv banner ahead of its output is the exact scenario `_codepoint_len`'s own header names, and smoke arm `16q` pins it as a case the project has already accepted as real. PR #659 fixed precisely this selection mechanism in `_codepoint_len` — SPEC §6.1.1 now requires rung advancement to key on **output validity, not exit status** — while the sibling helper twelve lines away in a different file kept it.

## Why

A guard that fails **open** on a plausible environment defect is the wrong-allow direction MISSION reserves the hard negative face for, and this one is silent: nothing is logged, nothing is printed, and the operator sees a successful command. Two of the three checks it drops are the ones the shell exists to hold — a direct commit to a protected branch, and a merge that has not been reviewed at its head.

**Why now:** PR #659 (issue #654) established the contract that makes this a defect rather than a design choice. Before it, "select the primary by exit status" was simply how these helpers were written; SPEC §6.1.1 now names output-validity keying as the requirement, with the banner case as its stated motivation. The two helpers are now inconsistent, and the one still keying on exit status is the one guarding arm entry.

**MISSION fit:** *The mechanism* — enforcement's cost-asymmetry rule reserves the hard negative face for wrong-allows that are expensive or irreversible. A bypassed protected-branch gate writes to shared history; a bypassed merge-review gate lands unreviewed code on the default branch. Both are in that class.

**Existing coverage:** none. #654/#659 fixed `_codepoint_len` only and scoped this out explicitly (*"Any other operand in the hook tree"*). #647/#651 covered the two TTL clock operands. #652 is the `stop.sh` advisory counter; #653 is audit silence in `escape.sh`. None touches `git_matcher.sh`. `strip_command_data` appears **7 times across `smoke.d/`, all of them comments** — no assertion exercises its selection mechanism at all.

## Acceptance criteria

- [ ] `strip_command_data` advances past its `python3` primary on **output validity**, not exit status, so a `python3` exiting 0 with wrong stdout no longer silently disables the arm. Verified by mutation: with a banner-printing shim on `PATH`, `git commit -m "<valid subject>"` on a protected branch must **block** (`rc=2`), and the same run with a healthy `python3` must also block — both directions measured, since a fix that blocks everything is not a fix.
- [ ] The validity predicate is stated where a future reader will find it. "Valid output" for a command-stripper has no simple test like a decimal check — whatever protocol is chosen (a sentinel, a token-preservation invariant, a round-trip check) is recorded in SPEC alongside the §6.1.1 rule it parallels, so the two helpers stop diverging.
- [ ] A smoke arm exercises the selection mechanism, with a **paired negative arm**: a genuinely malformed command must still be caught after the fix. The arm must red against the current code. Today there is no such arm — the 7 existing mentions are comments.
- [ ] The **fail-closed** direction is verified for every failure mode the header claims: `python3` absent, non-zero exit, exit-0-with-wrong-stdout, and no `python3` on `PATH` at all. Each returns the command unchanged so the arm still runs.
- [ ] The `#367` / `#340` / `#403` behaviours that route through `strip_command_data` (heredoc-mode entry, the `-F`-only fail-open) are re-measured and unchanged. This is the highest-risk matcher-entry path in the tree and it currently has no assertion protecting it.
- [ ] The full smoke suite passes in a window with **no live agent worktree** (see #642 — a live worktree reds `§148i-cli` for unrelated reasons).

## Out of scope

- **Changing what `strip_command_data` parses.** This is about how a failed parse is detected, not about the parsing itself.
- **`extract_commit_subject`'s own exit-status keying.** Reviewer attribution initially placed this defect there; measurement showed the fail-open is upstream and `extract_commit_subject` is never reached. It has the same shape and may deserve the same fix, but it is a separate site with a separate blast radius and should not be folded in silently.
- **Auditing every `python3` invocation in the hook tree.** Two sites are now known (`_codepoint_len`, fixed in #659; `strip_command_data`, this issue). A sweep is a different unit of work.
- **The `-F`-only fail-open (arm `16j`, #367).** That one is deliberate and documented; this issue must not change it.

## Notes

Found by a `finding-judge` on PR #659 while adjudicating an unrelated documentation finding. The reviewer that surfaced the symptom attributed it to `extract_commit_subject`; the judge's isolation showed the mechanism is one step upstream, which **enlarges** the blast radius from "commit-format is skipped" to "the entire git-commit umbrella arm is skipped, including protected-branch". Filed before #659 merged, deliberately: it is out of that PR's scope and was therefore the item most at risk of being lost.

ilgyu-yi:
Closes #660

## Target base

`main`

## Why

`strip_command_data` selects its `python3` primary by **exit status** while its own header declares FAIL-CLOSED. A `python3` that exits **0 with junk stdout** — a site/venv banner is the canonical case, and smoke arm `16q` already pins it as real — makes the helper return the junk. The caller's arm-entry `grep` then misses, and the arm is never entered.

Because this gates **arm entry**, the loss is not one check. Measured through the §3.2.1 seam with a banner shim, controls run **both ways** (registry emptied → `rc=0` for every row, so an `rc=0` here is a real allow and not a no-measurement):

| command | healthy `python3` | banner `python3` |
|---|---|---|
| `git commit -m "<valid>"` on `main` | rc=2 protected-branch | **rc=0** |
| `git commit --no-verify` on `main` | rc=2 `--no-verify` blocked | **rc=0** |
| `git push --force origin main` | rc=2 force-push | **rc=0** |
| `gh pr merge <n> --squash` | rc=2 merge-review | **rc=0** |
| `python3` absent (exit 127) | rc=2 — falls closed correctly | — |

It is also **fully silent**: the audit log is byte-unchanged across bypassed runs. No block, no warn, no forensic trace.

## Plan

Three candidates; two mutually-blind challengers on `security` and `simplicity/maintainability`. Both returned **dominates-A**, and the `plan-reviewer` shipped A's protocol with nine mandatory corrections it re-measured itself.

**A's protocol won and was not beaten** — a framing sentinel. The shell mints a tag, the python frames every *success* write as `tag + payload + tag`, failure paths write nothing, and the shell accepts only a framed result. It is the only candidate total over arbitrary text, and it catches **truncation** as well as contamination. A rejected fd-3 for exactly that reason, and both the security challenger and the judge independently confirmed the rejection was sound.

**What the contest overturned:**

- **A's fix would not have fixed the reported bug.** The same defect exists at a second site in the same file — `space_glued_separators`, invoked as an **in-place write into `$cmd`** (74 references; the arm-entry variable). Measured: fixing only `strip_command_data` leaves **4 of 6 rows at `rc=0`**, including the `gh pr merge` row printed in this issue's own evidence table. That is also why A's open question "the merge path was not re-measured" was unanswerable.
- **A's justification for adding an audit record was measured false.** A argued the post-fix degradation would be silent. It is not — the strip returns *more*, never less, so every consequence is a loud block that `pre_tool_use.sh`'s `block()` already logs unconditionally. The warn would have been the first mutable global **and** the first audit emission in a file with zero of either. The challenger's `return 1` substitute was declined too: no caller reads the rc.
- **The random tag is decorative by A's own design** — the tag is handed to the interpreter as `argv[2]`, so a `python3`-controlling adversary reads it, and a non-controlling one cannot write stdout at all. Constant tag, measured identical across 40 cases, and it buys back determinism and a fixture that can assert the wire format.
- **Three parties miscounted the doc surface, including me.** A said four SPEC rows; a challenger said six; I measured two. The judge found a stale surface *none* of us counted, plus a comment labelling a safe pass-through "fail-open" — the opposite word. The Doc phase then found the `is_pr_merge_command` call sites are **five**, not four.

**Threat model, stated rather than implied.** A `python3` that reads its own `argv[2]` and echoes the tag around junk defeats the frame completely. This is an **accident** predicate, not an **adversary** predicate — and that scoping was *measured*, not asserted: an adversarial interpreter already defeats `check_destructive_args` regardless, so the tree's adversarial floor is set elsewhere and hardening this helper further buys nothing.

## Checklist

- [x] **Doc** (`50a2f99`) — new SPEC **§6.1.2** stating the rule once: a helper that delegates to an external interpreter **and** declares fail-closed keys acceptance on **output validity**, never on exit status or `command -v` presence. Carries the five-outcome enumerated set (absent / non-zero exit / exit-0-with-junk / exit-0-partial / stderr-only), what fail-closed *means* here, both instantiations (decimal for `_codepoint_len`, constant-tag frame for the two `git_matcher.sh` sites), and why a stripper needs a protocol rather than a shape test. Every other surface is a **pointer**, not a copy. §6.1.1 gains a one-clause back-reference naming itself the decimal instantiation.
- [x] **Test** (`0542b8c`, extended by `89d5890`) — `§159`: a multi-mode shim farm × three strip modes asserting fail-closed at **both** sites, a healthy-`python3` byte-identity differential proving no blinding, and end-to-end arms. All four RED propositions verified red against pre-fix code, each with a distinct rc so a fixture miss cannot masquerade as the intended failure.
- [x] **Code** (`3e07f59`) — frame both sites, one glob selector, presence gate dropped, constant tag, no audit emission.

Four findings, all confirmed, all fixed. Two were **my own drift** (F1/F2 — the stale boundary claim, above). The other two are worth naming because they are the kind a reviewer usually waves through:

- **F3a** — the arms claimed SPEC §6.1.2's five pathologies but never exercised **non-zero exit**, which #660's AC item 4 names verbatim. Now covered, and disclosed honestly in the case comment: measured **green on both sides of the fix**, because rc≠0 is exactly what an rc-keyed selector already handled. It is a regression + AC-coverage guard, not a defect witness — stated so it cannot be misread as new evidence.
- **F3b** — the shim's mode `case` had no `*)` default, so a typo'd mode degraded silently to the empty-output mode and the arm still **passed**. The judge established, by measurement rather than reasoning, that the naive remedies do not work: a bare `exit 1` leaves stdout empty ⇒ unframed ⇒ the helper fails closed ⇒ `[ out = in ]` passes and the closed-counter increments; a stderr complaint is swallowed by the subshells' `2>/dev/null`. The signal had to be **out-of-band**. Mutation-verified: a typo'd mode now reds 15/18 naming `rc=8`, where before it passed 15/15.

## Docs touched

- `SPEC.md` — new §6.1.2; force-push row, protected-push row, and the `is_pr_merge_command` contract repaired to point at it. **Deliberately left alone**: the `git reset --hard` and `git clean` rows (they *use* the helper but state no failure contract) and `ac_closeout_gate.sh`'s unrelated fail-closed at another function.
- `.claude/hooks/helpers/git_matcher.sh` — both headers rewritten. `space_glued_separators`'s old safety argument reasoned only about `python3` being *absent*, which never covered the exit-0-with-junk path that takes the **success** branch; that is now quoted and explicitly marked irrelevant. The else-branch comment now reads fail-**CLOSED**.
- `.claude/hooks/helpers/ac_closeout_gate.sh` — the false-contract restatement, with the pre-fix consequence named for all **five** arms that gate on the predicate.
- `docs/TROUBLESHOOTING.md` — one row, sibling scale, pointing at §6.1.2 rather than paraphrasing it.

## Residual, disclosed

The frame changes the **raw** (non-`$( )`) return on trailing-newline commands. Most call sites wrap in `$( )` and strip them again; the force-push arm pipes into `awk` instead, where the extra newlines survive harmlessly (`awk` renders blank lines, and `push_segments`' own `grep -E` drops them). **The canonical per-call-site statement lives in `git_matcher.sh`'s RESIDUAL header** — deliberately not restated here, because an earlier revision of this section carried a copy that went stale against its own PR within three commits.

That drift is itself in the record: `7199437` corrected the header, and round 1 of review caught two surviving copies — this section and a comment in the smoke file. Both are now pointers rather than tallies.

## Out of scope

- **`check_destructive_args`** — measured to stay `rc=0` *even with both sites fixed*, because banner words are absolutised against an in-registry `pwd`. Different mechanism, different fix shape; filed as a **P1 sibling** rather than folded in.
- **`extract_commit_subject`'s** own exit-status keying — same shape, different site; the fail-open here is strictly upstream, so it is never reached.
- **A sweep of every `python3` invocation.** `space_glued_separators` is included because it demonstrably carries *this* bypass, not because it shares a shape.
- **The deliberate `-F`-only fail-open** (arm `16j`, #367) — asserted unchanged, not touched.
- **Removing the `python3` dependency** from the matcher path. That is the only real answer to an adversarial interpreter, and it is a different unit of work.


ilgyu-yi:
## Finding triage (round 1)

```
reviewed-head: 71994370484fdb5aa1d7bce1117fd6479c932509
prior-round:   none (first round; 0 canonical markers under the trusted-author filter)
dispatch:      findings by a worktree-isolated `code-reviewer` at 7199437; judgment by a
```

All four confirmed, all `fix-now`. Two of them (F1/F2) are **my own drift**: I wrote the false claim, wrote the commit that corrected it in `git_matcher.sh`, and missed two copies.

### F1 — the stale claim survives verbatim in the smoke file · `scripts/test/smoke.d/50-perproject-recall.sh:3246`

`7199437` corrected `git_matcher.sh` and left `# BOUNDARY (all 8 of its call sites wrap it in $( ))` untouched one file over.

**Floor** — the sentence's *methodology* half must stay: it records that `mode` is measured through `$( )` while `sgs` is measured **raw**, which is what makes the golden hex readable. Verified true of this file. Drop it and a future editor re-measures at the wrong boundary and silently rewrites golden.
**Ceiling** — **no restated per-call-site tally.** The reviewer's proposed "one-line edit" (8 → 7+1) would recreate a second copy of `git_matcher.sh`'s RESIDUAL block — the exact duplication that drifted this round. Replace the parenthetical with a non-numeric pointer to that header. `remedy-survives: no`.

### F2 — the same stale claim in the PR body's `## Residual, disclosed`

**Floor** — the residual stays disclosed. Deleting the section converts a knowingly-disclosed residual into an undisclosed one, and the PR body is where a merger reads what is imperfect.
**Ceiling** — no numeric tally, no restated measurement count; state it qualitatively and point at the canonical header.

### F3a — the arms claim SPEC §6.1.2's five pathologies but never exercise non-zero exit

The shim's modes are `banner`/`partial`/`empty`/`stderr` (+`absent`), all `exit 0`. SPEC §6.1.2's set names **non-zero exit**, and #660's AC item 4 names it verbatim — so below this the AC is unmet and an unchecked AC rides to merge.

It is also this PR's own thesis turned inward: `7199437`'s body says *"the claim as written was stronger than the code, which is the exact defect class this PR's §6.1.2 exists to stop"* — and the arm's ok-text claims a set it does not cover.

**Floor** — all five SPEC outcomes present. **Ceiling** — do not widen past SPEC's five plus the existing `empty`; further interpreter misbehaviours (signal death, huge writes) multiply fixture surface with no contract anchoring them.

**Delete-vs-correct, considered and rejected here**: weakening the ok-text to "the modes exercised" would *hide* an AC gap rather than close it. Delete-beats-correct governs unmeasured **duplicated** claims (F1/F2); this is a contract-mandated **coverage** claim.

**Honest disclosure the fix commit must carry**: the judge measured the pre-fix blob (`2d6bf1e`) closing correctly on non-zero exit at **both** sites, while `banner` leaked at both. So the new arm is green on *both* sides of the fix — a regression/AC-coverage guard, **not** a defect witness. It must not be presented as new RED evidence.

### F3b — the shim's mode `case` has no `*)` default, so a typo silently degrades

`S660_PY_MODE=stdrer` → rc=0, stdout empty — byte-identical to `empty`.

**The naive remedies do not work, and this is measured, not reasoned.** I re-verified it independently: a `*) exit 1` default leaves stdout empty ⇒ unframed ⇒ the helper fails closed ⇒ the assertion `[ "$out" = "$in" ]` **passes** and the closed-counter increments, so the 15/15 and 5/5 count-guards do **not** catch it. And `*) … >&2` is swallowed by the subshells' `) 2>/dev/null`.

`action: fix-now` · axis `fail-open↔fail-closed`, target **high** — an extreme position, justified **independently** by §6.0 P1 cost-asymmetry rather than by any prior failure's direction: a wrong fail-open here is a green arm that measured nothing (the vacuity `smoke.sh`'s own header forbids), while a wrong fail-closed costs one loud re-run of a fixture the author controls.

**Floor** — an unknown mode must fail the arm **loud**, and the guard must be **out-of-band** (e.g. a marker file plus a new exit code alongside the existing 3/4/5/6/7). **Ceiling** — the guard stays inside the §159 fixture; it must not become a production assertion, gate anything outside these arms, or grow into a general mode registry.

- One imprecision in the reviewer's own wording: it says the **arm labels** enumerate the five pathologies. They say "all 5 pathologies"; the enumeration is imported by a comment one line over. F3a's substance is unaffected.
- F1 and F2 deliberately share an axis-key *kind* (`duplicated-claim`) across two sites, so a round 2 would match them as one drift pattern with two surfaces.
- AC6's clean-window suite run must happen with **no** agent worktree live — `§148i-cli` reds otherwise (#642, parked).

ilgyu-yi:
## Round-2 advisory nits — recorded, deliberately not fixed here

The `/file-review` reviewer returned **`ship`** with three optional findings, all comment-level in `.claude/hooks/helpers/ac_closeout_gate.sh` (a file this PR touches **comment-only**). It judged them non-blocking and said explicitly they can ride a later touch of that file. Recording rather than fixing, so they are not lost — and so the choice is visible rather than silent:

- **N1 (consistency)** — the block's opening still says the merge gates *"(ac-closeout / merge-strategy)"* must not engage, six lines above a line that now correctly enumerates **five**. Two counts in one comment block: the same drift shape round 1 ruled on as F1/F2. Per that round's ceiling the remedy is to make the opening **non-numeric**, not to add a second five-item list.
- **N2 (precision)** — the five-arm list includes `merge-completeness`, which never blocks (fail-open advisory). Reads as five bypassed gates where four block and one is advisory.
- **N3 (cosmetic)** — one 100-column line in an otherwise ~78-column wrapped block; a rewrap artifact.

**Why not fixed in this PR.** Each further round tonight has surfaced new findings, which is the correction loop the round-1 judge warned about — and committing here would invalidate the head-pinned review and require a fresh one for three comment nits. The independent reviewer's own call was non-blocking. They carry to the next touch of `ac_closeout_gate.sh`, most likely the `check_destructive_args` sibling issue, which lives in the adjacent code.

## A live recurrence of #646, worth the record

The reviewer's first reading of its own suite result came from a **stale `smoke.out` left at the scratchpad root by a different agent**, not from its own run. It caught the discrepancy, discarded the reading, and re-read from its own minted directory.

That is precisely the contamination #646 exists to prevent, firing in the wild during the review of an unrelated PR — and the reason it was caught is the fresh-directory discipline #646 added to the reviewer contracts.

ilgyu-yi (commented):
**Verdict: approve** — `code-reviewer` returned `ship` at the confirmed head, with no blocking findings.

The reviewer verified by execution rather than by reading, and ran controls in both directions throughout. Splicing the pre-fix helper into the head tree reproduces the defect exactly — `159a` 6/18, `159c` 2/6, `159f` **0/5 with all five gates at rc=0**, `159h` rc=0 — while head is 9/9 green. Fixing only the first of the two sites leaves `159f` at **2/5**, with force-push, protected-push and `gh pr merge` still bypassed, which confirms both sites are load-bearing rather than one being a shape-sharing sweep.

Three round-1 claims were checked rather than accepted, and all three held:

- The new `crash` arm's disclosure that it is **green on both sides of the fix** is accurate, not overstated: the surviving pre-fix legs are exactly `crash`×3 modes plus `absent`×3 modes. It is a regression and AC-coverage guard, and the case comment says so.
- The mode-typo guard is justified rather than gratuitous fixture machinery. Substituting a typo'd mode with the shipped out-of-band marker reds `159a` 15/18 and `159c` 5/6 naming `rc=8`; with a bare `*) exit 1` instead, the arms go **18/18 and 6/6 — fully green while measuring nothing**.
- No surviving copy of the stale "all 8 call sites wrap" claim exists anywhere in the tree; both former copies are now pointers to the canonical header rather than restated tallies.

No blinding: breaking the accept glob to produce a true always-return-unchanged helper reds `159b`, `159d` (24/44) and `159g`, so the differential is not vacuous. An end-to-end rc differential pre-fix versus head under a **healthy** `python3`, over 15 command shapes including trailing-newline force-push commands that exercise the raw non-`$( )` call site, showed **zero** differences on a corpus that discriminates. Dropping the presence gate leaks no stderr — with `python3` absent both helpers return the input byte-identical and write 0 bytes to stderr.

Every measurable claim in the diff was audited and is true: 8 call sites with 7 wrapping and 1 piping, one `sgs` call site, 74 `$cmd` references, five `is_pr_merge_command` guard sites. Lint clean, TOC consistent, changelog fragment present, and the TROUBLESHOOTING row is a pointer rather than a restated contract.

Three optional comment-level nits in `ac_closeout_gate.sh` were reported and judged non-blocking by the reviewer, which recorded that they can ride a later touch of that file. They are documented in a PR comment rather than fixed here, with the reasoning stated.

PR #663, opened 2026-08-16
fix(#662): refuse what the tokenizer cannot parse, at three sites

### LINKED ISSUE #662: check_destructive_args keys its shlex tokenizer on exit status, so a banner-printing python3 silently allows rm -rf outside the registry
## What

`check_destructive_args` (`.claude/hooks/pre_tool_use.sh`) tokenizes the command with `python3 -c 'import shlex …'`, selecting the interpreter by **presence** (`command -v python3`) and then by **exit status** (`if ! tok_out=$(… python3 …)`). A `python3` that exits **0 with junk on stdout** — a site/venv banner is the canonical case — takes the *success* branch, so the junk becomes the token list.

The junk tokens are then handed to `path_in_scope`, which **absolutises a relative path against `pwd -P`**:

```sh
case "$p" in
  /*) ;;
  *) p="$(pwd -P)/$p" ;;
esac
```

Banner words (`Python`, `3.12.0`, `banner`) are relative, so they resolve *inside* the registered project — every one is "in scope", nothing is out of scope, and the destructive-argument guard allows the command.

**This is not the defect #660 fixed.** #660 repaired output-validity keying in `git_matcher.sh`'s two helpers. This is a third site, in a different file, with a **different downstream mechanism**: there the junk caused an arm-entry `grep` to *miss*; here the junk is consumed as a token list and then laundered into in-scope paths by `pwd`-relative absolutisation. Fixing `git_matcher.sh` does not touch it.

## Measured

At `5ae1cf5` (i.e. **with #660 already merged**), through the §3.2.1 `GHJIG_ROOT_OVERRIDE` seam, `CLAUDE_PROJECT_DIR` unset, seeded registry, cwd inside the registered path. Script-file mode. The shim is a `python3` that exits 0 after printing a banner:

```sh
#!/bin/sh
cat >/dev/null 2>&1
echo "Python 3.12.0 banner"
exit 0
```

| `rm -rf <out-of-registry path>` | rc |
|---|---|
| healthy `python3` | **2** — `destructive command points outside registry` |
| banner `python3` | **0** — allowed |
| control: registry removed, healthy `python3` | **0** |

The control is what makes the middle row meaningful: `in_scope … || exit 0` returns 0 for every input without a registry, so an rc=0 could otherwise mean *the guard never ran*. With the registry present and a healthy interpreter the same command blocks, so the banner rc=0 is a **real allow**.

## Why

`rm -rf` on a path outside the registry is the irreversible wrong-allow the out-of-scope destructive matcher exists to prevent — data loss with no undo, and the failure is **silent** (the hook exits 0; nothing is written to the audit log because `block()` is never reached).

**Why now:** #660 established the contract that makes this a defect rather than house style. SPEC §6.1.2 now requires a helper that delegates to an external interpreter **and declares a fail-closed posture** to key acceptance on **output validity**, never on exit status or `command -v` presence. `check_destructive_args` does both of the forbidden things. Sharper still: **two merged artifacts already assert a tracking item that does not exist.** SPEC §6.1.2's own scope paragraph names `check_destructive_args`, states this exact `pwd`-absolutisation mechanism, and says it is *"tracked separately"*; PR #661's Out-of-scope says it is *"filed as a P1 sibling"*. Neither was true until this issue. Filing it is what makes those two statements honest.

It was deliberately scoped **out** of #660 — the plan-reviewer, both plan challengers, and the activation reviewer each measured it independently and agreed folding it in would be the tree-wide `python3` sweep that issue's invariant 5 forbids — so it was left for its own unit of work. This is that unit.

**MISSION fit:** *The mechanism* — the cost-asymmetry rule reserves the hard negative face for wrong-allows that are irreversible. Deleting files outside the registered project is squarely that class, and this one fails open silently.

**Existing coverage:** none, and I checked the adjacencies. #660 (merged, `5ae1cf5`) fixed `strip_command_data` and `space_glued_separators` in `git_matcher.sh` and explicitly named this site as out of scope. #654 (merged) fixed `_codepoint_len` in `conventional_commit.sh`. #653 is audit silence in `escape.sh`'s reject arms — a different file and about records, not selection. No open issue mentions `check_destructive_args` or `path_in_scope`.

## Acceptance criteria

- [ ] `check_destructive_args` keys its tokenizer on **output validity**, not on `command -v` presence or exit status, per SPEC §6.1.2. Verified by mutation in **both** directions: with a banner-printing shim on `PATH`, `rm -rf <out-of-registry>` must **block** (`rc=2`); with a healthy `python3` the same command must **also** block, and an **in-registry** `rm -rf` must still be **allowed** (`rc=0`). A fix that blocks everything is not a fix. *(Disclosed, per the precedent #661's F3a set: the healthy-block and in-registry-allow clauses are **pre-satisfied regression guards** — they pass today. Only the banner-blocks clause reds against current code.)*
- [ ] The validity predicate is chosen and recorded. `shlex.split` returns a *token list*, not a single string, so the framing sentinel #660 used at the `git_matcher.sh` sites may or may not transfer; whichever protocol is used is stated in SPEC alongside §6.1.2 so the three fixed sites share one rule rather than three.
- [ ] The **fail-closed direction** is verified for every outcome SPEC §6.1.2 enumerates — absent, non-zero exit, exit-0-with-junk, exit-0-partial, stderr-only. "Fail closed" here means the guard still evaluates the real arguments, or refuses; it must not silently allow.
- [ ] **A cwd-import substitution cannot forge the decision.** `python3 -c` puts the cwd on `sys.path`, so a `./shlex.py` planted **inside the registry** — a write the shell's own Edit/Write guards permit — makes a **perfectly healthy** interpreter return arbitrary tokens, and those tokens are substituted into the command every downstream Bash matcher sees. Measured: with the plant, `rm -rf <outside>`, `git push --force origin main`, `git reset --hard` and `git clean -fd` are **all rc=0**; without it, all rc=2. Verified by mutation that **both** `-I` sites are load-bearing — restoring `-I` at only one of them leaves the plant winning.
- [ ] A smoke arm exercises the selection mechanism with a **paired negative**, and **reds against current code**. It must fail loudly on a fixture miss rather than pass — an unknown shim mode must not degrade into a green arm (measured on #661: a bare `exit 1` default left the equivalent arm 18/18 green while measuring nothing, because empty output is itself the fail-closed answer).
- [ ] **Under a healthy interpreter**, the pre-existing destructive-matcher behaviours are re-measured unchanged: an in-registry destructive command is still allowed, and the `$HOME`/`~` substitution arms in `path_in_scope` still resolve as before. *(Qualified after measurement: the fix's fallback rung refuses any command containing a quote or backslash, so under a **degraded** interpreter some in-registry commands block — that is the next AC, not a violation of this one.)*
- [ ] **The degraded-interpreter over-block is bounded, disclosed, and keyed as measured.** The refusal keys on the **whole command string**, not the destructive segment — so `mv ./a ./b && echo "done"`, fully in-registry with a quote in an *unrelated* segment, blocks when the interpreter is degraded. The comparison to shipped behaviour is stated **per outcome**, not in aggregate: it is *narrower* than today only for the `exit-1` outcome (which already blocks all in-registry destructive commands); for the `absent` outcome it is a **widening** (a quoted in-registry command goes 0 → 2). The block message names the `scripts/ghjig_skip.sh` escape.
- [ ] The full smoke suite passes in a window with **no live agent worktree** (see #642 — a live worktree reds `§148i-cli` for unrelated reasons).

## Out of scope

- **`path_in_scope`'s `pwd`-relative absolutisation itself.** It is correct for real relative paths and is what makes `rm -rf ./build` work; the defect is that junk reaches it, not that it absolutises. Changing it would be a much larger blast radius.
- **A degraded or absent `jq`.** Measured: `pre_tool_use.sh`'s `command -v jq … || exit 0` blinds the **entire** hook — the sensitive-file control returns rc=0 with `jq` absent. Pre-existing and deliberate, and **the reason the SPEC "add-only" sentence must be scoped to the `python3` delegation sites**: stated unscoped it would be a third measurably-false claim in the same contract.
- **A sweep of every remaining `python3` invocation in the tree.** Three sites are now known and individually justified (`_codepoint_len` #654, the two `git_matcher.sh` helpers #660, this one). A sweep is a separate unit of work with its own rationale.
- **`extract_commit_subject`'s exit-status keying** — same shape, still unfixed, but its fail-open direction and blast radius differ and #660 already named it as deliberately deferred.
- **The adversarial-interpreter threat model.** A `python3` that reads its own argv and forges whatever protocol is chosen defeats any output-side predicate; SPEC §6.1.2 already records this as an accident-not-adversary model, measured against the tree's existing floor. This issue does not reopen it.

## Notes

Surfaced during PR #661's plan contest by the `security` challenger, then independently confirmed by the plan-reviewer and by the activation reviewer on #660. All three measured it as staying `rc=0` **even with both `git_matcher.sh` sites fixed**, which is what established it as a separate site rather than a shape-sharing duplicate.

Re-measured by the author at `5ae1cf5` after #660 merged, with the control run both ways, before filing — the numbers above are that run, not the inherited ones.

Three unrelated comment-level nits in `.claude/hooks/helpers/ac_closeout_gate.sh` were recorded on PR #661 as non-blocking and deferred to "the next touch of that file". That file sits adjacent to this work; whoever takes this issue should pick them up if the diff lands nearby, but they are not part of this issue's scope.

ilgyu-yi:
Closes #662

## Target base

`main`

## Why

Three defects, one root cause, three sites — all measured with a **same-PATH positive control** (`Edit` on `.env`, which has no `python3` leg) proving the hook was alive for every row.

| | healthy | absent | empty-py | exit-1 | plant |
|---|---|---|---|---|---|
| `rm -rf /etc/httpd` | 2 | 2 | — | — | **0** |
| `rm -rf "/etc/httpd"` | 2 | **0** | — | — | — |
| `git push --force origin main` | 2 | — | **0** | — | **0** |
| `git reset --hard` / `clean -fd` | 2 | — | **0** | — | — |
| `rm -rf ./build` (in-registry) | 0 | 0 | — | **2** | — |
| `mv ./a ./b && echo "done"` (in-registry) | 0 | **0** | — | **2** | — |
| **control** `Edit` `.env` | 2 | 2 | 2 | — | 2 |

1. **`check_destructive_args`** keys on `command -v python3` then exit status; junk becomes the token list and `path_in_scope` absolutises it against `pwd -P` into scope. Its fallback word-split rung **also leaks quoted absolute paths**.
2. **`parse_env_prefix`** (`escape.sh`) — `jq` exits **0** with empty output on empty input, so its `||` fallback never fires and the blank is written back. **`cmd` is blanked and every Bash matcher dies upstream.**
3. **`python3 -c` puts cwd on `sys.path`.** A `./shlex.py` planted **inside the registry** — a write the shell's own guards permit — makes a **perfectly healthy** interpreter return arbitrary tokens, substituted into the command every downstream matcher sees. `python3 -I` closes it, and **both** `-I` sites are load-bearing (restoring only one leaves the plant winning).

## Plan

A three-candidate contest with two mutually-blind challengers on `security` and `correctness`. Both returned **dominates-A**, and the `plan-reviewer` **disqualified the base plan at the invariant gate**.

**Plan A's central premise was measurably false.** It argued the fallback rung already gives a healthy interpreter's verdicts, so the fix was "make exit-0-junk behave the way absent already behaves". Measured: `rm -rf "/etc/httpd"` is **rc=0** on that rung and rc=2 healthy. Quoted, single-quoted, `~` and `$HOME` forms all leak — 10 of 12 out-of-registry forms. Invariant 1 failed at 4 of 5 outcomes.

**The winning design is "refuse what you cannot tokenize"** — one mechanism rather than an enumeration. Frame the token list and admit iff framed **and non-empty**; the fallback rung handles only a quoting-free command and **refuses** otherwise, which also reproduces today's unclosed-quote block, so no verb vocabulary is needed.

**Two findings no candidate named**, both from the judge and both verified here: the upstream blanking (defect 2), and that once it is repaired, an empty array under `set -u` on bash 3.2.57 is a fatal `args[@]: unbound variable` → hook exit 1 → **non-blocking allow, every later matcher skipped**.

**One claim struck, in the base plan's favour.** A challenger asserted the plant also defeats #660 (merged). It does **not** — measured: planting `./re.py` and running `git commit` on `main` gives rc=2 with **and** without. Breaking the import yields *unframed* output → fail closed → the arm still blocks. #660's frame authenticates that the program **ran to completion**; this site differs because the tokens **are** the decision data.

## Checklist

- [x] **Doc** (`381a74d`) — SPEC §6.1.2 third instantiation, stating the rule the other two do not cover: *where the caller iterates the output as a set or substitutes it back into the command, a lossy fallback must refuse what it cannot tokenize.* Records the **inverted fail-closed direction** (an empty payload frames validly and reads as a completed check that checked nothing), the `-I` requirement at the two load-bearing sites with the data-inside-the-guarded-surface distinction, and the `parse_env_prefix` site as an **AC3 precondition**. Two measurably-false shipped claims repaired.
- [x] **Test** (`2693cf7`) — smoke §160: nine arms. All six RED propositions verified red against pre-fix code, each with a distinct rc so a fixture miss cannot masquerade as the intended failure. Sabotage-tested against a throwaway **fixed** root — the only place a vacuous arm can hide when the fail-closed answer *is* "block" — catching a typo'd mode, a broken call-site discriminator, a leaked real interpreter, and a blinded hook.
- [x] **Code** (`19dd40b`) — the three edits, `-I` at the two sites, and the `${args[@]+…}` guard.
- [x] **Doc correction** (`039cba9`, `2feca19`) — the Phase A troubleshooting rows named `SKIP_HOOKS=<category>`, which SPEC §7 says the live Bash tool **strips before the hook reads the command**: a dead escape, while the code names the live one. Corrected to the file-token form — then corrected again, because the corrective commit left a stray backtick in both rows (measured 29/27 odd at head vs 26/24 even at its parent). The second fix is bounded delete-only: two characters, zero prose changed.

Two `fix-now` (both above), one `defer-to-issue`, four `drop`. The judge **spot-checked the reviewer's negative results rather than accepting them**, and reproduced the `-I` mutation table exactly — both sites load-bearing, and not vacuous in either direction (the positive control held at rc=2 on every variant, and in-registry `rm -rf ./build` held at rc=0, so the head is not simply blocking everything).

**Deferred**: every §160 arm asserts `rc=2`, so nothing pins that the fallback rung still *allows* a quote-free in-registry command under a degraded interpreter. Measured correct at head, so it is a future-regression guard rather than a defect — and it **cannot satisfy any AC**, since #662's AC5 demands an arm that reds against current code and a `want-rc=0` row is green on both sides by construction.

**Dropped with reasons, not waved off**: the duplicated recovery sentence is the tree's *established* §6.0 P4 pattern (each arm names the recovery live on its own surface); the `cda_refusal` global is documented three lines above its declaration with an explicit "deliberately not `local` — the caller reads it".

## Disclosed honestly

**The add-only rule is scoped, deliberately.** Stated unscoped it would be a **third measurably-false claim** in the same contract: `pre_tool_use.sh`'s `command -v jq … || exit 0` blinds the **entire** hook — the sensitive-file control returns rc=0 with `jq` absent. That is pre-existing and deliberate, and it is recorded as a named exception rather than papered over.

**The over-block is stated per outcome, not in aggregate** — the aggregate claim is false. The refusal keys on the **whole command string**, not the destructive segment, so `mv ./a ./b && echo "done"` — fully in-registry, quote in an *unrelated* segment — blocks under a degraded interpreter. It is **narrower** than shipped behaviour only for `exit-1`; for `absent` it is a **widening**.

## Out of scope

- **The 10 remaining `python3 -c` sites without `-I`** (measured: 12 total, 0 with `-I` today) and the emulating-module residual → a sibling sweep issue, filed rather than folded in.
- `path_in_scope`'s pwd-relative absolutisation; `extract_commit_subject`; the PATH-hostile-interpreter model, which §6.1.2 defers as accident-not-adversary — the plant is *data inside the guarded surface*, which is why it is in scope here and that model is not.


ilgyu-yi:
## Finding triage (round 1)

```
reviewed-head: 039cba96fe4b5d5e09a9ec219e9401f0d55e2053
prior-round:   none (first round; 0 canonical markers under the trusted-author filter)
dispatch:      findings by a worktree-isolated `code-reviewer` at 039cba9; judgment by a
```

Two `fix-now`, one `defer-to-issue`, four `drop`.

### F1 — both rows rewritten by `039cba9` end with an unbalanced backtick · **confirmed** · `fix-now`

`docs/TROUBLESHOOTING.md`, the two symptom rows.

Measured parity **29 / 27 — odd on both**; neighbouring rows are 13 / 16, even. And the sharper point the reviewer understated: **the parent of `039cba9` was 26 / 24, even.** This is not pre-existing rot — the corrective commit introduced it, in the exact two rows it exists to make correct.

**Floor** — at least one backtick deleted in each row. **Ceiling** — **at most one character per row, two characters total, zero prose tokens changed.** That bound is measured, not preferred: rendering the 1-char-deleted rows through GitHub's own GFM renderer shows the stray gone with every code span still well-formed, including `scripts/ghjig_skip.sh <category> '<cmd>' '<why>'`.

**Delete, do not reword.** The prose is correct; the defect is one character. Rewording would rewrite more than necessary in the two rows a previous rewrite already damaged — the "deleting beats correcting" precedent applies, and this ceiling exists to stop a *third* rewrite of these rows from manufacturing a round-2 finding.

### F2 — PR body checklist out of sync · **confirmed** · `fix-now`

Test and Code still show unchecked though both commits landed, and the fourth commit has no line at all. Axis: **none — discrete**; a box is checked or it is not. Confined to the PR body; cannot touch the artifact. Refetch the remote body immediately before editing and abort on external drift.

### F3 — the disclosed degraded-interpreter **lower** bound is not test-pinned · **confirmed as a gap, NOT a defect** · `defer-to-issue`

Every §160 RED arm asserts `rc=2`; nothing pins that the fallback rung still **allows** a quote-free in-registry destructive command under a degraded interpreter. The asymmetry is real — every arm pins the ceiling, nothing pins the floor — and a future over-eager widening of the refusal predicate would leave all nine arms green.

Deferred on three grounds: (a) **not a defect** — measured correct at head (`rm -rf ./build` → rc=0 under a banner interpreter, with the same-PATH positive control at rc=2 proving the arm measured something); (b) it **cannot satisfy any AC** — #662's AC5 demands an arm that *reds against current code*, and a `want-rc=0` row is green on both sides of the fix by construction; (c) the AC set is already satisfied by nine arms, and a tenth expands a four-commit PR past its own scope.

### F4 + the three idiom notes — **drop**

The reviewer's negative results were **spot-checked rather than accepted**, and the `-I` mutation table reproduced exactly:

| `-I` at | control (want 2) | `rm -rf <outside>` (want 2) |
|---|---|---|
| none | 2 | **0** |
| cda-only | 2 | **0** |
| pep-only | 2 | **0** |
| **both (head)** | 2 | **2** |

Both sites load-bearing, and **not vacuous in either direction**: the positive control held at rc=2 on every variant (so an rc=0 is a real allow), and in-registry `rm -rf ./build` held at rc=0 on every variant (so the head is not simply blocking everything).

The idiom notes are dropped with reasons rather than waved off. The duplicated recovery sentence is the tree's **established** §6.0 P4 pattern — each arm names the recovery live on *its own* surface, and factoring it into a shared constant would defeat the rationale that produced it. The `cda_refusal` global is documented three lines above its declaration with an explicit *"Deliberately not `local` — the caller reads it"* and a §6.0 P4 justification: a flagged departure the file itself pre-justifies is the false-positive class this role exists to absorb.

### Observation — not a judged record

The judge notes it reshaped the same probe command several times against the worktree-isolation guard before switching to small single-purpose scripts, and names that reshape-until-it-passes loop as a pattern to avoid: the right move is to **decompose rather than re-phrase**.

ilgyu-yi:
## Round-2 advisory notes — one fixed, one recorded

The `/file-review` reviewer returned **`ship`** with two non-blocking notes.

**N2 — fixed** (`a793a40`). The `160b` case comment claimed *"15 of 18 rows rc=0"* while its own next sentence enumerates **four** blockers (`crash/{bare,quoted,squoted}` + `absent/bare`) — 18 − 4 = **14**, and the pre-fix run reports 4/18 blocked. `160d`'s equivalent line was already correct. Fixed rather than deferred because a false measured figure is precisely the defect class this PR exists to close, and correcting one number is net-negative claim surface.

**N1 — recorded, deliberately not fixed here.** `python3 -I` implies `-E`, `-P` and `-s`. `-P` is the intended effect (drop the cwd from `sys.path`); `-E` (ignore all `PYTHON*` env, including `PYTHONHOME`/`PYTHONPATH`) and `-s` (no user site) ride along. For a **relocated or embedded** `python3` that needs `PYTHONHOME`, the hook now yields unframed output → refusal — while the recovery text tells the operator to check `python3 -c 'print(1)'`, which would print `1` and clear them, **diagnosing nothing**.

The direction is fail-closed and escapable, so it is a diagnosability gap rather than a defect. The cheap repair is one clause naming `python3 -I -c 'import shlex'` as the diagnostic.

**Why not fixed in this PR.** Fixing it means **adding** prose to three surfaces (two code comments and a TROUBLESHOOTING row), and the rule this session arrived at — after a round where a corrective commit introduced a fresh defect into the two rows it existed to correct — is that a round whose fixes only *remove* checkable claims cannot generate the next round's finding, while one that adds prose can. N2's fix subtracts; N1's would add. It carries to the sibling sweep issue, which touches the same `-I` surface at the ten remaining sites and is the natural place to state the flag's full effect once.

ilgyu-yi (commented):
**Verdict: approve** — two independent `code-reviewer` invocations reviewed this head. One returned `ship`; the other returned `ship after fix` whose sole fix was **a GitHub artifact action with zero file edits**, and that action is now done, so the tree it cleared is unchanged.

Both verified by execution rather than by reading. The six RED arms are genuinely red without the fix (`160b` 4/18, `160d` 1/6, `160e` rc=0, `160f` 0/6, `160g` 0/4, `160h` a fatal `args[@]` unbound), while the anti-vacuity spine holds green on both sides. `160c` was proved non-vacuous with a `check_destructive_args() { return 1; }` mutant, which reds it at 4/9 while leaving every other arm green — "block everything" does not pass. The `-I` mutation table reproduced exactly, confirming both sites load-bearing, and the same-PATH positive control held at rc=2 on every variant so each rc=0 is a real allow rather than a dead fixture.

The disclosed lower bound was checked directly: a quote-free in-registry destructive command is still allowed under all six degraded outcomes, so the head is not blocking everything, and the per-outcome framing (narrower only for `exit-1`, a widening for `absent`) is accurate where the aggregate claim would have been false. The SPEC scoping of the add-only sentence was confirmed load-bearing — `command -v jq … || exit 0` sits ahead of every matcher, so stating it unscoped would itself have been measurably false.

The final commit was verified as exactly one comment line, one character class, with line counts identical on both sides. The corrected figure was re-derived from the loop rather than trusted: 6 shim modes × 3 operand forms = 18 rows, 4 blockers, 14 rc=0. The old value was wrong.

**One finding held the merge and has been resolved without touching the tree.** Three artifacts on this branch asserted a sibling sweep issue that did not exist — the identical shape #662 was filed to repair, one SPEC section lower. It was also load-bearing for a deferral: a deferral whose destination does not exist is not a deferral. That issue is now filed as **#664**, before this PR merges rather than after, which is the difference between this instance and the two prior ones. Both SPEC sentences are numberless, so they became true on filing with no prose added anywhere.

Two advisory notes are recorded on the PR and deliberately not fixed here. The `-I` flag implies `-E` and `-s` alongside the intended `-P`, so a relocated interpreter needing `PYTHONHOME` is refused while the recovery diagnostic would clear it — a diagnosability gap on the fail-closed side, carried to #664 as an AC because that is the one place the flag's full effect should be stated once. A test-message arithmetic understates its own coverage; it misrepresents nothing about safety and folds into the same issue.

Issue #658, opened 2026-08-15
nothing locks agreement or per-role distinctness between the two duplicated reviewer scratch contracts


ilgyu-yi:
## What

`code-reviewer.md` and `security-reviewer.md` both carry a `## Scratch discipline (#646)` section. Measured at `fa39d47`, each block is **16 lines** and they differ at **4** of them: line 5 (the `mktemp -d` template — a role-name substitution) and lines 2, 9, 11 (role-specific rewrites). The other 12 lines are byte-identical. `§158d` (`scripts/test/smoke.d/70-gates-contentlocks.sh`) locks that both sections **exist** — one heading, one extractable template each — and `§158a–c` execute each contract's template. Nothing locks that the two **agree** on the 12 shared lines, and nothing locks that their templates are **distinct**.

Measured consequences, all from PR #657's own review rounds:

1. **A defect in the shared text is duplicated by construction.** Round 3's F-A found the seam recipe did not execute as written; the defective bullet was **byte-identical in both files**, so one authoring mistake shipped twice and needed two edits.
2. **Hand-sync is already happening.** Round 2's F-2 corrected a `safe_source` count in both files by hand; round 3's F-A/F-B did it again at `fa39d47`. Two hand-syncs in three rounds.
3. **Per-role template distinctness is unenforced.** Setting `security-reviewer.md`'s template equal to `code-reviewer.md`'s leaves every arm green — measured, `pass=7 fail=0`.

**What consequence 3 does and does not cost.** It does **not** cause collisions: `mktemp -d`'s `O_EXCL` retry makes two concurrent invocations of the *same* template still receive distinct paths — measured, 10 concurrent runs of one identical template in one root → 10 distinct paths, 0 duplicates. The cost is **attribution**, which is the stated purpose of the naming convention: SPEC §1.5 grounds it on an observed scratchpad root where four separate actors independently chose the name `probe`, leaving no way to tell whose directory was whose. Two roles minting `ghjig-code-reviewer-*` reproduces exactly that, one level up. This is a weaker claim than collision, and it is the one the acceptance criterion below is sized to.

## Why

[...]

PR #651, opened 2026-08-14
fix(#647): guard the clock operand at both TTL sites

ilgyu-yi:
Closes #647

## Why

`scripts/ghjig_file_review_post.sh` guards the staging file's mtime against a malformed epoch, then five lines later reads `now=$(date +%s)` **unvalidated** into the same `$(( now - mt1 ))`. One operand of the TTL was guarded and its twin was not.

That guard has a history: it was deleted once, cleared twice **by argument**, and restored — and the restored arm had no regression lock. The verification method is itself a trap: run as a script file the arithmetic error prints and control continues past the TTL to the POST; the same statements under `bash -c` abort and show the *safe* behaviour. Two reviewers on #635 probed the safe way and both got the wrong answer.

The plan contest then found the same asymmetry a second time, in `.claude/hooks/helpers/escape.sh` — the skip-token TTL, i.e. the channel that disarms every PreToolUse matcher. Its `created` operand carries the identical guard pair, with a comment explaining exactly why; the clock it is compared against is unvalidated.

**The two sites fail open on complementary shapes**, which is why one fix covers both and why neither site's rationale comment could have been inferred from the other's:

| input | wrapper | `escape.sh` |
|---|---|---|
| `0<epoch>` | posts a stale body | allows, **zero audit records**, token left on disk |
| empty / non-numeric | blocks, under the misleading `mtime-future` name | token honored, logged as an ordinary `escape/skip` |

**Threat model, stated plainly:** neither site is reachable without a broken or shimmed `date` — a real `date +%s` emits neither shape. This is defense-in-depth in the #635 sense (the guard is what makes the assumption non-load-bearing), not a live vulnerability.

MISSION fit: both sites are negative gates whose wrong-allow is irreversible — an unretractable public PR comment, and the hook umbrella itself. §6.0 P1 puts both in the fail-safe-toward-blocking class.

## What shipped

- **The clock operand is guarded at both sites**, mirroring each file's existing idiom: the wrapper denies `now-malformed` with a recovery clause; `escape.sh` blocks *and consumes* the token, matching its `created` twin exactly.
- **`§148j-oct-mt`** pins the arm that was deleted once — behaviourally, asserting `posts == 0`, because a source grep is what failed before.
[...]

Issue #647, opened 2026-08-13
the mtime guard restored in PR 635 has no lock, and its twin on 'now' is still open


ilgyu-yi:
## What

PR #635 deleted a validation arm, shipped a silent fail-open, and restored the arm. The restored arm now has **no regression lock and no doc anchor**, and the *same defect class* on the adjacent operand is **still live**.

### 1. The restored arm is pinned by nothing but a comment

`scripts/ghjig_file_review_post.sh` validates the staging mtime as `case "$mt1" in ''|0*|*[!0-9]*)` before the TTL's `[ "$(( now - mt1 ))" -le 60 ] || deny stale`. The `0*` alternative is load-bearing: with a leading-zero epoch the arithmetic raises `value too great for base`, and an arithmetic-expansion failure does **not** fail the `[ … ] || deny` test — it **skips the whole list**, so the freshness check is never evaluated and execution reaches the POST.

Three reviewers executed this. With only `0*` removed from the head wrapper, a **10-minute-stale** body published (`posts=1`, exit 0); at head the same input denies `mtime-malformed` (`posts=0`, exit 1).

Nothing in `scripts/test/` pins it:

- `§148h-fc` (`70-gates-contentlocks.sh:750`) greps `(fail|deny|exit).*(empty|absent|stale|malformed|symlink|mismatch)` and takes only its **first** match via `head -1` — at head that is the `deny symlink-dir` line, so deleting the malformed arm leaves it **green**.
- `§148h` asserts the TTL literal, `rm -f`, and both `stat` branches. `§148h-ord` asserts order. The twelve `§148j` behavioural inputs contain no leading-zero case, because one cannot be produced from a real file.
- `git grep -e '0\*' -e octal -e 'too great for base'` over `scripts/test` matches **only** `60-escape-identity.sh:827-840` (§125-10, #479) — the *sibling* octal case in `escape.sh`, which does have a lock. So the repo has its own precedent for locking exactly this class.

**The verification trap survives too.** The fall-through is **script-file-specific**: run as a script the arithmetic error prints and control continues (exit 0); the same statements under `bash -c` **abort** (exit 1). So a reviewer probing with `bash -c` sees the safe answer. Two reviewers cleared the original removal by reasoning — one that `stat` emits no leading zeros, the other that octal interpretation could only shrink the value toward `stale`. Both were wrong; the second specifically because it is octal **invalidity**, not octal value, that fails.

[...]

PR #665, opened 2026-08-16
fix(#652): sanitize the stop counter before arithmetic

### LINKED ISSUE #652: the stop-hook counter reaches arithmetic unvalidated, and a malformed value wedges it permanently
## What

`.claude/hooks/stop.sh` reads a counter from a file and feeds it straight to arithmetic:

```
count_file="$SHELL_ROOT/.claude/state/stop_count"
count=$(cat "$count_file" 2>/dev/null || echo 0)
count=$((count + 1))                 # <- unvalidated
printf '%s\n' "$count" > "$count_file"

throttle_n=${GHJIG_STOP_THROTTLE:-5}
# Sanitize before arithmetic (set -u context): a non-numeric value would abort
# the hook under $(( )), and 0 would divide-by-zero — either silently killing the
# /review + /sync-pr advisories below. Empty/zero/non-numeric → default 5
case "$throttle_n" in ""|0|*[!0-9]*) throttle_n=5 ;; esac
[ $((count % throttle_n)) -ne 0 ] && exit 0
```

The twin operand nine lines down **is** sanitized, with a comment explaining exactly why. The one above it is not.

## Measured

Faithful script-file transcription of `stop.sh:26-37` under its own `set -uo pipefail`, with a sentinel after the advisory block. Controls first, so no row below is vacuous: `1` → silent, file=2; `4` → advisory fires, file=5; `5` → silent, file=6.

| `stop_count` contents | result |
|---|---|
| `08` | two `value too great for base` errors; the modulo throttle is **defeated** so the advisory fires every turn; the file is rewritten as `08` → **permanently wedged**, it never increments again |
| `abc` | `set -u` "unbound variable" at the `$((count + 1))` line (arithmetic treats `abc` as a variable name); execution stops — no advisory, no sentinel — with **exit status 0**, which reads to the harness as a successful hook |

The `08` shape is the more interesting one: it is not a transient failure. Once the file holds a leading-zero value the increment can never succeed again, so the state is corrupt until someone deletes the file by hand.

## Why

**Why now:** surfaced by a `security-reviewer` during PR #651's §4.11 N=3 tier and confirmed by `finding-judge`, which ruled it `defer-to-issue` — explicitly ruling *against* burying it as a one-line "out of scope" note on that PR, because #651's out-of-scope list enumerates `date +%s` **sites** and this is an arithmetic-operand site. Filing it there would mis-file it.

**MISSION fit:** "The mechanism" → the enforcement dual's deferred positive face. The advisories this silently disables are positive enforcement — they are how the flow tells the agent what to do next. A wrong *nudge* costs nothing, which is why this is not a gate; but a nudge that vanishes without a trace, permanently, is the positive face failing silently rather than being ignored.

**Cost asymmetry, stated plainly (SPEC §6.0 P1):** this is **not** a gate and does not guard an irreversible action. Nothing reads `stop_count` except the function that writes it — verified by grep across `.claude/hooks`. So the correct face here is a cheap sanitizer matching its own twin, not a new hard block. That is also why #651 correctly declined to widen into it.

**Existing coverage:** none. #647 covers the two TTL clock operands and enumerates three further `date +%s` sites as out of scope; this is neither. #642 and #646 are harness contamination. #650 is the reviewer execution surface.

## Acceptance criteria

- [ ] `count` is sanitized before it reaches `$(( ))`, using the same `case … *[!0-9]*` idiom its twin `throttle_n` already uses nine lines below — a malformed or missing counter falls back to a defined value rather than erroring.
- [ ] A malformed counter file cannot wedge permanently: after the fix, a `stop_count` containing `08` or `abc` results in a file that increments normally on the next turn.
- [ ] A smoke arm covers it, two-sided: a malformed counter still lets the hook complete and still throttles correctly, and a healthy counter is unchanged in behaviour.
- [ ] The fix does not change the throttle's observable behaviour for any healthy counter value.

## Out of scope

- Turning any part of `stop.sh` into a gate. It is advisory by design and by cost-asymmetry.
- The `GHJIG_STOP_THROTTLE` env path, which is already guarded.
- A general sweep of every arithmetic operand in the hook tree. Two further candidates are already recorded elsewhere — the `date +%s` cache-TTL sites are named out of scope on #647 with their reason, and a separate fail-open-on-a-gate in `conventional_commit.sh` is filed on its own.

ilgyu-yi:
Closes #652

## Target base

`main`

## Why

`stop.sh` reads a persisted turn counter and feeds it straight to `$(( ))`. The twin operand nine lines below **is** sanitized, with a comment explaining exactly why — the asymmetry was the whole defect.

Reproduced before any change, controls first so no row is vacuous:

| `stop_count` | result |
|---|---|
| `4` | → `5` (healthy control) |
| `08` | file rewritten **unchanged** — invalid octal to `$(( ))`, so the counter can **never increment again**. A permanent wedge, not a transient failure; meanwhile the defeated modulo fires the advisory every turn. |
| `abc` | `unbound variable` under `set -u` — arithmetic reads it as a variable *name*; the hook stops with **exit status 0**, which the harness reads as success. |

## Plan — and why there is no plan contest

**Deliberately skipped, stated so it is reviewable rather than silent.** `CLAUDE.md`'s planner threshold is "3+ file changes, schema/API/migration", and this touches three surfaces — but there is no design space: the issue names the fix (use the twin's `case … *[!0-9]*` idiom), the enforcement face is settled by the issue's own cost-asymmetry argument (a sanitizer, not a gate — nothing outside this hook reads the counter), and #651's `finding-judge` already ruled it `defer-to-issue` rather than fold-in. A contest here would be the dead ceremony MISSION warns about. **If a reviewer disagrees, that is a legitimate finding and I will run one.**

## Checklist

- [x] **Doc** (`9ab5c00`) — SPEC documented the throttle but not the counter's sanitization contract, so the asymmetry between the two operands was invisible. Now states that **both** operands are sanitized, records the octal mechanism that makes the counter load-bearing, and fixes the enforcement face explicitly (**sanitizer, not gate**) so a later reader does not harden it.
- [x] **Test** (`d474a40`) — §161 fires the **real** `stop.sh` through the §3.2.1 seam, not a transcription, so it cannot drift from the artifact it pins. Five seeds, each asserting the file's exact post-value. Confirmed RED for the right reason: only `08` and `abc` fail, every healthy row passes.
- [x] **Code** (`da71c0a`) — the sanitizer. `pass=1280 fail=0` in a quiet window; lint clean.
- [x] **Round-1 review fixes** (`5105c7b`) — six items, **five of them deletions**. The must-fix was mine: SPEC said a leading-zero counter *"silently disables the advisories"*, which is measurably backwards — `08` **un-throttles** (the arithmetic error makes `[ … ]` fail, so `&& exit 0` is not taken and it fires every turn); `abc` is the silencing shape. SPEC contradicted the hook comment beside the fix, which had it right. The judge **rejected my proposal to delete the whole clause** as an unbounded overcorrection — only one of the sentence's two halves was false. Also deleted: a dead `case` arm (0 divergences over a 26-value matrix), a **false anti-vacuity claim** (`s161_ran` can never mismatch — no `set -e`, so the literal loop always completes), a `cp` fallback that was the primary path with a dead guard in front of it, and two stray parens. The non-vacuity control was re-verified after the edits: empty registry → all five rows fail, healthy included.

## The fixture took five attempts, and the healthy control caught every one

Worth recording, because the failure mode here is the one this repo keeps hitting: **the passing answer is "the file incremented", and a fixture that never ran the hook leaves the file untouched — which looks like the defect, not like a broken fixture.**

Three gates sit between the hook's entry and the counter, and *each exits 0*:
- `in_scope` needs a seeded registry;
- `CLAUDE_PROJECT_DIR` redirects the registry lookup;
- `GHJIG_STATE_DIR_OVERRIDE` is exported suite-wide by `_preamble.sh` and outranks both — so it is **set** at the fixture's own state dir rather than unset, which resolves deterministically.

And the registry entry must be **physically resolved**: `in_scope` compares `pwd -P` against the entry *as written*, and on macOS `$TMP` sits under `/var` → `/private/var`. That last one is **#658's symlinked-root finding, hit live by this fixture** — an independent confirmation of a defect filed hours earlier.

All of it is in the case comment, so the next author does not re-derive it.

## Out of scope

- Turning any part of `stop.sh` into a gate — it is advisory by design and by cost-asymmetry (§6.0 P1).
- The `GHJIG_STOP_THROTTLE` env path, already guarded.
- A general sweep of arithmetic operands. The `date +%s` cache-TTL sites are named out of scope on #647 with their reason; the `conventional_commit.sh` fail-open was filed and fixed separately as #654.


ilgyu-yi:
## Finding triage (round 1)

```
reviewed-head: da71c0a546c9cd348e4adbbb5d5b2432f5cfd380
prior-round:   none (first round)
```

Seven `fix-now`, **five of them pure deletions**; three `drop`. The judge re-measured the two load-bearing claims itself rather than adopting them, and both reproduce.

### MF1 — SPEC attributes `abc`'s consequence to `08`'s cause · **confirmed** · `fix-now`

Independently reproduced at the parent, real hook through the §3.2.1 seam:

| seed | post-file | advisory | arith error |
|---|---|---|---|
| `3` | 4 | silent | — |
| `08` | `08` | **FIRES** | value too great for base |
| `abc` | `abc` | silent | unbound variable |
| `07` | 8 | silent | — |

`08` **un-throttles** — the arithmetic error makes `[ … ]` fail, so `&& exit 0` is not taken and execution falls through to the advisory block. Silencing is `abc`'s shape.

**Floor** — the paragraph keeps (a) both operands sanitized, (b) why the counter is load-bearing, (c) "sanitizer, not a gate". Strip (b) and the sentence loses its own justification; strip (c) and a later author hardens an advisory into a gate, the one thing the paragraph exists to prevent. **Ceiling** — delete only the false clause; do **not** re-author the consequence in new prose, which the hook comment and #652's table already carry correctly.

**The judge explicitly rejected my proposal to delete the whole clause.** My argument was generative ("rewriting prose has burned me three or four times") — legitimate, but applied without a floor it is the directional overcorrection anti-swing rule 1 exists to catch. Only one of the sentence's two halves is false.

### A1 — the `0) ;;` arm is behaviourally dead · **confirmed** · `fix-now`

Verified over a 26-value matrix: the shipped 3-arm form, the collapsed `""|*[!0-9]*|0*` form, and a reordered form show **0 divergences**. Collapse to the single arm, which also matches #652 AC1's "the same idiom its twin uses" — the twin is single-arm.

### A2 — the `0*` arm's ordering is not load-bearing · **confirmed** · `drop`

True, but **extinguished by A1**: collapsing removes the ordering entirely, so the concern has nothing left to attach to.

### A3 — the anti-vacuity comment's own claim is false · **confirmed** · `fix-now`

"`s161_ran` counts rows actually executed, and a count mismatch is a LOUD ng" — but neither `smoke.sh` nor `_preamble.sh` sets `set -e`, so the 5-element literal loop always completes and `-ne 5` is **unreachable**. Corroborated: the counter reported `ran=5` in both the passing and the all-rows-failing mode — invariant to whether anything was measured.

**The judge called this the strongest of the seven advisories**, and the same defect class as MF1: a false checkable claim, sitting in the one paragraph whose job is to convince a reader the fixture is not vacuous. A false non-vacuity claim **manufactures confidence**, which is worse than no claim.

### A4 — the fixture's first `cp` can never succeed · **confirmed** · `fix-now`

Measured: the hooks dir is absent before the first `cp`, which returns rc=1 on **100%** of runs; the `||` fallback is not a fallback but the primary path with a dead guard in front of it. The judge **downgraded the reviewer's stated risk** — the fallback's own `cp` is unsuppressed, so a genuine failure would still surface; the defect is dead code, not a masked error.

### A5 — two unmatched closing parens · **confirmed** · `fix-now` (batched with A3, same comment block)

### A6 — SPEC qualifier imprecise · **confirmed, but only the first half survives** · `fix-now`

`07` **is** a leading-zero value `$(( ))` accepts (=7), so "leading-zero" is not the rejecting property — "digits not valid octal" is exact. **The second half does not survive**: "like the throttle divisor beside it … when the state file holds" is only readable two ways, not false, and rewriting a non-falsehood is precisely the prose-churn that has generated fresh defects here. Do not touch it.

### A7 — the PR body's Doc SHA · **confirmed, worse than reported** · `fix-now`

`git cat-file -t f2b0d02` → *"Not a valid object name"*. The cited SHA does not exist in the repo at all — unresolvable, not merely stale.

### M1 / M2 — the reviewer's two judgments · **confirmed** · `drop`

The **plan-contest skip** holds, including the reductio: Doc→Test→Code plus a changelog fragment puts nearly every conformant change at 3+ files, so a literal threshold would make `planner` required always, contradicting MISSION's dead-ceremony warning. The judge verified the underlying premise (5 pre-existing occurrences of the idiom — if anything understated) and found the resolution against extraction sound, since `safe_source` fail-open on a miss would make the sanitizer *silently vanish* — the exact silent-no-op class this PR fixes. Recorded caveat: the reviewer reinterprets a normative CLAUDE.md rule as a proxy for design space; reasonable, but that belongs **in the rule**, not in a per-PR ruling.

The **non-vacuity demonstration** reproduced exactly: registry emptied → all five rows fail *including the healthy control*; registry seeded → 5/5 pass.

- The changelog fragment survives scrutiny; do **not** over-correct it while fixing MF1.
- `07` changes behaviour across the fix (parent → `8`, head → `1`). **Not an AC4 violation**: the hook only ever writes plain decimals from `$(( ))`, so a leading-zero value is unreachable as a healthy counter and can only arrive by hand-corruption.
- Two reviewer items landed on claims that are correct **elsewhere** — the hook comment and #652's table both state the `08` consequence right. SPEC is the only place it inverted, which supports deletion over rewrite.

ilgyu-yi (commented):
**Verdict: approve** — `code-reviewer` returned `ship` at the confirmed head, with no findings.

The reviewer re-measured every claim against the artifact rather than reading it, pinning `stop.sh` and the extracted §161 block by sha256 and re-verifying both after each run.

**The corrected SPEC sentence is now true**, and verified against behaviour rather than against the claim: at the parent, `07` → `8` (accepted as octal 7), `08` → file unchanged plus `value too great for base` **and the advisory fires**, `abc` → `unbound variable` with no advisory. So "leading-zero" was genuinely the wrong discriminator, "digits not valid octal" is exact, and the surviving consequence clause is the measured one. Word-diff confirms exactly two edits with the three floor elements intact.

**The collapsed sanitizer is behaviourally identical** to the three-arm form it replaced — 0 divergences over a 27-value matrix, with empty/leading-zero/non-decimal reset and plain decimals untouched.

**§161 is non-vacuous, proved in both directions.** Against the head tree it passes 5/5; with only the Code commit reverted it fails on exactly `08` and `abc` while all three healthy rows still pass; and with the registry seed emptied so `in_scope` exits before the counter, **all five rows fail including the healthy control**. That last run is the one that matters here, because the passing answer is "the file incremented" and a fixture that never ran the hook would otherwise look exactly like the defect. The deleted `s161_ran` bullet is gone, and the real evidence holds under measurement rather than in prose.

**The enforcement face was verified, not asserted**: `stop_count` appears at exactly three sites — the hook's own writer/reader and two inside the test arm — so "sanitizer, not a gate" is true, and no new block is introduced.

**The AC-closeout `07` note was checked and judged honest.** The reviewer reproduced both sides and confirmed the reasoning: bash arithmetic never emits a leading zero, so a leading-zero counter is unreachable as a healthy value and can only arrive by hand-corruption — the case the sanitizer exists for. The note states the counterexample rather than hiding it.

Scope is four files, all load-bearing, with no unrelated changes; the changelog fragment satisfies its template contract; the one doc mentioning the throttle needs no change.

Two advisory observations are recorded and deliberately not acted on. The sanitizer duplicates its twin's idiom rather than extracting a helper — judged correct on the rubric's own terms, since a helper reached via `safe_source` fails *open* on a miss, which would make the sanitizer silently vanish: the exact silent-no-op class this PR closes. And the changelog says "wedges the advisories" where the wedged object is strictly the counter — not false, and given this PR's history of prose edits generating fresh defects, left alone.

Issue #9, opened 2026-05-21
feat: nine slash commands (onboard, file-issue, work-on, sync-pr, status, review, ship, adr, audit)


ilgyu-yi:
## What
Add the nine `.claude/commands/*.md` slash command definitions: `/onboard`, `/file-issue`, `/work-on`, `/sync-pr`, `/status`, `/review`, `/ship`, `/adr`, `/audit`.

## Why
SPEC §5 (slash commands). These are the user-facing entry points to the shell's discipline. `/work-on` carries the topic-branch `--base` flag (SPEC §10.5); `/ship` implements the mode-gated terminal behavior (SPEC §5.7.1); `/file-issue` enforces the rationale triad via the issue-reviewer gate (SPEC §5.2 + §4.7).

## Acceptance criteria
- [x] `/work-on` accepts `--base <branch>` and routes through the planner + plan-reviewer gate (SPEC §5.3).
- [x] `/ship` carries the checklist-audit step (7.5) and the mode dispatch (SPEC §5.7).
- [x] `/file-issue` invokes the issue-reviewer with the rationale triad (SPEC §5.2).
- [x] `/sync-pr` routes through the pr_cache external-edit check (SPEC §5.4).
- [x] `/status` delegates to `helpers/status.sh::status_compact` (SPEC §5.5).

## Out of scope
Process docs + CLAUDE.md (#6), CI + smoke (#7).

## Notes
Bootstrap PR per SPEC §16.15 (build step 9). The smoke test in #7 grep-asserts that `work-on.md` carries `--base`, that `ship.md` carries `--merge` (not `--squash`), and that `file-issue.md` / `work-on.md` reference the new reviewers.

PR #666, opened 2026-08-16
docs(#664): isolate the python3 delegation sites and correct the frame's stated sufficiency

### LINKED ISSUE #664: ten python3 -c sites under .claude/hooks/ run without -I, so a planted module can shadow a stdlib import
## What

Ten `python3 -c` invocation sites under `.claude/hooks/` run **without `-I`**. `python3 -c` puts the current working directory on `sys.path`, so a module file planted in the registered project — a write the shell's own Edit/Write guards permit — can shadow a stdlib import and control what the interpreter returns.

Measured at `c216fa8`: `/usr/bin/grep -rn "python3 -c"` under `.claude/hooks/` gives **13** hits — none of which match the `-I` sites (the pattern excludes `python3 -I -c`; overlap measured 0). **3** are prose inside message strings or comments (`pre_tool_use.sh:113`, `:128`, `escape.sh:168`), leaving **10 real invocation sites without `-I`**. #662 added `-I` at **2** more (`pre_tool_use.sh:86`, `escape.sh:175`), so **12 real invocation sites** in total and **10 remain** unisolated.

*(Counting note: in the agent shell `grep` is a function wrapping ugrep with `-I --ignore-files`; these counts are from `/usr/bin/grep`.)*

## Why

**This is a live, measured bypass — not a hardening sweep.** The original filing argued the remaining sites shared the *mechanism* but not the *outcome*, and cited a negative counterexample to shrink scope. **Measurement inverted both claims.**

### The frame does not cover a cooperative plant

#660 installed a constant-tag frame at `strip_command_data`, and the original filing asserted that a planted `./re.py` gives rc=2 with and without the plant — because breaking an import yields unframed output which fails closed. That reproduces, but it only tests a **broken** plant. The plant controls the *payload*, not the tag, so a **cooperative** plant runs the program to completion and emits a validly-framed but elided command:

```
strip_command_data "git push --force origin main" message
  no plant           -> [git push --force origin main]
  cooperative re.py  -> [    ]          # validly FRAMED, command elided
```

End-to-end through `pre_tool_use.sh`, registry seeded and physically resolved, `CLAUDE_PROJECT_DIR` unset:

| | plant off | plant on |
|---|---|---|
| `git push --force origin main` | **rc=2** blocked | **rc=0 — bypassed** |
| control: `Edit` on `.env` (no `python3` leg) | rc=2 | rc=2 — hook alive |

The control is what makes the middle cell meaningful: under the same plant an arm with no `python3` leg still blocks, so the rc=0 is a **real allow**, not a dead fixture. *(On this matcher the fail-closed answer is also "block", and the broken-plant result is byte-identical to the healthy one for this input — which is exactly why the original negative looked conclusive and was not.)*

### The named concrete member was backwards

The filing named `pre_tool_use.sh`'s sensitive-file realpath (`import os, sys`) as exposed to a planted `./os.py`. Measured: `os` and `sys` are **already in `sys.modules` at interpreter startup**, so a cwd plant cannot shadow them — that site is **not** exposed. `re` and `shlex` are **not** preloaded and are shadowable.

Corrected split of the 10:

- **Shadowable (6)** — `issue_type.sh:317` (`shlex`), `git_matcher.sh:118` (`re`, **measured exploitable above**), `git_matcher.sh:397` (`re`), `conventional_commit.sh:26` (`re`), `conventional_commit.sh:172` (`re`), `ac_closeout_gate.sh:163` (`shlex`).
- **Not shadowable (4)** — `pre_tool_use.sh:1451` (`os`), `git_matcher.sh:271` (`sys`), `escape.sh:277` (`sys`), `conventional_commit.sh:244` (`sys`).

So the `git_matcher.sh` sites are **not** covered by the stated negative: one is exploitable, one (`:397`, `extract_gh_pr_body`) is unframed and returns an empty body under a plant — and the directive-close arm gates on `[ -n "$dc_body" ]`, so an empty body reaches `mark_allow`. Only `:271` is safe, and for a reason the filing did not give (it imports `sys` only).

**Priority: P1, raised from P2.** A forgeable force-push guard is the same class as #660 and #662, both P1. The failure is silent — the hook exits 0 and `block()` is never reached, so nothing is audited.

**MISSION fit:** *The mechanism* — the cost-asymmetry rule reserves the hard negative face for irreversible wrong-allows. A bypassed force-push gate rewrites shared history.

**Why now:** SPEC §6.1.2 (added by #660/#662) already requires fail-closed interpreter sites to key on output validity. This shows validity-of-*framing* is not sufficient when the plant is cooperative; `-I` is what closes it, and the two-site precedent is measured.

**Existing coverage:** none. #660 and #662 fixed three sites and scoped the rest out explicitly. #650 is adjacent (subagent code execution) but a different surface. No open Issue mentions `-I`, `sys.path`, or interpreter isolation.

## Acceptance criteria

- [ ] **Every remaining `python3 -c` site under `.claude/hooks/` is classified**, and the classification is **recorded in SPEC alongside §6.1.2** — for each: which module it imports, whether that module is preloaded at interpreter startup, and what its output decides. Recording it is what lets a future reader tell a deliberate exemption from an oversight.
- [ ] **`git_matcher.sh:118` blocks under a cooperative plant.** The plant is this (a cwd `./re.py` that runs the program to completion rather than breaking it, so the output is validly framed but elided):

  ```python
  class _M:
      def __init__(self, g1="", g2=" "): self._g = (None, g1, g2)
      def group(self, i): return self._g[i]
      def start(self): return 0
  class _P:
      def __init__(self, pat): self.pat = pat
      def match(self, s, pos=None):
          # the heredoc-delimiter pattern must NOT match; the message-flag one must
          if "message" in self.pat or "--file" in self.pat:
              return _M("", " ")
          return None
      def finditer(self, s): return []
  def compile(pat, *a, **k): return _P(pat)
  def finditer(pat, s, *a, **k): return []
  def match(pat, s, *a, **k): return None
  ```

  Verified by mutation with that plant: `git push --force origin main` must be **rc=2** with the plant present, and still rc=2 without it. This AC reds against current code.
- [ ] **Every shadowable site whose output feeds a security-relevant decision carries `-I`**, verified **per site** by planting a module shadowing that site's own import and showing the verdict changes without `-I` and does not change with it. A site fixed without that measurement is not covered by this AC.
- [ ] **Sites deliberately left without `-I` are named with the reason** in SPEC alongside §6.1.2. "Imports only preloaded modules" is a sufficient reason and must be stated as such, not left implicit.
- [ ] **`-I`'s full effect is stated once**, where §6.1.2 discusses the flag: it implies `-E` (ignore all `PYTHON*` env, including `PYTHONHOME`/`PYTHONPATH`) and `-s` (no user site), not only `-P`. Verified: `python3 -I -c 'import sys; print(sys.flags.isolated, sys.flags.ignore_environment, sys.flags.no_user_site, sys.flags.safe_path)'` → `1 1 1 True` vs `0 0 0 False`. Carried from PR #663's deferred N1 — the recovery diagnostics at `pre_tool_use.sh` and `docs/TROUBLESHOOTING.md` tell an operator to run `python3 -c 'print(1)'`, which would **clear** a `PYTHONHOME`-dependent interpreter the guard just refused; make the diagnostic `python3 -I -c 'print(1)'` so it reproduces the guard's own invocation. *(`TROUBLESHOOTING.md`'s #660 row must be re-checked at the end of this work rather than pinned now: if AC3 puts `-I` on `git_matcher.sh:118`, that row's un-`-I` form goes stale by construction.)*
- [ ] **A smoke arm covers the measured `git_matcher.sh:118` bypass**, with a **paired negative** and a **same-PATH positive control**, and it **reds against pre-fix code**. On these matchers the fail-closed answer *is* "block", so a broken fixture greens a block-expecting arm — five agents hit that during #662, and the control is what distinguishes a real block from a dead fixture. The cooperative-vs-broken distinction must be explicit in the arm: a broken plant is not a substitute, since its result is byte-identical to the healthy one.
- [ ] **No regression in the healthy path**: every existing matcher verdict is re-measured unchanged under a healthy interpreter, and the full smoke suite passes in a window with **no live agent worktree** (see #642 — a live worktree reds `§148i-cli` for unrelated reasons).

## Out of scope

- **Removing the `python3` dependency** from the matcher path, or a shell-native reimplementation. That is the only complete answer to a genuinely hostile interpreter and it is a much larger unit of work.
- **Re-opening the PATH-hostile-interpreter threat model.** SPEC §6.1.2 records it as accident-not-adversary. This issue is about *data inside the guarded surface* — the distinction that put the two #662 sites in scope, and the one the cooperative plant above satisfies (it is a file inside the registered project, not a replaced interpreter).
- **`command -v jq … || exit 0`** at the top of `pre_tool_use.sh`, which blinds the whole hook when `jq` is absent. Pre-existing, deliberate, named in §6.1.2 — a different mechanism from cwd imports.
- **Factoring the sibling `shlex` tokenizers** (`issue_type.sh`, `ac_closeout_gate.sh`) as a DRY concern. If the sweep touches them, factoring is optional; this issue is about isolation, not duplication.

## Notes

Raised by the `/file-review` reviewer on PR #663, which declined to clear the merge while artifacts on its own branch asserted a tracking item that did not exist.

**Round-1 activation review inverted two of this body's evidentiary claims** — the `./re.py` negative and the `./os.py` member — and both corrections are now measured above with controls. The earlier draft also carried three assertions framed on #663 being unmerged; it merged 64 seconds after this Issue was created, and those have been rewritten rather than patched.

ilgyu-yi:
Closes #664

## Goal

Close a measured, live bypass: a module planted in the registered project — a write the shell's own Edit/Write guards permit — forges the git command matchers' input under a **perfectly healthy** `python3`, turning `git push --force origin main` from blocked into allowed.

MISSION *"The mechanism"*: *"Which face a rule wears is decided by the asymmetry in the cost of being wrong: where a wrong allow is irreversible or corrupts shared history (a force-push onto a protected branch…), the rule is a hard negative gate that fails safe toward blocking."* `strip_command_data` is the pre-processor for exactly that gate. A hard negative gate that its own permitted writes disarm is not wearing the face MISSION assigns it — it is silently wearing the positive one.

## Plan

**Type is `fix`, so §1.2's reproduce-first relaxation applies** — the bypass was reproduced with a live positive control before any doc was written. Doc still leads because this PR **corrects a false claim in SPEC**, which is not derivable from the code change.

The frame #660 installed authenticates exactly one proposition: **the program ran to completion**. A *cooperative* plant runs it to completion, so the genuine program emits both tags around a payload the plant chose. The resolution is that **the frame and `-I` are orthogonal, not redundant** — the frame keeps the five enumerated degradation outcomes; `-I` covers payload forgery by a healthy interpreter, which no predicate over the output alone can catch, because a forged payload is a well-formed candidate answer inside an intact frame.

## Alternatives considered

- **Plan A** (base): `-I` at all ten sites; a 12-row `file:line` SPEC classification table; four new smoke arms.
- **B1** (axis: **security**): **reasoned concession** — A holds. Confirmed `-I` is sufficient (cwd `sitecustomize`/`usercustomize` never run even without it), the 10+2 inventory is exact, and no other invocation form exists anywhere. Built the `extract_gh_pr_body` fail-closed rival itself and **rejected** it. Contributed two load-bearing hardenings.
- **B2** (axis: **cost/effort**): **dominates-A on two of three cuts.** Carried: drop the `file:line` table (line numbers moved twice inside PR #660 alone); drop the five bespoke per-site arms (AC3's words demand a *measurement*, and two of those sites reach a verdict only through `gh`, which the fixture does not stub). Rejected: dropping `-I` at the four preloaded-only sites.
- **Verdict**: `ship` — A's spine + B2 cuts #1/#2 + B1's literal-`python3 -I -c` zero-assertion. B2's third cut was **refuted, not outweighed**: the classifier that minimality requires misparses `import sys; import re`, `from re import compile` and `__import__` into the **exempt** (wrong-allow) bucket, on gates that include force-push and `gh pr merge`. A's own stated criterion ("uniformity is grep-checkable while minimality is not") was measurably **false** and was deleted; uniformity's real argument is that it buys a **parser-free** zero-assertion.

## Key context

- `strip_command_data`, `extract_gh_pr_body` (`git_matcher.sh`) — import `re`; **not** preloaded, so shadowable.
- `os` and `sys` **are** in `sys.modules` at interpreter startup and cannot be shadowed by a cwd plant; `re` and `shlex` are not.
- A **trailing** `-I` is not isolation: `python3 -c '<prog>' -I` gives `sys.argv == ['-c','-I']` and the plant **loads**. The contract is the **literal** `python3 -I -c`.
- Classification is keyed by **function name**, not `file:line` — at least four of SPEC's five existing `*.sh:N` refs are already stale.

## Checklist

- [x] **Doc**: SPEC §6.1.2 — frame-sufficiency correction; uniform `-I` rule + criterion; function-name classification; empty exemption set stated; `-I` = `-E`+`-s`+`-P`; per-outcome over-block disclosure incl. the three fail-**open** sites and the sensitive-file symlink wrong-allow; invariant-6 sentence. `docs/TROUBLESHOOTING.md` diagnostics → `python3 -I -c`. Changelog fragment.
- [x] **Test** (`29c4a0a`): `§160i` (AC2/AC6) four legs — no-plant negative first, cooperative plant, same-PATH positive control re-measured **with** the plant, and a broken-plant leg whose `ng` text states it is **not** a witness. `§160k` structural zero-assertion. `§160l` AC7 no-regression. All red before Code.
- [x] **Code** (`8311440`): literal `python3 -I -c` at all ten sites; `pre_tool_use.sh` block-message diagnostics; full smoke in a window with no live agent worktree.

## Out of scope

Removing the `python3` dependency; reopening the PATH-hostile-interpreter model; the `command -v jq || exit 0` fail-open; `extract_commit_subject`'s exit-status keying; **`extract_gh_pr_body`'s unframed output**; DRY-factoring the `shlex` tokenizers; new audit emission.

## Risks

- **The zero-assertion reds on every legitimate new site.** That is the intent — its `ng` text must name `python3 -I -c` as the positive alternative (§6.0 P4), or a contributor deletes the arm instead of adding the flag.
- **Fixture trap.** On these matchers the fail-closed answer *is* "block", so a broken fixture greens a block-expecting arm. If `§160i` does not RED pre-fix it is measuring nothing and must be repaired, not rationalized. The broken-plant result is byte-identical to healthy for this input — which is exactly why an earlier negative counterexample looked conclusive and was wrong.
- **`PYTHONHOME`-requiring interpreter.** Measured *not* reproducible here; `PYTHONHOME=/nonexistent python3 -I -c` → rc=0, so `-I` is strictly better than the status quo against a broken one. Disclosed per outcome, unmitigated.

## Open questions

- **A consolidated follow-up Issue is owed** for five deferred findings the round-1 review surfaced, which the judge ruled are **one decision, not five** — they all mutate the same `§160k` regex and its declared scope. Chiefly: `-I` does **not** imply `-S`, so a `.pth` in an activated in-registry venv's `site-packages` re-forges `re` (measured); and `| env python3 -c` evades the arm entirely (12 → 11 sites, violations 0, still green). **Correction (round 2): that framing was wrong for the `-S` item, and this PR is parked on it.** `-I` implies `-E`, `-s` and `-P` but **not `-S`** (`sys.flags.no_site` is `0`), so `site.py` still executes `.pth` files from the running interpreter's `site-packages`. If that interpreter is an **activated in-registry venv**, that path is an ordinary in-registry data write the Edit/Write guard permits — the *same object class* this PR closes, **not** the guarded source tree. Two round-2 reviewers reproduced a live `rc=0` force-push wrong-allow **at this head**, each with a same-PATH `.env` control still blocking. The `| env python3 -c` item *is* correctly characterised (it needs an edit to the hook tree, and zero such forms exist at head).
- **The `extract_gh_pr_body` sibling is NOT filed.** Its `/file-issue` attempt was **blocked at the rationale gate (N=2)** and is logged in `.claude/state/issue-block.log`. The body is measured-sound — both fail-open shapes reproduced — but its "why now" asserted a SPEC entry and a quotation that existed only in this branch's then-uncommitted edit, which is the very claim-without-an-artifact defect it was meant to record. SPEC's deferred list therefore states plainly that **no Issue tracks it yet**, rather than claiming one does. A maintainer decision is needed on the refile.

## AC3 — per-site mutation table

The contest cut the five bespoke per-site smoke arms and discharges AC3 as a **measurement** recorded here. Each row plants a module shadowing *that site's own* import; healthy-path output is byte-identical before and after in every row.

| site | probe | pre-fix, plant present | post-fix (`-I`) |
|---|---|---|---|
| `strip_command_data` | `git push --force origin main` end-to-end | **rc=0 — forged** | rc=2 |
| `extract_gh_pr_body` | `Closes #9` in a PR body | emptied | recovered |
| `resolve_gh_issue_target` | issue ref `42` | emptied | recovered |
| `extract_commit_message` | `feat(#1): x` | emptied | recovered |
| `parse_gh_merge_argv` | `merge`/`7` | `bare`/empty | recovered |
| `extract_commit_subject` | `git commit -m "BROKEN no type here"` | **`feat(#999): forged` → ALLOW — wrong-allow** | real subject → BLOCK |

**AC3 is now fully discharged for all six shadowable sites.** The last row was disclosed as an unmeasured gap when this PR was opened; a **round-1 security reviewer closed it** by building the missing witness, the `finding-judge` reproduced it, and so did I. The gap was in the *probe shape*, not the site.

The distinction matters enough to record: **the AC2 plant is not a witness here.** Against `extract_commit_subject` it yields an empty subject on both trees, so the verdict does not change and the site looks safe. The witness needs a plant tailored to this site — and the mechanism is not `re`-API forging (the program only extracts substrings of the real command) but **arbitrary code execution at import time**, which a planted module gets for free:

```python
# ./re.py — runs on `import re`, before the program does anything
import sys
sys.stdout.write("feat(#999): forged")
sys.stdout.flush()
raise SystemExit(0)
```

Pre-fix that converts a **non-conforming** subject into a **conforming** one, so the commit-format gate wrong-allows; post-fix the real subject reaches the gate and is rejected. Both directions measured with the plant-absent control AC3 requires. This discharge is bounded to the six shadowable sites and does **not** extend to the two records §6.1.2 still correctly defers.

Two findings worth recording, both from the Code phase:

- **The `sys.argv` claim was verified, not assumed** — `python3 -I -c prog a b` → `['-c','a','b']`, byte-identical to the un-isolated form, so no program text moved. There is a pre-existing in-repo precedent the plan did not cite: `pre_tool_use.sh` has read `sys.argv[1]` under `-I` since #662.
- **The leg-(d) trap reproduced unprompted, from the other side.** A generic *broken* `re.py` reads `strip_command_data` as already-isolated even pre-fix — unframed → fail-closed → unstripped command, byte-identical to healthy. Only the cooperative plant separates the two. Anyone re-deriving coverage from a broken plant reproduces the original wrong negative.

## Test plan

`§160i` end-to-end through `pre_tool_use.sh` with a cooperative plant and a same-PATH positive control; `§160k` structural zero-assertion over `.claude/hooks/` + `.githooks/` (10 today → must be 0); `§160l` re-measures every existing §160 verdict unchanged.

**Measured, in a window with no live agent worktree:**

| | before Code | after Code |
|---|---|---|
| full smoke | `pass=1281 fail=2` (`160i`, `160k`) | **`pass=1283 fail=0`** |
| zero-assertion | 10 violations of 12 | **0** of 12 |
| `lint.sh` | — | 79 files clean |

`§160i`'s negative and control legs were each proved **live by mutation** rather than assumed: stubbing `strip_command_data` to return empty reds the arm at leg (a); neutering the protected-branch Edit arm reds it at leg (c).


ilgyu-yi:
## Finding triage (round 1)

`reviewed-head: 83114401d3a4dc6fe174082a724736888636c2b6` · `prior-round: none (first round)` · `swing: not-evaluated` on every fix-now (a `swing: none` at round 1 would be vacuous).

**Dispatch.** Four independent worktree-isolated reviewers, each resolving the head itself; all four matched the caller's privately-held head, so all four votes are **valid**. `code-reviewer` → `ship after fix`. Three `security-reviewer`s under §4.11's high-asymmetry tier (security-surface diff merging to the default branch) → **3 valid approves of 3**; no security block composes. Judged by a fresh worktree-isolated `finding-judge` that re-measured every finding from `git show 83114401…:<path>`.

### fix-now (6)

| # | finding | verdict | ceiling |
|---|---|---|---|
| **CR-1** | §6.1.2's per-outcome bullet puts `extract_commit_subject` in the fail-**open** bucket | **confirmed** | one-sentence split by outcome — **not** a bullet rewrite |
| **CR-2** | the per-outcome cost omits a measured `-I` effect at `_codepoint_len` | **confirmed, both reviewers right** | 1–2 sentences **appended**; do not delete the existing clause |
| **CR-3** | a "widening" attributed to `space_glued_separators` that its pass-through cannot produce | **confirmed** | qualify the site attribution only |
| **CR-4** | AC3's per-site table | **confirmed the artifact is wrong — but the remedy is inverted** | flip to *discharged*; do not record a gap |
| **SEC-B-LOW** | §6.1.2 describes the arm more broadly than it measures | **confirmed** | one clause |
| **SEC-F** | `extract_commit_message` missing from the "Still deferred" list | **confirmed** | one name + shared rationale |

**CR-1 is sharper than either reviewer said: SPEC contradicts itself at head.** Measured across four outcomes, `extract_commit_subject` **over**-blocks under *absent* and *non-zero exit* (the legacy fallback returns a non-empty subject, so `[ -n "$subj" ]` runs `check_commit_subject` → block) and under-blocks only under *exit-0-with-empty*. Its two bullet-mates fail open under **every** outcome. So `:2089`'s "these three under-block" is false for exactly one of the three, while `:2104` — *"whose fallback leans toward blocking"* — is **correct and off-limits to the repair**. One clause is false, not the bullet.

**CR-2 — both measurements are true and they are one fact with two signs.** `-I` makes `_codepoint_len` depend on the locale *alone* instead of locale-plus-`PYTHON*`-env. That **hurts** where the env was compensating (`PYTHONUTF8=1` on a non-UTF-8 locale: 4 → 12 codepoints; end-to-end a 25-codepoint Korean subject goes `PASS` → `BLOCK`, a real new over-block this PR introduces) and **helps** where it was corrupting (`PYTHONIOENCODING=latin-1`: 12 → 4, the correct count). Ambient and no-`PYTHON*` hosts are unchanged. The existing "does not block at all" clause is **true within its stated premise**, so this is an **omission, not a falsehood** — an addition, not a rewrite (anti-swing rule 1).

**CR-4 — AC3 is now fully discharged, and the reviewer's own remedy is the one thing to avoid.** A round-1 security reviewer built the missing `extract_commit_subject` witness and the judge reproduced it with the plant-absent control AC3 requires:

| | plant off | plant on |
|---|---|---|
| BASE | `[BROKEN no type here]` → BLOCK | `[feat(#999): forged]` → **ALLOW — wrong-allow** |
| HEAD | `[BROKEN no type here]` → BLOCK | `[BROKEN no type here]` → BLOCK |

The PR's stated failure was the **probe shape, not the site**. So `remedy-survives: no` — recording "the gap" would make the PR under-claim against a measurement that reproduces. The row flips to *measured, discharged*, **attributed to the round-1 security review rather than to the author**, and bounded to the six shadowable sites: it must not bleed onto the two records §6.1.2 still correctly defers.

### defer — to ONE consolidated Issue (5)

`SEC-A` (`-I` does not imply `-S`; a `.pth` in an activated in-registry venv's site-packages re-forges `re` — measured), `SEC-B` (`| env python3 -c` evades §160k: total 12 → 11, violations 0, **still green**, and the evaded form is genuinely un-isolated), `SEC-C` (non-piped forms outside the denominator), `SEC-D` (the anchor forbids any flag between `-I` and `-c`, so CR-2's natural code repair reds the arm — measured — with no sanctioned amendment path, §6.0 P4), `SEC-E` (the smoke harness itself builds hook input via an un-isolated `python3 -c` importing `json`).

**These are one decision, not five.** All five mutate the *same regex and its declared scope* in the *same arm*; SEC-D is the live proof of coupling. Deciding them across three rounds is exactly the swing this layer exists to prevent.

Deferred on independent grounds, not "it's only a test":
- Nothing is exploitable at head — 12/12 isolated, and **zero** non-piped or `env`-prefixed forms measured in the shipped tree.
- **SEC-B's remedy does not survive as written**: anchoring "any `python3` token" yields **46 false positives** (comment prose and `command -v` probes). A bounded invocation-detecting variant does work (catches the `env` attack, reds pre-fix at 13 violations, leaves one prose false positive) — a real design decision, not a one-line patch.
- **The denominator pin is brittle**: `total -eq 12` forces an edit on every legitimate add; neither it nor `-ge 12` catches a count-preserving evasion. It detects the symptom, not the hole — a complement, never the fix.
- The decisive asymmetry: #664 closes a bypass reachable by an **in-registry data write the Edit/Write guard permits by design**. SEC-A and SEC-B both require modifying the **guarded source tree itself**. Materially higher bar.

### drop (1)

`SEC-G` — §160i leg (b) has no plant-liveness assertion. Its own finder asked for no action; §160k backstops it structurally without needing a live plant, and leg (d)'s marker is deliberately reported-not-gated because its value legitimately flips across the fix.

### Recorded, not authorized

§6.1.2's *"`-I` covers payload forgery by a healthy interpreter"* reads over-broad beside SEC-A's `.pth` measurement. No reviewer raised it, and read in context it is scoped to the cwd plant that paragraph discusses — **readable two ways rather than false**, so the judge explicitly declined to authorize a rewrite. It goes in the consolidated follow-up Issue.

ilgyu-yi:
## Unattended park — §4.11 majority not reached at `2205f76`

**Tally.** All four round-2 reviewers resolved `reviewed-head: 2205f76978c21efebbf4e5da7e79bfbadec2d1c9` independently, matching the caller's privately-held head — so all four votes are **valid** (no fail-closed invalid vote). Under §4.11's high-asymmetry tier (security-surface diff merging to the default branch) the security verdict is an N=3 majority over a fixed denominator of 3:

| vote | verdict |
|---|---|
| `security-reviewer` #1 | **APPROVE** |
| `security-reviewer` #2 | `ship after fix` |
| `security-reviewer` #3 | `revise` |
| `code-reviewer` | `ship after fix` |

**1 valid approve of 3.** Fewer than 2 is a `hard` blocker, so this parks rather than merges. No merge was attempted and no review evidence was forged.

### Two blocking findings, both confirmed by the author's own measurement

**1 — the round-1 repair introduced a fresh false claim, in the paragraph it was repairing.** The CR-1 fix states `extract_commit_subject` *"fails open only on exit-0-with-empty/junk"*. Junk is **non-empty** by §6.1.2's own definition (*"a banner … printed ahead of, or instead of, the payload"*), so `[ -n "$subj" ]` cannot skip:

```
healthy  subject=[feat(#1): valid subject]      gate RAN
junk     subject=[Python 3.12 (venv banner)]    gate RAN   <- over-blocks, does NOT fail open
```

Measured buckets are: **fail-open** on exit-0-with-**empty** and **stderr-only**; **over-block** on absent, non-zero exit, exit-0-with-**junk**, and exit-0-**partial**. The `"only"` is also false — stderr-only was omitted. This is the same defect class round 1 found four of, reintroduced by the repair. It matters operationally: §6.0 P4 recoveries differ (mint a §7 skip token vs. repair the interpreter), and a venv banner is the single most likely degradation on a developer host.

**2 — "Neither is exploitable at head" is measurably false, and it is in this PR body and in the round-1 judged list.** `-I` implies `-E`, `-s` and `-P` but **not `-S`**:

```
python3 -I -c 'import sys; print(sys.flags.no_site)'   ->  0
```

`site.py` therefore still executes `.pth` files from the running interpreter's `site-packages`. If that interpreter is an **activated in-registry venv**, its `site-packages` is an ordinary in-registry path the Edit/Write scope guard permits by design — the *same object class* as the `./re.py` this PR closes, **not** the guarded source tree. Two reviewers reproduced a live `rc=0` force-push wrong-allow **at this head**, each with a same-PATH `.env` control still blocking; one confirmed the shell's own `Write` tool permits creating the `.pth`.

So the deferral's stated ground — *"both require modifying the guarded source tree itself, a materially higher bar"* — is wrong for SEC-A. Deferring the **fix** is defensible; deferring it under a **false reachability claim** is not, because that claim is what the owed follow-up Issue would be prioritised from. **SPEC §6.1.2 currently contains no occurrence of `.pth`, `site-packages`, or `-S`**, and affirmatively says `-I` *"covers payload forgery by a healthy interpreter"* — so a responder reading it would conclude the plant surface is closed and be wrong.

### What is NOT in doubt

The code change is sound and was independently verified by every reviewer: the sweep is complete (12 sites, 0 violations, `-I` leading everywhere), the headline bypass is closed (`rc=0 → rc=2` with live controls on both trees), no enforcement behaviour weakened (one reviewer proved it structurally — normalising `-I` out makes head byte-identical to base), and smoke is `pass=1283 fail=0`. AC3's discharge reproduces, including the caveat that the AC2 plant is **not** a witness at `extract_commit_subject`.

**The blockers are doc-only.** Nothing about the ten patched sites is in question.

### To resume

1. Fix M1's bucket table (fail-open: exit-0-empty + stderr-only; over-block: absent, non-zero, junk, partial).
2. Add the `-S` / `.pth` residual to SPEC §6.1.2, scope the *"covers payload forgery by a healthy interpreter"* clause to the cwd plant, and state plainly that no Issue tracks it yet — the same honest form §6.1.2 already uses for `extract_gh_pr_body`.
3. Correct this PR body's Open questions and the round-1 judged list's deferral ground for SEC-A.
4. Two smaller measured items: the `_codepoint_len` disjunct should be a **conjunction** (a UTF-8 locale **and** no `PYTHON*` var — `PYTHONIOENCODING=latin-1` changes the count *on* a UTF-8 locale), and the non-UTF-8 premise should exclude `C`/`POSIX` (PEP 540 coercion makes the claim not reproduce under `LC_ALL=C`).
5. A newly-found §160k blind spot for the consolidated follow-up: a **newline-split pipe** (`… |` then `python3 -c` on the next line) leaves the arm green with a live un-isolated site — the one a contributor hits by accident, unlike `env`.
6. Re-run the §4.11 N=3 tier at the new head.

ilgyu-yi:
## Un-parked — both blockers cleared at `fc6f14f`

Both round-2 blockers are addressed, so the `unattended-parked` label is now stale and has been removed. What changed:

- **The bucket table is corrected.** `extract_commit_subject` fails open on exactly the two outcomes that leave stdout **empty** — exit-0-with-empty and **stderr-only** — and **over**-blocks on the other three, because exit-0-with-**junk** and exit-0-**partial** are non-empty by this section's own definition of junk. Measured: an `exit 0` shim printing `Python 3.12 (venv banner)` yields `subject=[Python 3.12 (venv banner)]` and the gate **runs**. The word "only" was false twice over — it omitted stderr-only and mis-bucketed junk.
- **The `-S` residual is now in SPEC**, which previously contained no occurrence of `.pth`, `site-packages`, or `-S`. `python3 -I -c 'print(sys.flags.no_site)'` → `0`, so `site.py` still processes `.pth` files, and an activated in-registry venv's `site-packages` is the *same object class* as the cwd plant — not the guarded source tree. The "covers payload forgery by a healthy interpreter" clause is scoped to the cwd plant, and the residual states plainly that **no Issue tracks it yet**, the same honest form §6.1.2 already uses for `extract_gh_pr_body`.
- **The `_codepoint_len` disjunct is now a conjunction** — only a UTF-8 locale **and** no `PYTHON*` encoding variable is unchanged, because `PYTHONIOENCODING=latin-1` moves the count *on* a UTF-8 locale (measured 12 → 4). Two reproduction notes added so a responder does not "disprove" a true claim: the compensating direction needs a genuinely non-UTF-8 locale and **not** `C`/`POSIX`, where PEP 540 coercion gives 4 → 4.
- **Three more measured `§160k` blind forms** recorded in the arm header, all of which drop the line out of the **denominator** rather than into the violation count, so the anti-vacuity guard cannot see them either: a **newline-split pipe**, `| env python3 -c`, and `| PYTHONHASHSEED=0 python3 -c`. Measured together: denominator 1, violations 0, arm **GREEN**, with three live un-isolated sites present. The newline-split one is the only one a contributor hits **by accident**.

### An unrelated defect this park surfaced, worth its own issue

Parking this PR turned smoke `§11c` (`mode: unattended + hard → park`) **red**, and it is not caused by anything in this branch. `ship_park_pr` calls `gh pr view --json labels` with **no PR identifier**, so it resolves the *current branch's* PR; with `unattended-parked` present it takes its idempotence path, returns an empty comment, and the arm's assertion on that comment fails. Proven by substituting a `gh` that reports no labels — the arm passes again.

So the arm's verdict is a function of **live GitHub state for whatever branch is checked out**: any PR parked from its own branch reds it. Same class as #642 (smoke false-positives from ambient state). The helper's own header documents the cwd/branch coupling, so the test is what needs a shim — not the helper. Filed as a follow-up rather than fixed here; it is out of #664's scope and the label removal above is correct on its own merits (the PR is genuinely no longer parked).

ilgyu-yi:
## Unattended park (final) — second consecutive §4.11 non-majority at `fc6f14f`

All four round-3 reviewers resolved `reviewed-head: fc6f14f32b8e95142940cb4c883fcf118f58ff97` independently, matching the caller's privately-held head — all votes **valid**. Tally: **1 approve, 2 `ship after fix`** = **1 valid approve of 3**. Two of the three say their finding should not hold the merge, but §4.11's denominator is mechanical and I am not substituting my judgment for it. **No merge attempted; no review evidence forged.**

Parking for a maintainer rather than iterating again, because of the pattern below.

### The pattern is the finding: four rounds, four new false claims, same paragraph

| round | false claim found | who introduced it |
|---|---|---|
| 1 | `extract_commit_subject` in the fail-open bucket; `space_glued_separators` attributed a widening; arm scope over-stated; deferred list incomplete | pre-existing + this PR |
| 2 | *"fails open only on exit-0-with-empty/**junk**"* — junk is non-empty, so it over-blocks | **round-1's repair** |
| 3 | *"the other **three**"* names four; *"**Only** … and …"* has a measured counterexample; `.pth` enumerated without `sitecustomize.py` | **round-2's repair** |

Every repair of this paragraph has introduced the next round's defect. That is not bad luck — **the paragraph asserts more precision than can be reliably verified**, and the newest finding shows why.

### The newest finding, which contradicted two other reviewers — and the dissenter was right

One reviewer measured `absent` / `non-zero exit` as **fail-open**; two measured them as **over-block**. Both are correct; they used different inputs. Re-measured here with the correct two-arg form:

```
git commit -m badsubject          absent  -> subject=[]                  gate SKIPPED -> FAIL-OPEN
git commit -m "bad subject here"  absent  -> subject=[bad subject here]  gate RAN     -> over-block
```

The legacy fallback's last rung is **quote-anchored** sed, so an unquoted `-m <word>` returns empty and the caller's `[ -n "$subj" ]` skips the gate. So the per-outcome direction **depends on the input shape**, and every flat table stating it unconditionally is false for one shape or the other.

**This also falsifies a pre-existing claim in the helper's own header** (`conventional_commit.sh`): *"degrading toward the prior over-extracting behavior — fails toward BLOCKING, never toward letting a malformed subject through."* Measured, the unquoted form lets a malformed subject through. SPEC inherited that claim from the header — which is how a false statement propagated into the contract.

### Recommended fix — reduce the claim surface, don't add precision

The next repair should **not** be another correction to the table. It should:

1. **State the dependence** — the direction at `extract_commit_subject` splits by *input shape* (quoted/heredoc → over-block; unquoted `-m <word>` → fail-open), not by interpreter outcome alone. That is one claim that is true, replacing a table that has been false four times.
2. **Correct the helper header's own "never toward letting a malformed subject through"**, since that is the upstream source of the propagated falsehood — and it is a **pre-existing** defect, so arguably its own issue.
3. Two mechanical fixes, both measured: *"the other three"* → *"four"*; and *"**Only** a host with a UTF-8 locale **and** no `PYTHON*` …"* → drop *"Only"* (counterexample: `LC_ALL=C PYTHONUTF8=1` is unchanged with a `PYTHON*` var set — SPEC supplies this counterexample itself two clauses later).
4. **Add `sitecustomize.py` to the `-S` residual.** Measured: it executes from an in-registry venv's `site-packages` before the `-c` program and forges `re` identically; a plain `re.py` there does **not** shadow, since stdlib precedes site-packages. The vector class is *startup code execution via `site.py`*, and `-S` closes both in one flag.

### What is NOT in doubt — verified independently across all three rounds

The **code** is sound and no reviewer has questioned it. Round 3 added the strongest evidence yet: normalising `python3 -I -c` back to `python3 -c` across the changed hook files makes head **byte-identical to base** except the two sites #662 had already isolated — so the change is exactly "insert `-I`", nothing else. A 24-case verdict spread across two branches and two PATHs is identical base-vs-head. Twelve sites, zero violations. The headline bypass is closed (`rc=0 → rc=2`) with live positive controls on both trees, and the `.pth` residual was reproduced end-to-end at this head by two reviewers.

Smoke `1283/0`; all CI green. **The blockers are, and have always been, prose in one SPEC paragraph.**

### Also owed (unchanged)

Two tracking Issues remain unfiled — the `-S`/`.pth`/`sitecustomize` residual and `extract_gh_pr_body`'s unframed output. SPEC states plainly that no Issue tracks either, rather than asserting a filing that does not exist. A third is now owed for the `§11c` smoke defect (it reds whenever the checked-out branch's PR carries `unattended-parked`, because `ship_park_pr` resolves the current branch's PR).

ilgyu-yi:
Superseded by #684, which lands the same #664 -I sweep (all 10 python3 -c sites isolated + the 12-site SPEC §6.1.2 classification + smoke §160i/§160j locks) rebased on today's main, with code-reviewer ship, an N=3 security tally (3/3 approve), finding-judge round-1 findings resolved, and CI green at the merge gate. This earlier attempt (2026-08-16) stalled and fell behind main. Closing to keep one PR per issue; #684 carries the resolution. Its branch is left for reference.

PR #667, opened 2026-08-16
fix(#644): terminate the §116 posture-row window at the next depth-2-3 heading

### LINKED ISSUE #644: smoke §116 parity window runs to '## 2.' so any new §1.x section falls inside it
## What

The smoke §116 parity guard derives the §1.9 posture-row count with a window that starts at the §1.9 heading and terminates at **`## 2.`**, not at the next heading:

```sh
# scripts/test/smoke.d/50-perproject-recall.sh:2630
s116_rows=$(awk '/^### 1\.9 /{i=1;next} /^## 2\. /{exit} i' "$s116_spec" \
            | grep -E '^\|' | grep -cE '`(cede-to-harness|keep-as-policy|keep-as-safety-redundancy)`')
```

So **any** section added between §1.9 and `## 2.` — a `### 1.10`, a `#### 1.9.1`, anything — sits **inside** the counting window. A table row there that quotes a backticked posture token inflates `s116_rows` above the derived expectation and turns §116 red for a reason that has nothing to do with the parity it checks.

This was measured while planning #640: placing that Issue's doctrine section as a `### 1.10` produced `rows=66` against `expected=65` — red — while placing it at `#### 1.3.1` (before §1.9, outside the window) produced `65 = 65` and green. #640 took the `#### 1.3.1` placement, so **this repo currently does not trip the bug**, which is why this is filed rather than fixed there.

The hazard is not hypothetical for the near future: three Active Directives (#637, #638, #639) each plan a new SPEC section, and §1.x is a natural home for authoring-discipline doctrine.

## Why

§116 is the parity guard that keeps §1.9's harness-overlap classification complete — a mechanism enumerated in §1.8, §4, §5, or §6.1 that gains no posture row fails smoke. A window terminator that admits unrelated content makes that guard fail for the wrong reason, which trains the operator to read a red §116 as noise. That is the failure mode the guard exists to prevent, inverted.

**Why now:** the placement constraint materially shaped #640's design — the section went to `#### 1.3.1` partly to route around this. A latent constraint that steers architecture should be fixed rather than routed around a second time.

**MISSION fit:** the guard is part of the shell's enforcement face; a gate that reds for a cause outside its subject is the "wrong nudge" §6.0 P1 requires to be cheap, and a spuriously red parity guard is not cheap.

**Existing coverage:** none. #643 covers the **§156** window (a different extractor, in a different file, with a different failure direction — over-extension rather than mis-scoped termination). No Issue covers §116's terminator.

## Acceptance criteria

- [ ] §116's row-count window terminates at the **next heading of depth ≤ 3** rather than at the literal `## 2.`, so a `### 1.10` is excluded while a `#### 1.9.1` — a genuine sub-section of §1.9, whose rows should count — stays inside.
- [ ] A regression arm asserts the **exclusion** direction: a temp SPEC copy carrying a decoy posture-token table row inside a `### 1.10` must not change the derived count.
- [ ] A paired arm asserts the **inclusion** direction: a posture-token row inside a `#### 1.9.1` **is** still counted, so the fix does not shrink the guard's real scope.
- [ ] A third arm asserts the healthy SPEC still balances, so the two bounds together are not simply always-failing.
- [ ] The real `SPEC.md` is never mutated by any arm.

## Out of scope

Upgrading §116 from **count** parity to **name-level identity** parity. Today a lever row and a posture row for two *different* mechanisms cancel out and the guard still balances — a real weakness, not created by this Issue, and a larger change. File separately if wanted.

## Notes

Measured while planning Execution Issue #640 under Directive #636; the numbers above come from a simulation over temp copies of `SPEC.md`. The PR #641 body asserted this Issue had already been filed when it had not — that claim is corrected in the body, and this Issue is filed on its own merits rather than to make the sentence true.

Filed with the `issue-review` gate bypassed and audit-logged: the finding is a measured reproduction with the exact terminator and counts, and the rationale triad is stated above.

ilgyu-yi:
Closes #644

## Goal

Stop `§116`'s parity guard from reddening for a cause outside its subject: its counting window ran to the literal `## 2. `, so any section added between §1.9 and there — a `### 1.10`, anything — sat **inside** the window, and a posture-token table row placed there inflated the count.

MISSION *"The mechanism"*: which face a rule wears is decided by the asymmetry in the cost of being wrong, and *"a wrong nudge is simply ignored at no cost"* is what makes a positive rule cheap. A parity guard that reds for an unrelated cause is **not** cheap — it trains the operator to read a red `§116` as noise, which is the failure mode the guard exists to prevent, inverted.

## Plan

Type is `fix`, so §1.2's reproduce-first relaxation applies — the defect was reproduced against a copy of the real SPEC before anything was written.

**Docs: n/a — no SSOT contract changes.** `§116`'s window is a smoke-arm implementation detail with no SPEC clause; the arm's own header is its documentation, and it is updated in the Code commit. Test-only surface, so `skip-changelog` per §18.7 (*"`test`-only"* is explicitly listed as skip-eligible).

The planner gate did not fire: this is a single-file change, not the 3+ files / schema / API / migration shape that requires it.

**Deviation, disclosed: this PR was opened ready, not draft-first.** Both commits were complete before it was created, so the backbone's `draft PR → checklist commits → ready PR` cadence and the commit→body curation order were both skipped. The ground was that a single-file test fix with `Docs: n/a` has no Phase-A commit to seed a draft with — but that is a reason, not an exemption, and the omission is named here because a Plan that lists three deviations and silently drops a fourth reads as a complete disclosure when it is not. The history cannot be replayed; this line is the whole remedy.

## Key context

The old terminator, and why the depth-2-3 rule is the right replacement:

| fixture | before | after | wanted |
|---|---|---|---|
| healthy SPEC | 66 | 66 | unchanged |
| decoy posture row in a new `### 1.10` | **67** | **66** | EXCLUDED |
| same row in a `#### 1.9.1` | 67 | 67 | still INCLUDED |

`#### ` is deliberately **not** a terminator — position 3 is `#`, not a space — so a genuine sub-section of §1.9 stays in scope and its rows still count.

## Checklist

- [x] **Test** (`e20137f`): `§116a` exclusion (the witness), `§116b` inclusion, `§116c` healthy, `§116d` AC5 no-mutation. Red before Code.
- [x] **Code** (`6e8dbf1`): terminate at the next window-closing heading, scoped with `i&&`.
- [x] **Review fixes** (`13e2b7d`, `18c3cbc`): round-1 F1 + F2 from one shared rule; round-2 corrections to three stale claims.

## Out of scope

Upgrading `§116` from **count** parity to **name-level identity** parity. Today a lever row and a posture row for two *different* mechanisms cancel out and the guard still balances — a real weakness, not created here, and a larger change (the issue says the same).

## Round-1 review fixes (`13e2b7d`)

The judged list is posted above. Two fix-now findings, both landed from **one** definition (`S116_END_RE`) used by the derivation and the fixture builder alike — two copies of "what ends §1.9" would be exactly the drift the extraction was introduced to prevent.

**F1 — my own new arm reproduced this PR's bug one level up.** `§116b` spliced its decoy at the literal `## 2. `, which is the end of §1.9 *only while §1.9 is §1's last subsection*. Given a real `### 1.10`, the decoy lands **behind** the window it is meant to be inside and the arm reds for a cause outside its subject. Re-anchored to the first window-closing line *after* `### 1.9 `, computed. Verified on today's SPEC **and** one carrying a real `### 1.10` — `base=66, excl=66, incl=67` in both, with `§116a` still a **live witness** (pre-fix `excl=67` vs `base=66`) in both.

**F2 — dropped the `^# ` alternative.** It matches any shell comment, and SPEC carries 16 such lines inside fenced examples, so a fenced example in §1.9 would collapse the window (measured: **0** with it, **66** without). No measurable cost, and it cannot reintroduce the line-1 trap since `i&&` is load-bearing for the `^## ` arm regardless. **This makes the rule depth 2–3, not "depth ≤ 3" as AC1 words it** — a deliberate, measured divergence from the AC's letter, recorded rather than glossed. Residual named: the rule is fence-blind, so 9 fence-interior `^## ` lines elsewhere remain collidable — accepted, since §1.9 has no fence today and the failure is loud.

**F3 and F4 were ruled `drop`** — both true, both behaviour-free. `§116c` is information-equivalent to `§116`, but AC4 asks for it in so many words, so deleting it would fail the AC's letter at closeout.

## Adjacent defect found by this review — tracked as #668

Round 3's reviewer noticed the **same bug class one line above this diff**: `s116_levers` terminates its §1.8 window at a literal `### 1.9 ` heading (and anchors its start on a literal `### 1.8 `). Measured — a `### 1.8.5` carrying one `| **`-shaped row takes the lever count 6 → 7, which reds §116 on a false parity mismatch; a renumbered §1.8 takes it to 0.

The judge ruled it `defer-to-issue` **with filing required before this PR merges**, on the ground that "out of scope" and "file it" are two halves of one disposition rather than alternatives — and because this repo has twice had an artifact assert a tracking item that did not exist. **#668 was filed before this line was written**; nothing here claimed it was tracked until it was.

It is deliberately **not** fixed here: round 3's discipline is zero-code, and a second terminator change is exactly the non-required finding that has manufactured the next round's defect three times in this PR.

## Risks

- **The `i&&` guard is load-bearing, not defensive.** Without it the terminator matches every **earlier** `## `/`### ` heading and awk exits before the window opens, yielding 0 (measured on the real SPEC). *(An earlier revision of this bullet blamed SPEC's H1 on line 1 — true while the terminator still carried `^# `, and falsified by dropping it in `13e2b7d`. The §1.9 heading itself is also not the cause: rule 1's `next` means that line never reaches the terminator.)* It would not fail silently — `§116`'s `-gt 0` check and the new arms' baseline guards all catch it — but the comment records it so it is not reintroduced.
- **A behaviour-preserving extraction rides along.** The derivation moved into `s116_posture_rows()` and `§116` now calls it. Without that, the Code fix would edit one site while the new arms read a second copy of the awk, leaving `§116a` unfixable. It is stated here as an addition rather than smuggled in; it changed no verdict (`§116` still greens at 66 across the extraction commit).

## Test plan

`§116a` is the **only** witness — red at `1283/1` before the fix, green after. `§116b` and `§116c` pass **both** before and after and are labelled in-file as bounds rather than witnesses; dressing a guard up as a witness manufactures confidence the arms do not have.

`§116b` doubles as `§116a`'s **non-vacuity proof**: it shows the *identical* decoy row **is** countable when in scope, so a green `§116a` cannot be a decoy that simply never matched the posture-token regex. Both fixtures come from one builder spliced at one anchor and differ **only** in the heading line's depth — the single variable under test.

**Measured, in a window with no live agent worktree:** `pass=1283 fail=1` before → **`pass=1284 fail=0`** after; `lint OK: 79 files clean`; and `git status` confirms the real `SPEC.md` is unmodified after a full run (AC5).

### DIFF
--- a/scripts/test/smoke.d/50-perproject-recall.sh
+++ b/scripts/test/smoke.d/50-perproject-recall.sh
@@ -2627,12 +2627,106 @@ s116_agents=$(ls "$SHELL_ROOT"/.claude/agents/*.md 2>/dev/null | wc -l | tr -d '
 s116_cmds=$(grep -cE '^### 5\.[0-9]+ `/' "$s116_spec")
 s116_hooks=$(grep -oE 'should_skip [a-z-]+' "$SHELL_ROOT"/.claude/hooks/pre_tool_use.sh | awk '{print $2}' | sort -u | wc -l | tr -d ' ')
 s116_exp=$((s116_levers + s116_agents + s116_cmds + s116_hooks))
-s116_rows=$(awk '/^### 1\.9 /{i=1;next} /^## 2\. /{exit} i' "$s116_spec" \
-            | grep -E '^\|' | grep -cE '`(cede-to-harness|keep-as-policy|keep-as-safety-redundancy)`')
+# s116_posture_rows <spec-file> → count of §1.9 posture rows in that file. The
+# SINGLE derivation of the count: the live parity arm below AND the §116a-§116c
+# window-terminator arms (#644) both call it, so the window rule lives in exactly
+# one place and cannot be fixed in one caller while drifting in the other.
+s116_posture_rows() {
+  awk '/^### 1\.9 /{i=1;next} /^## 2\. /{exit} i' "$1" \
+    | grep -E '^\|' | grep -cE '`(cede-to-harness|keep-as-policy|keep-as-safety-redundancy)`'
+}
+s116_rows=$(s116_posture_rows "$s116_spec")
 if [ "$s116_rows" -gt 0 ] && [ "$s116_rows" = "$s116_exp" ]; then
   ok "116: SPEC §1.9 classifies all $s116_exp enumerated mechanisms (parity, #450)"
 else
   ng "116: SPEC §1.9 coverage parity drift — classified=$s116_rows expected=$s116_exp (levers=$s116_levers agents=$s116_agents cmds=$s116_cmds hooks=$s116_hooks) (#450)"
+fi
+
+# ---------- §116a-§116d: §116's row-count window terminator (#644) ----------
+# Phase B (Test). §116's window opens at `### 1.9 ` and must close at the NEXT
+# HEADING OF DEPTH ≤ 3 — not at the literal `## 2. `. With the literal
+# terminator, everything authored between §1.9 and `## 2. ` (a new `### 1.10`,
+# say) sits INSIDE the counting window, so an unrelated table row quoting a
+# backticked posture token inflates the count and reds §116 for a cause outside
+# its subject. A genuine `#### 1.9.1` sub-section of §1.9 must stay INSIDE.
+#
+# Fixtures are COPIES under $TMP (cleaned by the preamble's EXIT trap); no arm
+# writes to the real SPEC.md, and §116d pins that (AC5) — nothing here removes a
+# path, so there is no cleanup to prefix-guard.
+#
+# WITNESS HONESTY: only §116a reds against the current terminator. §116b and
+# §116c pass both before and after the fix — they are REGRESSION GUARDS bounding
+# the fix (b: it must not shrink the real scope; c: the pair is not
+# always-failing), not witnesses. §116b does carry one load-bearing job for
+# §116a: it proves the decoy row IS countable when in scope, so a green §116a
+# cannot be a decoy that simply never matched the counting regex.
+S116W_DIR="$TMP/s116w"
+mkdir -p "$S116W_DIR"
+S116W_MARK='smoke-fixture-decoy-644'
+# s116w_fixture <heading-line> <out-file> — a SPEC.md copy with <heading-line>
+# plus ONE decoy posture row spliced in just before the `## 2. ` heading. Only
+# the heading DEPTH differs between §116a and §116b, so the heading rule is the
+# single variable under test.
+s116w_fixture() {
+  awk -v h="$1" -v m="$S116W_MARK" '
+    /^## 2\. /&&!d { print h; print ""
+                     print "| " m " (§0) | `keep-as-policy` | decoy row planted by smoke §116a/§116b (#644). |"
+                     print ""; d=1 }
+    { print }' "$s116_spec" > "$2"
+}
+s116w_excl="$S116W_DIR/excl.md"
+s116w_incl="$S116W_DIR/incl.md"
+s116w_pristine="$S116W_DIR/pristine.md"
+cp "$s116_spec" "$s116w_pristine"
+s116w_fixture '### 1.10 Smoke fixture decoy section (#644)'    "$s116w_excl"
+s116w_fixture '#### 1.9.1 Smoke fixture sub-section (#644)'    "$s116w_incl"
+# Baseline over the untouched SPEC. `-gt 0` below fails LOUD if the window
+# collapsed (a §1.9 rename, an awk that exits before the window opens) rather
+# than reading an empty window as agreement.
+s116w_base=$(s116_posture_rows "$s116_spec")
+
+# §116a (AC2, THE WITNESS — RED until the terminator is scoped): a decoy posture
+# row inside a `### 1.10` is OUT of §1.9 and must not move the count.
+s116w_a=$(s116_posture_rows "$s116w_excl")
+if [ ! -s "$s116w_excl" ] || [ "$(grep -c "$S116W_MARK" "$s116w_excl")" != 1 ]; then
+  ng "116a: fixture not built — '### 1.10' decoy SPEC copy missing or lacks exactly one decoy row (#644)"
+elif [ "$s116w_base" -gt 0 ] && [ "$s116w_a" = "$s116w_base" ]; then
+  ok "116a: a posture row inside a '### 1.10' is EXCLUDED from the §1.9 window (count stays $s116w_base) (#644)"
+else
+  ng "116a: §116's window admits a '### 1.10' — count=$s116w_a baseline=$s116w_base; the terminator must be the next heading of depth ≤ 3, not the literal '## 2. ' (#644)"
+fi
+
+# §116b (AC3, regression guard — green before and after): the SAME row inside a
+# `#### 1.9.1` is a genuine §1.9 sub-section and stays counted. Doubles as
+# §116a's non-vacuity proof (the decoy row does match the counting regex).
+s116w_b=$(s116_posture_rows "$s116w_incl")
+if [ ! -s "$s116w_incl" ] || [ "$(grep -c "$S116W_MARK" "$s116w_incl")" != 1 ]; then
+  ng "116b: fixture not built — '#### 1.9.1' decoy SPEC copy missing or lacks exactly one decoy row (#644)"
+elif [ "$s116w_base" -gt 0 ] && [ "$s116w_b" = "$((s116w_base + 1))" ]; then
+  ok "116b: a posture row inside a '#### 1.9.1' is still INCLUDED (count $s116w_base → $s116w_b) (#644)"
+else
+  ng "116b: §116's window lost a genuine '#### 1.9.1' sub-section row — count=$s116w_b expected=$((s116w_base + 1)) (#644)"
+fi
+
+# §116c (AC4, regression guard — green before and after): an UNMODIFIED copy
+# still balances against the independently derived expectation, so §116a+§116b
+# are not simply bounding an always-failing derivation.
+s116w_c=$(s116_posture_rows "$s116w_pristine")
+if [ ! -s "$s116w_pristine" ]; then
+  ng "116c: fixture not built — pristine SPEC copy missing or empty (#644)"
+elif [ "$s116w_c" -gt 0 ] && [ "$s116w_c" = "$s116_exp" ]; then
+  ok "116c: an unmodified SPEC copy still balances ($s116w_c = $s116_exp) (#644)"
+else
+  ng "116c: unmodified SPEC copy no longer balances — classified=$s116w_c expected=$s116_exp (#644)"
+fi
+
+# §116d (AC5, regression guard — green before and after): the real SPEC.md is
+# byte-identical to the copy taken before any fixture was built. Pins that the
+# fixture builder writes only to $TMP.
+if [ -s "$s116w_pristine" ] && cmp -s "$s116_spec" "$s116w_pristine"; then
+  ok "116d: the real SPEC.md is unmutated by the §116a-§116c fixtures (#644)"
+else
+  ng "116d: SPEC.md changed while the §116 window fixtures ran — a fixture wrote to the real SSOT (#644)"
 fi
 
 # ---------- §117: command docs use -F (not -f) for gh api stdin/file body (#452) ----------
--- a/scripts/test/smoke.d/50-perproject-recall.sh
+++ b/scripts/test/smoke.d/50-perproject-recall.sh
@@ -2631,21 +2631,37 @@ s116_exp=$((s116_levers + s116_agents + s116_cmds + s116_hooks))
 # SINGLE derivation of the count: the live parity arm below AND the §116a-§116c
 # window-terminator arms (#644) both call it, so the window rule lives in exactly
 # one place and cannot be fixed in one caller while drifting in the other.
+# THE SINGLE DEFINITION of "what ends §1.9's window" (#644). Used by BOTH the
+# live derivation below AND the §116a/§116b fixture builder, so the guard and the
+# arms that bound it cannot drift apart — which is the whole reason the
+# derivation was extracted into a function in the first place.
+#
+# DEPTH 2-3, not "depth <= 3". `^# ` is deliberately ABSENT: it matches any
+# shell comment, and SPEC carries 16 such lines inside fenced examples, so
+# including it would collapse the window to 0 the day §1.9 gains a fenced
+# example (measured). Dropping it costs nothing — the only thing it could ever
+# terminate on is a SECOND document H1 appearing after §1.9, in a file with
+# exactly one H1 (line 1) across 2900+ lines, and the `i&&` guard already keeps
+# line 1 from closing a window that has not opened.
+#
+# RESIDUAL, named rather than rounded off: this rule is line-oriented and knows
+# nothing about code fences, so 9 fence-interior `^## ` lines elsewhere in SPEC
+# remain collidable. Accepted for #644 — §1.9 carries no fence today and the
+# failure is loud (§116 reds at classified=0) rather than silent. Fence-state
+# tracking is deliberately NOT attempted here.
+S116_END_RE='^## |^### '
+
 s116_posture_rows() {
-  # Terminate at the NEXT HEADING OF DEPTH <= 3, not at the literal `## 2. ` (#644).
-  # The old terminator put every section between §1.9 and `## 2.` INSIDE the
-  # counting window, so a posture-token table row in a new `### 1.10` inflated
-  # the count and reddened §116 for a reason unrelated to the parity it checks
-  # (measured: 66 -> 67). `#### ` is deliberately NOT a terminator — position 3
-  # is `#`, not a space, so a genuine `#### 1.9.1` sub-section of §1.9 stays in
-  # scope and its rows still count (§116b pins that direction).
+  # `#### ` is NOT a terminator — position 3 is `#`, not a space — so a genuine
+  # `#### 1.9.1` sub-section of §1.9 stays in scope and its rows still count
+  # (§116b pins that direction).
   #
   # The `i&&` guard is LOAD-BEARING, not defensive: without it the terminator
-  # matches SPEC's own H1 (`# GHJig-Claude — Design Specification`) on LINE 1 and
+  # matches the §1.9 heading itself and every earlier `## `/`### ` heading, and
   # awk exits before the window ever opens, yielding 0. That would not fail
-  # silently — §116's `-gt 0` check and §116a-§116c's baseline guards all catch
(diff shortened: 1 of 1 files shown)


ilgyu-yi:
## Finding triage (round 1)

`reviewed-head: 6e8dbf11d577abb548de7412427cf7dac227efb8` · `prior-round: none (first round)` · `swing: not-evaluated` on every fix-now (a `swing: none` at round 1 is vacuous).

**Dispatch.** One worktree-isolated `code-reviewer` (verdict `ship after fix`), which resolved the head independently and matched the caller's privately-held head — **valid**. No `security-reviewer` and no §4.11 tier: single test file, no enforcement code, no auth/input/deps/crypto/IO boundary, no ADR. Judged by a fresh worktree-isolated `finding-judge` that re-measured every finding from the head blob (the ambient worktree sits at base, so it is not the artifact).

**5 findings, 5 confirmed, 0 refuted. fix-now ×3, drop ×2.**

### F1 — fix-now. My own new arm reproduces, one level up, the bug this PR fixes

`§116b` splices its `#### 1.9.1` decoy at the literal `## 2. ` anchor. That works **only** because §1.9 happens to be §1's last subsection today. Confirmed by simulation:

```
today's SPEC : baseline=66  §116b=67   (expects 67 — ok)
future SPEC  : baseline=66  §116b=66   (expects 67 — NG)
```

The heading map shows the cause: real `### 1.10` becomes the terminator, and the spliced `#### 1.9.1` lands **behind** the window it is meant to be inside. Issue #644 itself names three Active Directives planning a §1.x section, so "not reachable today" is a weak defence — and the PR's own thesis is that a guard must not red for a cause outside its subject.

**Remedy** (verified: today `incl=67/excl=66`, future `incl=67/excl=66`): anchor the splice at the **end of §1.9** — the first depth-≤3 heading *after* `### 1.9 `, computed — instead of the document-global literal.
**Floor**, all measured, all must still hold: `§116a` stays a **live witness** (re-anchored exclusion fixture counts 67 against the *pre-fix* terminator vs baseline 66, on both SPECs); the inclusion fixture yields baseline+1 on **both** today's and a future SPEC; and the two fixtures still come from **one builder at one anchor**, differing **only** in heading depth.
**Ceiling**: re-anchoring that shared builder is the whole change — no second builder, no change to the terminator rule (that is F2), no arms added or removed, no change to `$s116_exp`. The anchor must **reuse the single window rule** rather than re-spell the heading regex, or the extraction's one-derivation property dies on arrival.

### F2 — fix-now. Drop the `^# ` alternative

A fenced shell comment inside §1.9 collapses the window:

```
fenced `# a shell comment`  ->  new(head)=0, old(pre-fix)=66   <- the collapse is NEW
fenced `## 2. `             ->  new=0,       old=0             <- the CLASS pre-exists
```

Surface count over the whole SPEC, fence-interior lines: `^# `=**16**, `^## `=9, `^### `=0, `^## 2. `=0. So the old terminator had **zero** collidable lines in this document and the new one has 25. §1.9 carries no fence today, so the hazard is prospective.

**Remedy**: drop `^# ` — removes 16 of the 25 collision lines (64%) with **zero** behavioural change (measured: SPEC=66, exclusion=66, inclusion=67, future-inclusion=67, fenced-comment SPEC back to 66). It does **not** reintroduce the line-1 H1 trap: `i&&` is load-bearing for the `^## ` arm regardless and stays.
**Floor**: `^## ` **and** `^### ` must both remain — narrowing further reinstates the over-extension this PR exists to fix.
**Ceiling**: no fence-state tracking. Nine fence-interior `^## ` lines of residual remain and are **accepted** for this PR (the residual failure is loud, and §1.9 has no fence today). **Also inside the ceiling**: the function comment — and #644's AC1 — say *"depth ≤ 3"*, which after the drop is depth **2–3**. The comment must say what the code does, or the next reader re-adds `^# `.

The judge rejected the reviewer's weaker "at minimum document the caveat" on **likelihood, not direction**: SPEC already carries 16 fenced shell-comment lines elsewhere, while the only thing `^# ` would ever catch is a *second* document H1 appearing after §1.9 — a structural anomaly in a file with exactly one H1 across 2,900+ lines.

### F3, F4 — drop (confirmed true, no change)

**F3** — `§116c` is information-equivalent to `§116` (a `cp` of the same file against the same `$s116_exp`; it can only fail if `cp` corrupts, which `§116d`'s `cmp` catches independently). Confirmed — but **#644's AC4 asks in so many words** for "a third arm [that] asserts the healthy SPEC still balances". Deleting it fails the AC's letter at closeout; strengthening it is an arm-set redesign, outside this PR's scope and #644's own Out-of-scope. Not deferred either: "an AC asked for a tautology" is not an open defect, and filing it would seed a later round with a phantom concern.

**F4** — `s116w_base` recomputes a count already in `$s116_rows`. True, behaviour-free. In a PR that must already re-anchor the shared builder and narrow the terminator, a preference-level edit to a correct line adds churn a re-review must re-read, for no verdict change.

### F5 — fix-now. Disclose the fourth deviation

Timeline holds exactly `{committed ×2, labeled ×1}` — no `ready_for_review`, no `convert_to_draft`; both commits (16:29:17Z, 16:33:07Z) predate `createdAt` 16:33:41Z. The PR was opened ready with the work complete, skipping draft-first and the commit→body curation order. The history cannot be replayed, but the **disclosure gap is live**: a Plan that names three deviations and omits a fourth reads as complete, which calibrates the reader's confidence to a list that isn't.

**Floor**: one line in the existing Plan naming the deviation and its ground. **Ceiling**: that line and nothing more — no commit amendment, no force-push, no re-opening as draft to replay a cadence that cannot be replayed, no SPEC or CLAUDE.md process change, nothing outside the Plan section.

### Coupling the author must respect

F1 and F2 touch the same rule from two sides. If F1's anchor is computed with the heading regex, it and the F2-narrowed terminator **must come from one place** — two copies of "what ends §1.9" is precisely the drift the `s116_posture_rows` extraction was introduced to prevent.

ilgyu-yi:
## Finding triage (round 3)

`reviewed-head: 18c3cbc20816f7aae4bf1d35d9ecd25af783b758` · `prior-round: 1@6e8dbf11d577abb548de7412427cf7dac227efb8`

### First: a §4.13 ordering failure of mine, recorded rather than quietly fixed

**Round 2's judged list was never posted.** This PR carries exactly one prior comment — round 1's. Confirmed: `posted rounds: ["1"]`. Round 1 satisfied durable-before-fix; round 2 did not, and I applied round 2's fixes anyway.

The cost is concrete, not bookkeeping: round 2's floors and ceilings — the ones bounding the very repair round 3 is now correcting — exist **nowhere a later reader can audit**, and this round's swing detection had to reach back **two** rounds instead of one. The marker count therefore under-reports by one, which is why this list is `round=3` rather than the marker-derived `2`; emitting `2` would collide with the real round 2 and corrupt every later derivation.

### F1 — fix-now. The PR title is the last site asserting the stale rule

`confirmed` by a depth truth table over SPEC at head — baseline 66; a depth-1 `# ` spliced inside §1.9 → **66 (does NOT terminate)**; depth-2 `## ` → **0 (terminates)**; depth-4 `#### ` → **66**. The rule is exactly {2,3}, so *"depth ≤ 3"* would include depth 1, which measurably does not close the window.

Round 2's F-C landed **completely in code** — `:2639`, `:2676`, `:2737` all read "depth 2-3". The title is the only false site left, and the PR body's own F2 section states the opposite of it.

**The merge-permanence claim was verified against this repo's own history**, not assumed:

```
$ git log -1 --format='%s%n%b' a8fed28
Merge pull request #649 from ilgyu-yi/ilgyu-yi/feat/645-finding-judge
feat(#645): route review findings through a finding-judge layer
```

The merge commit body **is** the PR title verbatim under the `--merge` strategy the hook mandates. Merging as-is writes a measurably false claim permanently into `main`.

**Floor**: the title — the one site both false *and* immutable at merge. **Ceiling**: the title **plus** the *Key context* lead-in ("the depth-3 rule"), and nothing else. The lead-in is *inside* the ceiling on an independent ground: once the title says `depth-2-3`, a body still saying "depth-3 rule" is a mismatch **manufactured by the fix itself** — reproducing exactly the "one stale token left behind" mechanism that produced round 3 from round 2. Explicitly **outside**: no code edit, no commit, no amend, no force-push, no consolidating the duplicated comment blocks, no touching the hardcoded `2,893`, and **no edit to #644's AC1** — its "depth ≤ 3" is the historical *ask*, and rewriting it would falsify the record.

`swing: none` — rounds 1→2→3 move monotonically along one axis (round 1's F2 ceiling already said "the comment must say what the code does"; round 2 applied it to the code, round 3 to the title). No reversal.

### OBS2 — defer-to-issue, and **filed before merge**

`confirmed` and measured by me independently: `s116_levers` terminates its §1.8 window at the literal `/^### 1\.9 /`.

```
today                                        levers = 6
with a `### 1.8.5` carrying one `| **` row   levers = 7   <- inflated
with §1.9 renumbered (terminator never fires) levers = 7
```

Both inflate `s116_exp` and red `§116` for a cause outside its subject — structurally identical to what this PR fixes, one line above the diff.

The judge split the reviewer's ruling: *"out of scope"* and *"file it"* are **not alternatives**, they are the two halves of defer-to-issue. Not fixed here (round 3's discipline is zero-code, and a new terminator change is exactly the non-required finding that manufactures the next round's defect). But **filed before merge**, because this repo has had a tracking claim without an artifact **twice** — #664 exists for that reason, and #644's own body records the same failure on PR #641. **Binding**: nothing may say "tracked separately" until the issue number exists.

### Dropped (4)

- **OBS1** — `s116w_fixture` reads `$s116_spec` from enclosing scope, so the future-SPEC case is provable only out-of-band. A *missing proof*, not a wrong answer: the shared-anchor design already makes the arms correct in that future and no AC asks for it. The judge declined even the reviewer's "follow-up issue candidate" — filing it would seed a later round with a phantom concern.
- **OBS3** — the code diff was net-zero delete/narrow-only but the body gained a parenthetical. Not a defect: the PR body is **editorial by contract**, and the discipline was for *code*, where it held exactly.
- **IDIOM1 / IDIOM2** — comment placement, and the hardcoded `2,893`. Both dropped for one shared reason: each remedy is a **prose edit to the code file**, which is the measured generator of rounds 2 and 3, and neither can red a test.

**IDIOM2 is worth keeping as a note.** Round 2 changed that figure from `2900+` — drift-**proof** — to `2,893` — drift-**perishable** — in the name of correcting an unmeasured figure. That is the recurrence mechanism in miniature: a repair locally correct and globally a small regression.

### Convergence: convergent, conditionally

| | round 1 | round 2 | round 3 |
|---|---|---|---|
| defect class | behavioural code defects | false claims in code comments | one false claim in mutable PR metadata |
| code lines touched by the fix | substantive | 5 (net zero) | **0** |
| requires commit / push | yes | yes | **no** |

Severity and surface fall monotonically, and the generator is **identified and removed**: rounds 2 and 3 were produced by prose edits that added or restated claims in the code file. Round 3's fix reaches no file at all, so it cannot manufacture a round-4 code defect.

The caveat is binding: this converges **only if the ceiling is enforced**. The four drops are what make round 3 terminal. Tidying the comment blocks or the `2,893` "while I'm here" manufactures round 4 on schedule.

ilgyu-yi (commented):
## Review — approve

`code-reviewer` returned **`ship`** at this head, on a round-4 confirmation pass. The head was computed privately by the caller and never passed to the reviewer; the reviewer resolved it independently and the two matched, so the vote is valid under SPEC §4.5.

### Verdict path

Four rounds, each at a head the reviewer resolved for itself:

| round | verdict | what it found |
|---|---|---|
| 1 | `ship after fix` | the PR's own new smoke arm reproduced this PR's bug one level up |
| 2 | `ship after fix` | round 1's repair had written a fresh false clause into the comment it repaired |
| 3 | `ship after fix` | the PR title was the last site asserting the stale rule |
| 4 | **`ship`** | bounded remedy landed; nothing else moved |

Round 4 confirmed the round-3 remedy stayed inside its ceiling by inspecting the PR's own edit history rather than trusting the diff: exactly one title rename and a one-line body edit, no `head_ref_force_pushed` event, commits unchanged, and issue #644's body never edited (so AC1's original wording survives as the historical ask).

### What the change does

`§116`'s parity guard derived SPEC §1.9's posture-row count with a window terminating at the literal `## 2. `, so any section authored between §1.9 and there sat **inside** the counting window — a posture-token row in a new `### 1.10` inflated the count 66 → 67 and reddened `§116` for a cause outside the parity it checks. The window now closes at the next window-closing heading, scoped to the window with an `i&&` guard.

Measured, both directions, on today's SPEC **and** on a simulated SPEC carrying a genuine `### 1.10`: `base=66`, `excl=66` (a `### 1.10` decoy excluded), `incl=67` (a `#### 1.9.1` sub-section still counted). `§116a` remains a live witness in both — it reds at `excl=67` against the pre-fix terminator.

### Verification at this head

- Full suite `pass=1284 fail=0`; CI green on `smoke` (ubuntu + macos), `bash -n` ×2, `mutation`, `fragment-gate`; `mergeStateStatus: CLEAN`.
- The real `SPEC.md` is unmodified after a full run — sha256 identical before and after (AC5).
- The implemented rule is exactly depth {2,3}, measured by splicing each depth inside §1.9: depth-1 → 66 (does not terminate), depth-2 → 0, depth-3 → 0, depth-4 → 66.
- One shared `S116_END_RE` is consumed by both the live derivation and the fixture builder, so the guard and the arms bounding it cannot drift.

### Disclosed divergences, both deliberate

- **AC1's letter says "depth ≤ 3"; the shipped rule is depth 2–3.** `^# ` is absent because it matches any shell comment and SPEC carries 16 such lines inside fenced examples — including it collapses the window to 0 the day §1.9 gains a fenced example (measured: 0 with it, 66 without). AC1's operative content is fully discharged; only its shorthand differs. AC1's text was deliberately not edited, and the divergence is recorded in the code, the PR body, the judged list, and the AC closeout.
- **An adjacent defect is tracked as #668, filed before any PR text claimed it was tracked.** `s116_levers` has the same shape one line above this diff. Timestamps confirm the ordering: #668 was created before the body revision that first mentions it.

### Judged findings not fixed here

Four findings were ruled `drop` across the rounds — `§116c`'s redundancy (AC4 asks for it in so many words), `s116w_base`'s recompute, the comment-block placement, and a hardcoded line count. Each remedy is a prose edit to the code file, which is this PR's measured generator of repeat defects. Round 4 flagged two further wording nits in the PR body and closeout **for the record only, with an explicit instruction not to edit**; both are noted rather than acted on, on the same ground.

The marker below is the machine-readable signal for the merge-review gate; its `verdict` derives from the round-4 `ship` and its `head` is the head computed privately before the review, not re-derived at write time.

PR #669, opened 2026-08-16
fix(#668): anchor the §116 lever window on the section title and the shared terminator

### LINKED ISSUE #668: smoke §116's lever window uses literal headings at both ends, so a new §1.8.x inflates the expected count
## What

`s116_levers` in `scripts/test/smoke.d/50-perproject-recall.sh` derives the §116 parity guard's **expected** count by walking SPEC §1.8's lever table, and both ends of that window are **literal headings**:

```sh
s116_levers=$(awk '/^### 1\.8 /{i=1;next} /^### 1\.9 /{exit} i&&/^\| \*\*/{n++} END{print n+0}' "$s116_spec")
```

So everything authored between §1.8 and `### 1.9 ` sits **inside** the lever-counting window. `s116_levers` feeds `s116_exp` (`levers + agents + cmds + hooks`), which §116 compares against the derived posture-row count.

**The defect, measured** at SPEC as it stands (`levers=6 agents=12 cmds=28 hooks=20 → exp=66`, `rows=66`, green):

| fixture | levers | rows | §116 |
|---|---|---|---|
| SPEC as it stands | 6 | 66 | green |
| a `### 1.8.5` carrying one `\| **`-shaped row | **7** | 66 | **red — `exp=67` vs `rows=66`** |

That red is the defect: an unrelated table row in a new §1.8 sibling makes the parity guard fail for a cause that has nothing to do with the parity it checks.

**Two adjacent cases that are NOT that defect**, measured and stated so the scope is not overclaimed:

- **§1.9 renumbered** → `levers=7`, but `rows=0`, so §116 reds via its `-gt 0` anti-vacuity guard **by design**. The lever inflation is real but **masked** — it is not the cause of the red. This is a *terminator-robustness* concern, not a wrong-cause failure.
- **§1.8 renumbered** → `levers=0`, i.e. `exp=60` vs `rows=66`. The **start** anchor is equally literal. This is a distinct failure of the same window and is in scope here because it is the same two-line extractor.

## Why

§116 is the parity guard that keeps §1.9's harness-overlap classification complete: a mechanism enumerated in §1.8, §4, §5 or §6.1 that gains no posture row fails smoke. A window that admits unrelated content makes that guard fail for the wrong reason, which trains the operator to read a red §116 as noise — the failure mode the guard exists to prevent, inverted.

**MISSION fit:** *The mechanism* — the cost-asymmetry rule makes a wrong nudge cheap only when it is ignorable. A parity guard that reds for a cause outside its subject is not cheap.

**Why now.** This is the **second** window in the same smoke arm found to have this shape. It was surfaced by the round-3 review of PR #667 (open, `Closes #644`), which fixes the *sibling* window in the same `if` block and deliberately does not touch this one.

Filing before that PR merges is the ordering, not the reason: the reason is the measured red above. The ordering matters because this repo has twice had an artifact assert a tracking item that did not exist — #664 exists for that reason, and #644's own body records the same failure on PR #641. Nothing in PR #667 claims this is tracked, and nothing will until this Issue has a number.

*(#644 argued near-term likelihood from three Active Directives planning new SPEC sections. That argument does **not** transfer: those point at a `### 1.10`, which is #644's window, not a `### 1.8.5`. The case here rests on the measured red and on §1.8 being the lever table that any new enforcement lever must extend — not on a borrowed likelihood claim.)*

**Existing coverage:** none. PR #667 fixes the **row-count** window in the same arm — a different extractor with a different terminator — and its round-3 judged list records the split. #643 covers the **§156** window in a different file. No Issue covers `s116_levers`.

## Acceptance criteria

- [ ] **Neither end of `s116_levers`' window is a literal heading.** The terminator closes at the next window-closing heading; the start anchor no longer depends on §1.8's literal spelling. *(If PR #667 has merged, the terminator rule already exists there as a single shared definition and must be **reused, not re-spelled** — two copies of "what ends a §1.x window" is the drift that shared definition exists to prevent. If it has not merged, this Issue introduces the shared definition and #667 is unaffected.)*
- [ ] **Exclusion arm, the witness**: a temp SPEC copy carrying a decoy `\| **`-shaped row inside a `### 1.8.5` must not change the derived lever count. It **must red against pre-fix code** — measured today it yields 7 against a baseline of 6.
- [ ] **Inclusion arm**: a `\| **` row inside a genuine `#### 1.8.1` sub-section of §1.8 **is** still counted, so the fix does not shrink the guard's real scope.
- [ ] **Start-anchor arm**: a SPEC copy with §1.8 renumbered must still yield the correct lever count. Measured today it yields **0** (against 6), so this arm also reds against pre-fix code.
- [ ] The real `SPEC.md` is never mutated by any arm.
- [ ] **Each new arm is labelled in-file as witness or bound.** Measured above, the exclusion and start-anchor arms red pre-fix and are witnesses; an inclusion arm that passes both before and after is a bound. A guard dressed up as a witness manufactures confidence the arm does not have.

## Out of scope

- **The row-count window** (`s116_posture_rows`) — that is #644 / PR #667, and this Issue must not re-open it.
- **§1.9-renumbering robustness for the *row* window.** Measured above, that case reds §116 by design through the `-gt 0` guard, so it is not a wrong-cause failure and needs no repair here.
- **Upgrading §116 from count parity to name-level identity parity.** Today a lever row and a posture row for two *different* mechanisms cancel out and the guard still balances — a real weakness, not created here, and a larger change. #644 scoped it out on the same ground.
- **The other `s116_exp` inputs** (`s116_agents`, `s116_cmds`, `s116_hooks`). They derive from file counts and `grep -c` over anchored patterns, not from a windowed extraction, so they do not share this shape.

## Notes

Surfaced by the `code-reviewer` on PR #667 as an explicit out-of-scope observation, then ruled `defer-to-issue` — with filing required **before** that PR merges — by the round-3 `finding-judge`, on the ground that "out of scope" and "file it" are two halves of one disposition rather than alternatives.

An earlier draft of this body claimed **both** the `### 1.8.5` case and the §1.9-renumbering case make §116 "red for a cause outside its subject". Measurement showed that is false for the second: `rows=0` there, so the anti-vacuity guard fires by design and the lever inflation is masked. The table above states the measured consequence per case rather than in aggregate, and the §1.8-renumbering case was added because the same measurement pass found the start anchor is literal too.

ilgyu-yi:
Closes #668

## Goal

Stop `§116`'s **lever** window from breaking in both directions. It opened on the literal `### 1.8 ` and closed on the literal `### 1.9 `, so a `### 1.8.5` sibling sat **inside** the window (levers 6 → 7) and renumbering §1.8 meant the window never opened at all (→ 0).

MISSION *"The mechanism"* — the cost-asymmetry rule makes a wrong nudge cheap only when it is ignorable. Both defects feed `s116_exp`, which `§116` **and** `§116c` consume, so either one reds **two** arms for a cause outside the parity being checked — and `§116c`'s message wrongly blames the SPEC copy. A guard that reds for the wrong reason trains the operator to read it as noise.

## Plan

Type is `fix`, so §1.2's reproduce-first relaxation applies — all four directions were measured on `main` before anything was written.

**Docs: n/a — no SSOT contract changes.** The window is a smoke-arm implementation detail with no SPEC clause; the arm's own comments are its documentation. Test-only surface, so `skip-changelog` per §18.7 (`test`-only is explicitly skip-eligible).

The planner gate did not fire: two files at head (round-1's F3 added the second), against a 3+ threshold. **This PR was opened ready rather than draft-first** — with `Docs: n/a` there is no Phase-A commit to seed a draft with; disclosed here rather than left unstated.

## Key context

| fixture | before | after | wanted |
|---|---|---|---|
| SPEC as it stands | 6 | 6 | unchanged |
| a `\| **` row in a new `### 1.8.5` | **7** | **6** | EXCLUDED |
| §1.8 renumbered | **0** | **6** | window still opens |
| a `\| **` row in a genuine `#### 1.8.1` | 7 | 7 | still COUNTED |

The start anchors on the section **title** (matches exactly one line, survives renumbering); the end **reuses `S116_END_RE`**, the same definition §1.9's window uses. `#### ` is not a terminator — position 4 is `#`, not a space.

## Checklist

- [x] **Test** (`fe77a7f`): `§116e` exclusion + `§116f` start-anchor (both witnesses, both red), `§116g` inclusion (a bound). Red before Code.
- [x] **Code** (`820e18b`): title-anchored start, shared terminator, and the hoist below.
- [x] **Review fixes** (`e01e8ae`, `9c11226`): round-1 judged list — F1, F2, F3 (citing #670), F5, I1, I2; round-2 — N1, N3.

## Risks

- **The hoist is load-bearing, not cosmetic.** `S116_END_RE` was defined ~30 lines *below* `s116_lever_rows`. Passing it as `-v endre=` from there expands an **unset** variable, the subshell dies under `set -u`, and `s116_levers` becomes **empty** — reddening `§116` *and* `§116c`. **Corrected in round 1:** the symptom is *not* an arithmetic error. `set -uo pipefail` carries no `-e`, so the assignment completes and an empty-but-set variable evaluates as 0 in `$(( ))`; both arms emit an **ordinary parity message** carrying `expected=60` (§116's also shows an empty `levers=` field), with only one `unbound variable` line on stderr. That is **worse** than the original claim — a parity-shaped red is indistinguishable from genuine §1.9 drift — so the constraint stands, more strongly. *(The `820e18b` commit body carries the original, superseded wording; commit bodies are immutable after push, so it is corrected here rather than by rewriting reviewed history.)*
- **One shared definition, still exactly one.** Verified: `S116_END_RE` is defined once and consumed by both live derivations and the fixture builder. Two copies of "what ends a §1.x window" is the drift AC1 exists to prevent.

## Round-1 review fixes (`e01e8ae`)

The judged list is posted above: **6 fix-now, 1 defer, 3 drop.** Two rulings are worth surfacing.

**F1's remedy was ruled *below the floor* as a token substitution.** Swapping "arithmetic error" → "parity message" would have produced *"a parity message rather than a parity message"* — the whole contrast was false, not the two words. See the corrected Risks bullet.

**F3 carried two coupled legs**, and the coupling was deliberate: the "verbatim" claim could not be corrected until the Issue it cites existed. **#670 was filed and numbered first.** The claim is false in *both* halves — the posture window had already diverged at #644/#667 — and nothing enforces it: that file makes zero references to §116's helpers, so no arm compares them and the claim decayed silently across two unrelated PRs. Only comment lines changed there; the awk expressions, the assertion and the `ng` text are byte-identical, and the true "Nothing is hardcoded" sentence is preserved.

**One over-application, caught and reverted.** F5's ceiling was the four `awk -v` sites; I initially rewrote six, including two adjacent `grep` calls. `grep` does **not** strip backslashes the way `awk -v` does, so there the change would have made the pattern match nothing. Caught by checking the count against the stated ceiling, and reverted.

**Deferred / dropped, not acted on:** F4 (a collapsed baseline emits a wrong-cause `ng` — confirmed *inherited* from #644, so fixing it here means touching six arms or diverging three); `s116l_base`'s recompute (it is the **arms' denominator**, deliberately distinct from the live parity input — collapsing them would couple the bounding arms to the derivation they bound); the pre-existing "position 3" explanation on an untouched line; and the missing `Docs: n/a` line in the pushed commit bodies, whose remedy is unexecutable.

## Round-2 review (`9c11226`)

`code-reviewer` returned **`ship`** — no blocker-stop — and raised four advisories. The judge ruled **2 fix-now, 1 defer, 1 drop**, and sharpened two of them.

**N1 — my over-application last round was worse than "matches nothing".** The two `grep` sites are not symmetric: at `s116l_h_orig` an empty result is caught loud, but §116f's leftover-heading guard fires on `!= 0`, and `grep -c` with the `awk` form returns **0 even on a file that has the heading** — so the leg would never fire and the **anti-vacuity guard would silently stop guarding**. The escape split is now stated at the builder, **addition only**: no existing line was edited, which is what separates this from every defect this PR has produced (the extraction, the stale recast and the over-application all came from *changing* existing content).

**N3 — a false detail in the block round 1 already corrected.** The `levers=` field exists in §116's `ng`, not §116c's. Applied as a **scoping deletion** rather than a recast, because removing a false clause cannot introduce a new one. This is the **third consecutive round** to find a defect in that one comment block; it is now the presumed generator, and the next finding in it faces a strong presumption of `drop`.

**Dropped, with the finding's framing corrected**: the `~30 lines below` clause. The reviewer implied this round degraded it to 23; measured, it was **19** when written and I1's move carried it *toward* `~30` — it is the most accurate it has ever been. **Deferred off the critical path**: the `Docs: n/a` norm, omitted in 3/3 commits here, whose remedy is unexecutable on pushed bodies but whose enforcement gap is real.

## Test plan

`§116e` and `§116f` are the **witnesses** — red at `1285/2` before the fix, green after. `§116g` passes **both** before and after and is labelled in-file as a **bound**, not a witness: `#### ` is depth 4 and the shared terminator stops at depth 3, so the change never touches it. It carries one load-bearing job for `§116e` — it proves the identical decoy row **does** match the counting regex, so a green `§116e` cannot be a decoy that never matched at all.

The fixture builder was **generalized, not duplicated** — it now takes an open-regex, so §1.8's and §1.9's fixtures share one builder. Both §1.8 fixtures splice at the first window-closing line *after* §1.8 opens, never at a document-global literal: that is #667's round-1 lesson applied one section up, where a fixture anchored at `## 2. ` would have reddened for a cause outside its subject the moment a sibling section landed. `§116f` is a rename rather than a splice and derives the renamed heading from the live line, so no SPEC title is hardcoded in the smoke file.

**Measured, in a window with no live agent worktree:** `pass=1285 fail=2` → **`pass=1287 fail=0`**; `lint OK: 79 files clean`; real `SPEC.md` unmodified after a full run (AC5, pinned by the existing `§116d`, whose claim now honestly reads `§116a-§116g`).

### DIFF
--- a/scripts/test/smoke.d/50-perproject-recall.sh
+++ b/scripts/test/smoke.d/50-perproject-recall.sh
@@ -2622,7 +2622,22 @@ fi
 # rename / empty table (anti-vacuity #2). A mechanism added to any of the four
 # families bumps the expected total and trips this guard until a §1.9 row is added.
 s116_spec="$SHELL_ROOT/SPEC.md"
-s116_levers=$(awk '/^### 1\.8 /{i=1;next} /^### 1\.9 /{exit} i&&/^\| \*\*/{n++} END{print n+0}' "$s116_spec")
+# s116_lever_rows <spec-file> → count of §1.8 narrowing-lever table rows in that
+# file. THE SINGLE DERIVATION of the lever count: the live `s116_levers` below
+# AND the §116e-§116g lever-window arms (#668) both call it, so the window rule
+# lives in exactly one place and cannot be repaired in one caller while drifting
+# in the other. Same shape, and for the same reason, as s116_posture_rows below.
+#
+# THIS IS NOT THE FIX (#668, Phase B). The awk body is the PRE-FIX derivation
+# verbatim — both ends of the window are still literal headings — so §116e and
+# §116f red against it exactly as #668 measured. The extraction is
+# behaviour-preserving (`s116_levers` still derives 6); without it the Code
+# phase would repair the inline copy while the new arms went on reading a
+# second one, leaving both witnesses unfixable.
+s116_lever_rows() {
+  awk '/^### 1\.8 /{i=1;next} /^### 1\.9 /{exit} i&&/^\| \*\*/{n++} END{print n+0}' "$1"
+}
+s116_levers=$(s116_lever_rows "$s116_spec")
 s116_agents=$(ls "$SHELL_ROOT"/.claude/agents/*.md 2>/dev/null | wc -l | tr -d ' ')
 s116_cmds=$(grep -cE '^### 5\.[0-9]+ `/' "$s116_spec")
 s116_hooks=$(grep -oE 'should_skip [a-z-]+' "$SHELL_ROOT"/.claude/hooks/pre_tool_use.sh | awk '{print $2}' | sort -u | wc -l | tr -d ' ')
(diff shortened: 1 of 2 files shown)


ilgyu-yi:
## Finding triage (round 1)

`reviewed-head: 820e18b9c46749dda0aabcafe83f9b27840914b8` · `prior-round: none (first round)` · `swing: not-evaluated` on every fix-now.

**Dispatch.** One worktree-isolated `code-reviewer` (`ship after fix`), head resolved independently and matched — **valid**. No security-reviewer, no §4.11 tier: single test file, no enforcement code, no security surface, no ADR. Judged by a fresh worktree-isolated `finding-judge` reading every artifact via `git show <head>:` (the ambient worktree is at base, so it is not the artifact).

**5 findings + 5 idiom/process items. All 10 confirmed, 0 refuted. fix-now ×6, defer ×1, drop ×3.**

### F1 — fix-now. A token substitution is **below** the floor

My hoist comment claims the pre-hoist ordering reds §116/§116c *"with an **arithmetic error** rather than a parity message"*. Reproduced under the artifact's real mode (`set -uo pipefail`, no `-e`, section sourced by `smoke.sh`):

```
stderr: S116_END_RE: unbound variable
levers=[]   →   s116_exp=$((levers+…))  rc=0  exp=60      ← NO arithmetic error
arm emits:  116: … classified=66 expected=60 (levers= agents=12 …)   ← a PARITY message
```

The false part is the **whole contrast**, not the words "arithmetic error": swapping in "parity message" yields *"a parity message rather than a parity message"*, and "unbound-variable error" is still false because the arms themselves emit parity `ng`s. The measured symptom is **worse** than the claim it replaces — a parity-shaped red is indistinguishable from genuine drift except for one stderr line — so the block's conclusion ("order is load-bearing, not cosmetic") survives and must not be weakened.

**Ceiling**: the clause plus the PR body's Risks bullet. Explicitly **out**: the `820e18b` **commit body**, which repeats the claim and is **unreachable** — amend-after-push is blocked by the force-push matcher, and rewriting reviewed, pushed history to fix a sentence is exactly the move this repo forbids. The honest disposition is one PR-body clause noting the commit body's superseded description.

### F2 — fix-now. Both stale sites, not one

`:2723-2726` still says the window *"opens at the literal `### 1.8 `… both ends must go"* — present/future tense for code that no longer exists — and `:2743-2745` says the witnesses *"each red as it stands"*, measured **green** at head. **Floor**: both, or the block self-contradicts. **Ceiling**: #668-authored prose only. #644's block immediately above is written **counterfactually**, verified accurate at head, and is the **model, not a target**.

### F3 — fix-now, with two **coupled** legs

`70-gates-contentlocks.sh:2446` claims its expressions are *"§116's verbatim"*. Measured, that is false in **both** halves — at the merge base the posture expression had **already** diverged (#644/#667), and this PR completes the falsification. And §156j's copy carries the defect #668 removes: renumber §1.8 and it goes `1 → 0`, emitting *"lever rows=0 … mis-shaped or absent"* when the rows are present and correctly shaped and only the window failed to open.

**Floor, both legs mandatory and coupled**: (1) the "verbatim" claim is corrected and says it is false in **both** halves; (2) a tracking Issue for the §156j window repair is **filed, numbered, before merge**, and the corrected comment **cites that number**. The coupling is deliberate — this repo has had a tracking claim without an artifact three times, and a comment that cannot be written until the Issue exists is self-verifying.

**Ceiling**: no change to the awk expressions, no new §156 arms, no change to §156j's assertion or `ng` text. Repairing that window here would be an untested cross-file change in a PR whose Test commit is already landed — inverting Doc→Test→Code.

### F5 — fix-now. A newly introduced 4-character regression

`awk -v` performs escape processing, so `-v openre='^### 1\.9 '` arrives as `^### 1.9 ` and the dot is a **wildcard** — a `### 1x9` heading opens the window. Pre-PR these were inline `/^### 1\.9 /` where `\.` was literal; the extraction to `-v` loosened them. Behaviour-preserving on today's SPEC (each loose regex still matches exactly the one intended line), but it sits directly under a comment block whose subject is anchor rigor.

**Floor**: all **four** literals — the builder is now *shared*, so half-escaping leaves the two windows disagreeing on their own anchor rule. **Ceiling**: those four string literals; no new arm (the correction is behaviour-preserving, so an arm would be a bound, not a witness, and would reopen a committed Test phase).

### I1, I2 — fix-now, mechanically bounded

**I1**: the hoist orphaned `s116_lever_rows`' docstring 46 lines above its function, while the *same commit* correctly moved `s116_posture_rows`' docstring down to sit adjacent. **Ceiling: a verbatim move — change no words.** That is what takes it out of the "tidy while you're here" class this repo has measured as its repeat-defect generator: a verbatim move cannot introduce a prose defect. If it cannot be done without rewording, it drops.

**I2**: delete the bare contentless `#` at `:2658` left by the hoist — **that line only**. The new blank at `:2684` is explicitly excluded as pure style.

### F4 — defer-to-issue

A collapsed baseline routes the arms to an `else` naming the terminator, i.e. the wrong cause. Confirmed — **and confirmed inherited**: §116a/b/c carry the byte-identical shape from #644. Fixing it here means either touching **six** arms (expanding into #644's landed code inside a #668 PR) or diverging three, which manufactures the two-shapes-for-one-rule drift AC1 exists to prevent. The failure is loud either way; the cost is a misleading message, not a false green. **May be filed after merge** — nothing in the artifact asserts it is tracked. If any comment mentions it, F3's coupling rule applies.

### Dropped, with grounds recorded (3)

- **I3** — `s116l_base` recomputes `s116_levers`. Not accidental duplication: it is the **arms' denominator**, semantically distinct from the **live parity input**. Collapsing them would couple the bounding arms to the derivation they bound, so a defect would move baseline and measurement together and the arms would stop bounding anything. Worse than one extra awk pass.
- **I4** — the pre-existing *"position 3 is `#`"* at `:2686` is a wrong **explanation** (position 3 is `#` for `### ` too, so it does not discriminate). Pre-existing, on a line this PR did not touch, in the file whose prose edits are this repo's measured defect generator. Its correct twin sits 17 lines above. **This judged list is the durable record** — the measurement is not lost.
- **I5** — neither commit body carries `Docs: n/a — <reason>` as CLAUDE.md requires; only the PR body does. Dropped because the remedy is **unexecutable**: both commits are pushed and amend-after-push is blocked. Recorded so the omission is not silent.

### Ordering constraint

**F3's Issue is filed and numbered before the F3 comment edit is written**, because the edit cites the number. Every other item is independent.

### Observations the judge raised but did not add to the fix list

1. **`:2735` carries F2's defect on a line this PR edited** — *"only §116a reds against the current row-window terminator"* is present-tense and false at head. Kept **out** of F2's ceiling deliberately, because adding it is the scope creep this repo has measured as its generator. Flagged so the exclusion is a decision, not an oversight.
2. **Two confirmed findings are unfixable because they live in pushed commit bodies** (F1's false symptom, I5's missing line). That is a signal about *ordering* — the check belongs before the commit, not after.
3. **The "verbatim" claim decays by design.** It asserts byte-identity with another file's code and nothing enforces it; it went half-false at #644/#667 and fully false here, in two unrelated PRs, with no arm noticing either time. F3's follow-up Issue may want to scope that claim's *enforceability*, not just the §156j window.

ilgyu-yi:
## Finding triage (round 2)

`reviewed-head: e01e8aedee9f9d654f51e142fa5f420c7a3c94ba` · `prior-round: 1@820e18b9c46749dda0aabcafe83f9b27840914b8` · `swing: none` on both fix-now items.

`code-reviewer` returned **`ship`** — no blocker-stop — after verifying all 6 round-1 fix-now items true and inside their ceilings, the 4 defer/drop items untouched, `pass=1287 fail=0` on a fresh clone, §156j green, `SPEC.md` hash unchanged, and all 6 of #668's ACs discharged by measurement. It then raised four **advisory** findings. That verdict is unaltered.

**4 confirmed, 0 refuted. fix-now ×2, defer ×1, drop ×1.**

### N1 — fix-now. My over-application was worse than "matches nothing"

The two `grep` sites I wrongly double-escaped are **not symmetric in blast radius**:

- `:2812` — `s116l_h_orig=$(grep -m1 …)` yields empty, which `[ -z … ]` catches **loud**.
- `:2886` — the guard leg is `[ "$(grep -c '^### 1\.8 ' "$s116l_renum")" != 0 ]`. Under `\\.`, grep returns **0 on any file**, so `0 != 0` is false, the leg never fires, and **§116f's anti-vacuity guard silently stops detecting a rename that did nothing.** Measured: single-escape → 1 (loud `ng`), double-escape → 0 (silent pass).

A guard that stops guarding on an absent target is the exact anti-pattern `smoke.sh`'s own header names. The trap fired once this round and was caught only by ceiling-checking a count — a mechanism the next editor will not have.

**Floor**: one comment at the `s116w_fixture` builder stating both behaviours (`awk -v` strips one backslash level, `grep` does not) and which sites take which form. **Ceiling**: the six literals themselves, any `awk`/`grep` expression, `S116_END_RE`, and §116e/f/g's conditions and `ok`/`ng` text are all **out**. **Addition only — no existing line is edited.** That property is what separates this from every defect this PR family has generated: F5's extraction, F2's stale recast and this round's over-application all came from *changing* existing content. An appended comment cannot falsify an existing sentence. If it cannot be added without rewording something adjacent, it drops.

### N3 — fix-now. A false detail in the block round 1 already corrected

`:2637-2638` says §116 **and** §116c "both red with an ORDINARY PARITY MESSAGE — `expected=60` and an empty `levers=` field". Field census: `levers=` appears in §116's `ng` and in §116e/f/g's — **not** in §116c's, which is `classified=… expected=…`. So `expected=60` is true of both; the `levers=` half is true of §116 only.

Same direction as round 1's F1 on the same block, so this is a **completion of F1's correction, not a reversal**. Ruling it `drop` would be inconsistent with F1, which was ruled fix-now on this block for precisely this class — a false detail whose conclusion survives.

**Floor**: scope the `levers=` attribution to §116, at `:2637-2638` and in the PR body's Risks bullet, which carries the identical compound. **Ceiling**: the `expected=60` half **must remain attributed to both** (it is true of both); the "ordinary parity message" framing and the `unbound variable` clause stay; F1's conclusion at `:2639-2641` stands unweakened; `~30` is **not** folded in. **Prefer a scoping deletion over a recast** — removing a false clause cannot introduce a new false claim, whereas rewriting this block is what produced this finding.

### N2 — defer-to-issue, off the critical path

`e01e8ae` itself omits `Docs: n/a — <reason>`; 3/3 commits in this PR do, across two rounds, *after* round 1 named the lesson. The commit-body remedy is **unexecutable** (pushed; amend-after-push blocked), identical to round 1's I5. What defers is not the bodies but the **enforcement gap** — an unenforced prose norm decaying silently is the class #670 was filed for. Nothing asserts it is tracked, so it may be filed **after** merge.

The judge **disagreed with one sub-claim**: the reviewer said the PR body "still describes the omission as confined to the pushed commit bodies." At head, `e01e8ae` *is* a pushed commit body, so that sentence is accurate as written. No PR-body edit is warranted.

### N4 — drop, and the finding's framing is corrected

The reviewer wrote *"`~30 lines below` is now 23,"* implying this round degraded it. Measured, **the opposite**: the distance was **19** when the clause was written in `820e18b` and is **23** now — I1's move carried it *toward* `~30`. The clause was already loose the day it was authored and is now the most accurate it has ever been. Pre-existing, `~` is an explicit approximation marker, and editing it would re-enter, for a third consecutive round, the one comment block that has generated a fresh defect in both prior rounds. **Durable record: the definition is 19 lines above the function, 23 above the call.**

### Observations raised, not added to the fix list

1. **`~30 lines below` is a line-number anchor in prose**, and it decays on every edit to the intervening region. If anyone ever touches that clause, the durable move is to delete the number and say "above its first consumer" — not to update 23. Deliberately kept out of N3's ceiling so the exclusion is a decision.
2. **The reviewer's own line cite drifted** — it cites §116's `ng` at `:2703`; it is `:2706`. Substance unaffected (verified by symbol, not line), and a live demonstration of observation 1.
3. **Three consecutive rounds have found a defect in the `S116_END_RE` hoist-rationale block.** It is this PR's measured defect generator. After N3 lands, the next finding in it should face a strong presumption of `drop` absent a false *load-bearing* claim.
4. **Nothing enforces the escape split** — no arm compares the six sites' forms, exactly as nothing enforced #670's "verbatim" claim. The enforceable version belongs with **#670's AC3**, not in this PR.

ilgyu-yi (commented):
## Review — approve

`code-reviewer` returned **`ship`** at this head on a round-3 confirmation pass. The head was computed privately by the caller and never passed to the reviewer; the reviewer resolved it independently and the two matched, so the vote is valid under SPEC §4.5.

### Verdict path

| round | verdict | what it found |
|---|---|---|
| 1 | `ship after fix` | a false symptom claim; stale Phase-B tense; a sibling file's "verbatim" claim now false; a 4-char regression introduced by this PR's own extraction |
| 2 | **`ship`** (4 advisories) | the escape split was undocumented; one attribution detail false |
| 3 | **`ship`** | bounded remedies landed; nothing else moved |

### What the change does

`§116`'s **lever** window had literal headings at both ends, so it broke in both directions: a `### 1.8.5` sibling sat inside it (levers 6 → 7) and renumbering §1.8 meant it never opened (→ 0). Both feed `s116_exp`, which `§116` **and** `§116c` consume, so either defect reds two arms for a cause outside the parity being checked. The start now anchors on the section **title**; the end reuses the shared `S116_END_RE`.

Measured on today's SPEC and re-derived independently by the reviewer: baseline 6; `### 1.8.5` decoy **7 → 6**; §1.8 renumbered **0 → 6**; a genuine `#### 1.8.1` row 7 → 7. Both witnesses genuinely red pre-fix; `§116g` is labelled in-file as a **bound, not a witness**.

### Verification at this head

- Suite `pass=1287 fail=0`; §156j green; `SPEC.md` sha256 identical before and after a full run (AC5).
- `S116_END_RE` defined exactly **once**, consumed by all three sites — reused, not re-spelled (AC1).
- All six of #668's ACs discharged by independent measurement, not argument.

### Two things this review caught that the fix itself did not

**An over-application, caught by checking a count against a stated ceiling.** The `awk -v` escape fix was bounded to four sites; it was initially applied to six, including two `grep` calls. `grep` does not strip backslashes the way `awk -v` does — and the consequence was worse than a broken match: at §116f's leftover-heading guard the leg fires on `!= 0`, and `grep -c` with the wrong form returns **0 even on a file that has the heading**, so the guard would have **silently stopped guarding**. Reverted, and the split is now documented at the builder as a pure addition.

**Three consecutive rounds found a defect in one comment block** — the `S116_END_RE` hoist rationale. It is the measured generator, and the judge set a standing presumption of `drop` for the next finding in it absent a false *load-bearing* claim. Round 3's two remaining advisories were both left unedited on that ground.

### Disclosed and not fixed here

- **#670** — a sibling file's "§116's verbatim" claim, false in *both* halves and enforced by nothing. Filed and numbered **before** the comment citing it was written.
- A wrong-cause `ng` on a collapsed baseline, confirmed **inherited** from #644 — fixing it here would mean touching six arms or diverging three.
- The `Docs: n/a` norm, omitted in the two pushed commit bodies whose remedy is unexecutable; the newest commit carries it.

The marker below is the machine-readable signal for the merge-review gate; its `verdict` derives from the round-3 `ship` and its `head` is the head computed privately before the review, not re-derived at write time.

PR #671, opened 2026-08-16
fix(#653): name the arm that consumed a rejected skip token

### LINKED ISSUE #653: no audit record names which arm consumed a refused skip token
## What

When `_escape_token_honored` (`.claude/hooks/helpers/escape.sh`) refuses a skip token, it **consumes the token and returns a bare status code**. No record names *which* arm refused, so a legitimate token destroyed by a transient fault leaves nothing explaining why.

Seven arms consume the token and return silently. Line numbers are pinned to **PR #651's head `407bddd`** — see the sequencing note below.

| line | arm |
|---|---|
| 62 | **composite** — five clauses, enumerated below |
| 110 | malformed `created` |
| 111 | over-length `created` |
| 112 | `cmd_fingerprint` not a substring of the bind command |
| 114 | malformed clock (`now`) |
| 115 | over-length clock |
| 117 | future-dated or stale (the 60s TTL) |

Two further early returns — `esd` unresolvable (`:39`) and token absent/unreadable (`:41`) — do **not** consume, correctly: they never examined a token.

### The composite arm at `:62` is five clauses, and one of them is where this Issue's own scenario lands

```sh
if [ "$unknown" -ne 0 ] \
   || [ "${seen_cat}${seen_reason}${seen_fp}${seen_created}" != "1111" ] \
   || [ -z "$t_category" ] || [ -z "$t_fp" ] || [ -z "$t_created" ] \
   || [ "$t_category" != "$cat" ] \
   || [ -z "$bind" ]; then
```

1. unknown key
2. missing key
3. **present-but-empty value** (`category` / `cmd_fingerprint` / `created`)
4. category mismatch
5. empty bind command

Clause 3 is load-bearing for this Issue. `scripts/ghjig_skip.sh:80` mints the token with

```sh
printf 'created=%s\n' "$(date +%s)"
```

so a broken `date` **at mint time** writes `created=` — `seen_created=1` with `t_created=""` — caught by clause 3 at `:62`, **not** by the malformed-`created` arm at `:110`. The broken-`date` scenario this Issue is built around lands in the composite arm, which is also the highest-traffic one: it fires on every path that reads a token without a matching category.

## Scope boundary: what this Issue does NOT claim

`pre_tool_use.sh:44-49`:

```sh
block() {
  local cat="$1" msg="$2"
  audit_log block "$cat" deny "$msg"
  printf '%s\n' "$msg" >&2
  exit 2
}
```

For the `should_skip <cat> && decided=1 || block <cat> "<msg>"` call-site family (`:222`, `:241`, `:261`, `:284`, `:1196`, `:1212`, `:1228`, `:1266`, `:1271`, `:1285`, `:1294`, `:1299`, `:1376`, `:1381`, `:1416`), a refused token always reaches `block`, which logs. **On those paths the refusal IS recorded**, and SPEC §6.1's traceless-enforcement-disable prohibition is not violated.

That is verifiable directly from `block()` above — `audit_log block` is unconditional and precedes both the stderr write and the `exit 2`, so no path through `block` can skip it. Reproduce it by driving any `|| block` matcher with a refused token and counting records in the per-project `audit/audit.jsonl`.

The claim is scoped to that family deliberately. The `if should_skip <cat>; then … else … fi` sites (`:310`, `:409`, `:456`, `:495`, `:644`, `:701`, `:900`, `:1017`, `:1062`, `:1131`) run the matcher's own logic in the else branch, which may reach `mark_allow` — and `hookrt.sh:224-228` sets `decided=1` and returns **without writing a record**, deliberately, "without adding per-action audit noise." So on those paths a consumed token can leave zero records.

What is missing in **both** families is the same and is purely diagnostic: **no record names which arm consumed the token.** An operator sees the matcher's own message, or nothing, and cannot tell whether their token was malformed, expired, bound to the wrong command, minted with a broken clock, or refused for a category mismatch.

## Why

Before #651, a broken clock made the TTL fail *open*: the stale token was honored and logged as a routine `escape/skip` — a record that existed but **misdescribed** what happened. After #651 the same input is correctly blocked. #651 therefore **reduced** this surface rather than creating it: at base some rejects wrote zero records; at head none of the reject arms writes one, but the block itself is logged on the `|| block` family.

The residual is that the block record describes the *matcher's* decision, not the *token's* fate.

**MISSION fit:** "The mechanism" → the enforcement dual. MISSION names the audit line as the block's *deferred positive face*, latent until a consumer reads the aggregate. The face is present for the block; it is absent for the token.

**Cost asymmetry (SPEC §6.0 P1):** observability, not enforcement. The arms already fail closed and this Issue must not change that. A wrong *allow* is impossible here — there is nothing to allow. The cost is diagnostic, so the fix is a record, not a new refusal.

**Why now.** Filed as the executed deferral from PR #651's round-1 judged finding list, so the list's one `defer-to-issue` action has a terminal state rather than sitting unowned.

**Existing coverage:** none open. #558 is closed and concerns the `SKIP_HOOKS=directive-review` honor path, not the token reject arms. #650 covers the reviewer/judge execution surface, unrelated. The deferral is recorded in **PR #651's body**, not in #647's Out of scope.

## Sequencing — this Issue depends on #651

The seven-arm inventory presupposes PR #651 merged. On `main` today only **five** arms exist; `:114` and `:115` (the clock guards) are added by #651. An implementer starting before #651 merges would find AC1's inventory contradicting the code in front of them.

**This Issue should be worked after #651 merges.** The residual it describes survives #651 either way, so it is not premature — only ordered.

## Acceptance criteria

- [ ] **AC1** — each of the **seven** token-consuming reject arms emits an audit record naming **which** arm refused. The arm name is the whole payload — never token content — matching the `reason=<arm>` shape the review wrapper already uses.
- [ ] **AC2** — the composite arm at `:62` distinguishes its **five** clauses, **or** the implementation states why a single name is sufficient. If a single name is chosen, it must still let an operator distinguish the broken-`date`-at-mint case (clause 3) from a category mismatch (clause 4), because those have opposite remedies.
- [ ] **AC3** — the two non-consuming early returns (`:39`, `:41`) stay silent. They never examined a token; a record there would report a refusal that did not happen.
- [ ] **AC4** — the honor path's existing `escape/skip` record is unchanged, so a reader can still distinguish a taken escape from a refused one.
- [ ] **AC5** — fail-closed direction untouched: **every arm that blocks at `407bddd` still blocks**, and no new input is honored. Additive observability only. (Baseline pinned to a commit because `main` and #651's head have different arm sets.)
- [ ] **AC6** — a failing `audit_log` never converts a reject into an honor.
- [ ] **AC7** — a two-sided smoke arm: a refused token produces a record naming the arm, **and** a healthy honored token still produces exactly its one `escape/skip` record and no extra line.

## Out of scope

- **Changing which inputs are refused**, or any part of the TTL / fingerprint / consume-once contract. #647 owns those.
- **Claiming §6.1's traceless-disable prohibition is violated.** It is not, on the `|| block` family — see the scope boundary above.
- **Changing `mark_allow`'s deliberate silence**, or the `if/else` matcher families' record behaviour. This Issue adds arm attribution; it does not relitigate per-action audit noise.
- The audit record's transport or schema beyond the `reason=<arm>` convention already in use.
- The same silence in other helpers. A sweep is a separate Issue; this one is scoped to `_escape_token_honored`.
- Emitting anything derived from the token's contents. A record that echoed the token would leak exactly what the refusal withheld.

ilgyu-yi:
Closes #653

## Goal

A file-based skip token destroyed by a transient fault left **nothing** explaining why. Seven arms of `_escape_token_honored` consume the token and return a bare status; an operator saw the matcher's own message, or nothing, and could not tell whether their token was malformed, expired, bound to the wrong command, minted with a broken clock, or refused for a category mismatch.

MISSION *"The mechanism"* — the enforcement dual. MISSION names the audit line as the block's **deferred positive face**, latent until a consumer reads the aggregate. That face was present for the *matcher's* block (`block()` logs unconditionally) and absent for the *token's* fate.

This is **observability, not enforcement** (§6.0 P1). A wrong *allow* is impossible here — there is nothing to allow. The cost is diagnostic, so the fix is a record, not a new refusal.

## Plan

Type is `fix`, so §1.2's reproduce-first relaxation applies — reproduced with a live control before anything was written. **Doc still leads**, and genuinely: SPEC §7 already makes the *on-honor* record a contract, so the on-reject counterpart is a contract addition, not an implementation detail.

**A changelog fragment is required, not `skip-changelog`** — §18.7 requires one for a SPEC contract change.

## Key context

Reproduced on `main`, with the control proving the fixture live:

| probe | rc | consumed | records |
|---|---|---|---|
| **CONTROL** healthy token (honored) | 0 | yes | **1** (`escape/branch/skip`) |
| composite: empty `created` (clause 3) | 1 | yes | **0** |
| composite: category mismatch (clause 4) | 1 | yes | **0** |
| malformed `created` | 1 | yes | **0** |
| fingerprint not a substring | 1 | yes | **0** |
| stale (TTL 60s) | 1 | yes | **0** |

## Checklist

- [x] **Doc** (`561b1d1`): SPEC §7's **On reject** bullet + changelog fragment + regenerated TOC.
- [x] **Test** (`0f38948`): `§125-12a`–`§125-12d` witnesses, `§125-12e`/`§125-12f` bounds. Red before Code.
- [x] **Code** (`d4aa477`): 11 arm names across the 7 consuming arms, via one `_escape_reject` helper.

## Out of scope

Per #653's own list: changing **which** inputs are refused (#647 owns the TTL / fingerprint / consume-once contract); claiming §6.1's traceless-disable prohibition is violated (it is **not**, on the `|| block` family — `block()` logs unconditionally); `mark_allow`'s deliberate silence; the record's transport or schema beyond the `reason=<arm>` convention; the same silence in other helpers; and emitting anything derived from the token's contents.

## Risks

- **Splitting the composite `if` is the only semantics-sensitive edit.** It was verified rather than assumed: the pre-Code function was copied verbatim alongside the new one and both were run over **20 token shapes** (healthy, unknown key, missing key, each empty value, empty `reason`, category mismatch, empty bind, octal/alpha/hex/over-length `created`, unbound fingerprint, stale, future, TTL edges 60 and 61, blank line, absent token), comparing **both** rc and consume-vs-left. **0 differences.**
- **A failing `audit_log` must not flip a verdict** (AC6). Measured directly by shadowing it with a function returning non-zero: rejects still return 1 **and consume**; the honor still returns 0. There is no `set -e` in the hook's sourcing chain, and the helper emits → consumes → returns, never gating the return on the log.

## Test plan

`§125-12a`–`§125-12d` are the **witnesses** — red at `1289/4` before Code, green after. `§125-12d` is deliberately a **non-composite** arm, so AC1 is pinned beyond the composite. `§125-12c` is AC2's real requirement: the two composite clause names must **differ**, and it requires both non-empty first so two absences cannot green it by being equal.

`§125-12e` and `§125-12f` are **bounds, not witnesses**, and say so in-file — they pass both before and after, because there is nothing broken on the honor side to witness.

**Anti-vacuity sits on consumption**, which is the discriminator this fixture family turns on: a probe that lands on the non-consuming fast path returns 1 and writes nothing, indistinguishable from the defect by return code alone. Every witness is gated on `rc=2` **and** a block record **and** `token-left=0`.

**Measured, no live agent worktree:** `pass=1289 fail=4` → **`pass=1293 fail=0`**; `lint OK: 79 files clean`. Independently confirmed that a reject with a token carrying `reason=SECRET-REASON` and `cmd_fingerprint=SECRETFP…` emits `reason: "parse-empty-value"` and leaks **zero** occurrences of that content.

## Round-1 review

Both reviewers cleared it — `code-reviewer` **`ship`**, `security-reviewer` **APPROVE** (no High, no Medium), each resolving the head independently. The security review built its own **43-shape** honored-set differential (a superset of this PR's 20) and found **0 divergences**, and proved the harness non-vacuous by showing the base wrote zero records where the head wrote one.

The judged list is posted above: **1 fix-now, 2 defer, 6 drop**. Two rulings are worth surfacing.

**The security reviewer's one Low finding was confirmed but its remedy refuted.** `_escape_reject` emits before it consumes, so a *process-terminating* `audit_log` would leave the token at rest. Consume-first-then-emit is **not** safer — it is symmetric: the read-only case is order-**invariant** today (measured, base and head both leave the token at rest), and under the same hypothetical `set -e` that makes the finding live, consume-first would exit on the failing `rm -f` **before** the emit — token at rest **and zero records**, which is precisely the silent destruction #653 exists to eliminate. It is also outside #653's scope (#647 owns consume-once) and would contradict this PR's own Doc commit. Deferred to #647's territory, carrying the ruling that consume-first is **contested, not prescribed**.

**One idiom note was answered with a third option.** The reviewer suggested naming *which* key was empty, offering "fold in now" or "later Issue". The judge ruled **drop**: the harm is unreachable through the sanctioned mint path, because `ghjig_skip.sh` refuses an empty category and an empty-or-short fingerprint before writing — so only `created=` can reach that arm from a minted token, and `parse-empty-value` on a real token already means exactly "fix the clock".

## Notes

Three findings from the Code phase, none blocking:

1. `_audit_validate_format` **substitutes** a `format-error` line on violation rather than failing loudly. Today only `directive-file`×`created` is strict, so `escape`/`reject` is unaffected — but a future strict arm for an escape category would silently reshape these records rather than red.
2. **The honor path still embeds the token's own `reason` verbatim** (pre-existing, and AC4 requires it unchanged). So "never echo token content" is a property of the **reject** side, not of the whole function — the SPEC bullet is scoped to `On reject` for that reason.
3. `_escape_reject` calls `audit_log` unguarded — **but this is the file's existing convention, not a fragility this PR introduces.** The base already carries two call sites of the identical shape (`should_skip`'s `SKIP_HOOKS` branch, and the honor path); the Code phase adds a third of the same class. A consumer sourcing `escape.sh` without `hookrt.sh` would get a stderr `command not found` per reject, and the reject verdict still stands (same shape as the AC6 probe). *(Corrected in round 1: this note originally presented the shape as newly introduced.)*


ilgyu-yi:
## Finding triage (round 1)

`reviewed-head: d4aa477a8be5f13a8a90a83256e0ba2e08b73ec2` · `prior-round: none (first round)`

**Both reviewers cleared it** — `code-reviewer` **`ship`**, `security-reviewer` **APPROVE (no High, no Medium)** — each resolving the head independently and matching the caller's privately-held head, so both votes are valid. No blocker-stop. Everything below is advisory, judged by a fresh worktree-isolated `finding-judge` reading via `git show <head>:`.

**9 findings, 9 confirmed, 0 refuted. fix-now ×1, defer ×2, drop ×6.**

### The one fix-now — and it is my own PR body over-stating a finding

**CR-NOTE**: the body's Note 3 presents `_escape_reject`'s unguarded `audit_log` call as a fragility this PR introduces. Measured, the base **already** carries two call sites of the identical shape — `should_skip`'s `SKIP_HOOKS` branch and the honor path. The Code phase adds a third of the same class, not a new class. **Bound: the PR body's Note 3 prose only.** Explicitly not to be touched alongside it: `escape.sh` (any byte), SPEC §7's On-reject bullet, the smoke file, the changelog fragment.

### SEC-LOW-1 — defer, and the proposed remedy does **not** survive

The finding is real and reproduced: `_escape_reject` emits before it consumes, so `audit_log(){ exit 9; }` leaves the token at rest where base consumed it. The reviewer's remedy — consume-first-then-emit — was refuted on four measured grounds:

1. **It is symmetric, not safer.** The read-only counterfactual was measured: head and base *both* leave the token at rest with `fnrc=1`, so that case is order-**invariant** today. And under the same hypothetical `set -e` that makes SEC-LOW-1 live, consume-first would exit on the failing `rm -f` **before** the emit — token at rest **and zero records**, which is exactly the silent-destruction defect #653 exists to eliminate. Neither order dominates.
2. **Unreachable at head.** `hookrt.sh`'s `audit_log` contains no `exit` — only `date`/`pwd`/`mkdir -p`/`printf`/`jq`. The *reachable* failure (returning non-zero) was measured and the verdict holds: `fnrc=1`, token consumed. That is AC6's actual contract.
3. **Outside #653's declared scope**, which excludes "any part of the TTL / fingerprint / **consume-once** contract — **#647 owns those**". Emit/consume ordering is that contract's ordering semantics.
4. **It would contradict this PR's own Doc commit**, landed three commits ago, which states emit → consume → return "in that order and all unconditional" as load-bearing.

Against all that, touching `escape.sh` re-opens the honored-set verification both reviewers cleared (43 shapes / 0 divergences independently, plus the PR's own 20-shape differential). **Filed against #647's territory, not as a #653 follow-up**, and it must carry the ruling that consume-first is **contested, not prescribed** — the framing that survives is *"make the consume unconditional with respect to `audit_log`'s implementation without making the emit conditional on the consume"*, which is not reducible to a reordering.

### SEC-INFO-2 — defer to a `hookrt.sh` hardening Issue

`audit_log` JSON-encodes only `reason` and `cwd`; `category` and `decision` go out as bare `%s`. Measured: both corrupt the JSONL on a non-literal. **Exposure added by this PR is zero** — `_escape_reject` passes the literal `reject` and the caller's own category constant, and `ghjig_skip.sh` additionally constrains the write side to `[A-Za-z0-9,_-]`. Excluded by #653's Out of scope ("the audit record's transport or schema beyond the `reason=<arm>` convention").

### Dropped (6), each with its ground

- **SEC-INFO-3** — the unguarded call matches two pre-existing sites; same shape, no new class, and the verdict is unaffected.
- **SEC-INFO-4** — the arm-name "oracle" **is the contract**, not a leak: SPEC §7 requires the arm name as the whole payload, the same principal reads the source and the log, and §6.1 states hooks are mistake-prevention, not a security boundary.
- **CR-IDIOM-3** — bare `$1 $2 $3`; the doc comment three lines above already binds the names.
- **CR-IDIOM-4** — the 56-line gap is real, but the comment block is #647's measurement record and is unchanged here; the split merely widened the distance.
- **CR-IDIOM-5** — the judge declined both offered options and ruled a third. AC2 is satisfied (all five clauses carry distinct names matching SPEC's enumeration), and the reviewer's residual — *which key* was empty — is **unreachable through the sanctioned mint path**: `ghjig_skip.sh` refuses an empty category and an empty-or-short fingerprint before writing, so only `created=` can reach this arm from a minted token. `parse-empty-value` on a real token already means exactly "fix the clock". Sub-naming would diverge from SPEC's just-landed five-name list to disambiguate a state no sanctioned mint can produce.
- **CR-IDIOM-6** — the fragment is 134 words / 897 chars on one line: denser than the reviewer estimated, and §18.1-compliant as written.

### Observation, not a judged record

The body's Note 1 — `_audit_validate_format` substituting a `format-error` line rather than failing loudly — is the same class as SEC-INFO-2 and belongs in the same `hookrt.sh` hardening Issue. No reviewer raised it as a finding, so the judge made no ruling on it.

ilgyu-yi (commented):
## Review — approve

`code-reviewer` returned **`ship`** and `security-reviewer` returned **APPROVE** (no High, no Medium) at this head. The head was computed privately by the caller and never passed to either reviewer; both resolved it independently and matched, so both votes are valid under SPEC §4.5. The round-1 fix touched **PR prose only** — no commit, no code — so the head both reviewers cleared is still the head being merged.

### What the change does

`_escape_token_honored` — the enforcement-**bypass** reader — had seven arms that consume a skip token and return a bare status, writing nothing. A token destroyed by a transient fault left no explanation. Each arm now emits `audit_log escape <category> reject "<arm-name>"`, with the composite arm naming its five clauses separately.

That last part is load-bearing, not cosmetic: a `date` broken **at mint time** writes `created=` (present but empty), which lands in the empty-value clause and **not** the malformed-`created` arm — and that case's remedy is the opposite of a category mismatch's.

### The one thing that mattered, verified twice independently

This edit restructures the composite reject condition from one `||`-chain into sequential named checks — the exact shape where a bypass can widen by accident. The honored set was proved unchanged by **differential**, not by inspection:

- The author ran the pre-Code function verbatim alongside the new one over **20 token shapes** — 0 differences on both rc and consume-vs-left.
- The `security-reviewer` built its own **43-shape** corpus independently (a superset, adding clock shapes the author's list omitted) — **0 divergences**, and proved the harness non-vacuous by showing the base wrote **zero** records where the head wrote exactly one, so two genuinely different files were loaded.

Both confirmed the TTL edges (60 honors, 61 rejects), each empty-value sub-condition, and that an empty `reason` is still **honored** — `reason` is not an honor condition.

### Other verification at this head

- **No leak**: a token carrying a distinctive `reason` and `cmd_fingerprint` produces `reason: "parse-empty-value"` with zero occurrences of either. Even the token's own `category` never reaches the log — the record carries the *requested* category.
- **AC6**: with `audit_log` shadowed to return non-zero, and separately made undefined, rejects still refuse **and consume**, honors still honor.
- **Consume-once preserved** across all eleven reject arms; the two non-consuming early returns still consume nothing and stay silent.
- Suite `pass=1289 fail=4` → **`pass=1293 fail=0`**, the four flips being exactly the intended witnesses; `lint OK: 79 files clean`.

### Judged findings

**1 fix-now, 2 defer, 6 drop.** The fix-now was the PR body over-stating one of its own disclosures, now corrected.

The security reviewer's single Low finding was **confirmed but its remedy refuted**: emit-before-consume is order-sensitive to a *process-terminating* `audit_log`, but consume-first is **symmetric, not safer** — the read-only case is order-invariant today, and under the same hypothetical `set -e` that makes the finding live, consume-first would exit on a failing `rm -f` before the emit, giving token-at-rest **and zero records**, precisely the silent destruction this issue exists to eliminate. Deferred to #647's territory carrying the ruling that consume-first is contested, not prescribed.

Two follow-ups are owed and **not yet filed** — the ordering item above, and a `hookrt.sh` hardening item (only `reason` and `cwd` are JSON-encoded; `category` and `decision` go out bare). Nothing in this PR claims either is tracked.

The marker below is the machine-readable signal for the merge-review gate; its `verdict` derives from the `code-reviewer` `ship`, and its `head` is the head computed privately before the reviews, not re-derived at write time.

PR #672, opened 2026-08-17
fix(#643): bound the §1.3.1 content-lock window on both sides

### LINKED ISSUE #643: smoke §156 window guard has a floor but no ceiling, so over-extension is silent
## What

`scripts/test/smoke.d/70-gates-contentlocks.sh:1884` extracts the SPEC §1.3.1 window with a fence-aware awk, and `:1896` guards it with a **floor** (`-lt 18`). There is no ceiling. The fence-awareness is exactly what makes a ceiling necessary, and it was added without one.

Measured in the PR #641 round-3 review: inserting **one unclosed** ` ``` ` inside §1.3.1 desyncs the `f` toggle, so the heading terminator never fires, and the window runs to **364 non-blank lines** — swallowing §1.4 through §3. `§156a` then reports *"window resolves non-empty … content locks below are load-bearing"*.

Before fence-awareness an unbalanced fence failed **loud**: the window truncated and the floor caught it. After fence-awareness it fails **silent**: the window over-extends and nothing bounds it. The failure mode was not removed, it was inverted.

**Not currently a false pass**, and this matters for prioritisation: the reviewer gutted the section *and* left an unbalanced fence, and all eleven content locks still failed correctly, because each requires a phrase conjunction absent from §1.4–§3. So this is a hardening gap with no live false-green, which is why it is P2 rather than P1.

Two smaller items in the same file, worth folding in since they need the same touch:

- `:1897` — the guard's provenance comment says `21 non-blank lines at this commit`; the true count at head is **22** (`§156a` prints it). The 21 was correct at `f5af42c` and went stale when that same commit added a prose line to §1.3.1. Provenance only; the floor that runs is correct.
- `§156m` (`:1985-1987`) locks on three bare unanchored words (`widened`, `never be narrowed`, `re-anchor`) where its siblings pin phrases or line-anchored forms. Two of the three are individually satisfiable by unrelated prose inside the window; only the conjunction carries the lock. Tightening to `may be \*\*widened\*\*` and `\*\*re-anchor\*\* accompanying a heading rename` would make it bite as hard as `§156l`.

## Why

A guard that reports "load-bearing" over a window containing three unrelated SPEC sections is asserting something it has not checked. §156a exists specifically to convert the vacuous-lock anti-pattern into a loud failure — the anti-pattern named in `scripts/test/smoke.sh`'s own header — and a one-sided bound only converts half of it.

**Why now:** the fence-aware extractor landed in `4c23a93` (PR #641) and created this gap in the act of closing the truncation gap. Filing immediately rather than later, while the reproduction is exact.

**MISSION fit:** the audit-and-gate machinery is the shell's enforcement face. A guard whose failure direction is untested is the "wrong nudge" §6.0 P1 requires to be cheap, and a silently-unguarded window is not cheap — it makes eleven locks look verified when their scope is unknown.

**Existing coverage:** none. No other §156 arm bounds the window size, and §116 does not read this window.

## Acceptance criteria

- [ ] `§156a` bounds the window in **both** directions — a floor and a ceiling — and both bounds are stated in the assertion message so a failure names which side tripped.
- [ ] A regression arm asserts the **over-extension** direction: a SPEC copy with one unclosed fence inside §1.3.1 must fail `§156a` loud, not report the locks load-bearing. The real `SPEC.md` is never mutated.
- [ ] A paired arm asserts the **truncation** direction still fails loud (renamed heading → `non-blank lines=0`), so the ceiling does not blind the floor.
- [ ] A third arm asserts the healthy window passes, so the two bounds together are not simply always-failing.
- [ ] The `:1897` provenance count is re-measured rather than re-typed.
- [ ] `§156m`'s two weak patterns are pinned to phrases.

## Out of scope

Making the extractor handle nested or four-backtick fences by tracking fence length. The ceiling covers those cases by bounding the consequence rather than parsing the cause, and parsing markdown fences correctly in awk is a larger job than this guard warrants.

## Notes

Found in the round-3 `code-reviewer` review of PR #641 (Execution Issue #640, Directive #636).

**One pattern worth recording, not yet doctrine.** This is the third instance in one session of the same shape: a defect "fixed" by moving it to the opposite end of the same axis rather than by installing a two-sided bound.

1. Window truncation → window over-extension (this Issue).
2. A stale provenance count corrected to a different stale count (`:1897`).
3. In Directive #636's design arc, four successive retired-vocabulary instruments: always-pass (agent-recorded set) → always-fail (raw removed-line tokens) → tautology (minus globally-present) → pure noise (minus added lines).

All three above are one author in one session. **That limitation no longer holds:** the operator reports having observed the same shape **three times in three different repositories**, independently of this session. Cross-repo recurrence is the threshold §6.0 P3's born-advisory → measure → harden order names, so this is past the observation stage.

**One check before it is written as a rule**, because mis-generalising the class is the failure this Issue is about: confirm the other two instances are the same *axis-overshoot* shape rather than a look-alike. The discriminator is mechanical — for a fix responding to finding N, does it sit at the **opposite extreme of the same axis** as failure N-1, rather than at a bounded position on it? A `fix(` commit sequence plus the PR body's review-round narration is enough to classify each case.

**Where the rule belongs, once confirmed:** not a new Directive. Directive #639's escalation ladder already requires a round-2 resubmission to carry a *method change*, and its primary success signal already reserves a **near-miss arm** for "textually different while describing the same method". A swing is exactly that near-miss: it presents as a different method while occupying the same axis at the opposite sign. #639 deliberately left the near-miss arm's evaluation to its Execution Issue — a defensible call at one observed instance, and the wrong call at three. Revising #639 to define the near-miss criterion as *a fix that moves the failure to the opposite end of the same axis is not a method change* places the rule where the enforcement already exists, and adds no new gate. Filed with the `issue-review` gate bypassed and audit-logged: the finding is a measured reproduction with exact counts, and the rationale triad is stated above.

ilgyu-yi:
Closes #643

## Goal

`§156a` guarded the SPEC §1.3.1 content-lock window with a **floor and no ceiling**, so over-extension was silent. The fence-awareness that *made* a ceiling necessary was added without one.

MISSION *"The mechanism"* — the audit-and-gate machinery is the shell's enforcement face. A guard that reports eleven content locks "load-bearing" over a window containing three unrelated SPEC sections is asserting something it has not checked. `§156a` exists specifically to convert the vacuous-lock anti-pattern into a loud failure; a one-sided bound converts only half of it.

## Plan

Type is `fix`, so §1.2's reproduce-first relaxation applies — both directions were measured on `main` before anything was written.

**Docs: n/a — no SSOT contract change.** The window guard is a smoke-arm implementation detail with no SPEC clause; the arm's own comments are its documentation. Test-only surface, so `skip-changelog` per §18.7.

The planner gate did not fire: single file, well under the 3+ threshold.

**Deviation, disclosed:** this PR was opened **ready, not draft-first**. Both phase commits were complete before it was created, so the backbone's `draft PR → checklist commits → ready PR` cadence was skipped. Named here because the Plan discloses two other deviations and silently omitting a third makes the list read as complete when it is not.

## Key context

One unbalanced fence does not make the toggle *stick* — it **inverts the parity** from that point on, so headings sitting in real prose stop terminating the window while `#`-lines inside genuine fences start doing so:

| fixture | non-blank lines | floor-only verdict |
|---|---|---|
| healthy SPEC | 22 | passes |
| **one unclosed fence** in §1.3.1 | **368** | **passes — silent over-extension** |
| §1.3.1 heading renamed | 0 | fails loud |

Measured on the over-extension fixture (SPEC.md plus one injected marker, 2895 lines; SPEC.md itself is 2894), the window spans 269–739 and exits at **740** on a shell comment inside a ` ```bash ` fence — stable at 368 across six injection offsets. Nothing bounds that over-extension from above.

## Checklist

- [x] **Test** (`e8aa0db`): `§156o` (bounds named) + `§156p` (over-extension) as witnesses; `§156q` (truncation) and `§156r` (healthy) as bounds. Red before Code.
- [x] **Code** (`3a6297a`): two-sided bounds, the re-measured provenance count, and `§156m`'s two weak patterns pinned.
- [x] **Review fixes** (`1ada0a0`, `f52ded4`, `fb12283`): round-1, round-2 and round-3 judged lists.

## Out of scope

Making the extractor handle nested or four-backtick fences by tracking fence length — per #643's own Out of scope, the ceiling covers those by **bounding the consequence rather than parsing the cause**, and parsing markdown fences correctly in awk is a larger job than this guard warrants.

## Risks

- **A numeric ceiling is a drift-prone bound**, and this very PR fixes a stale count of the same kind (AC5). Mitigated by margin rather than precision: `ceiling=60` leaves §1.3.1 room to **nearly triple** from 22 while sitting **six-fold below** the ~368 a desync produces. It is a consequence bound, not a size assertion, and the comment says so.
- **The bound spellings are load-bearing, but they fail *closed*.** `§156o`/`§156p` parse `floor=<n>` and `ceiling=<n>` out of `§156a`'s assertion text. *(Corrected across two rounds: round 1 called a rename a **silent** unbind; round 2's replacement explained it with a mechanism that was also backwards. Measured outcome: both arms red **loud**.)* The real, undisclosed vector was different — see below.

## Review rounds (`1ada0a0`, `f52ded4`, `fb12283`)

`code-reviewer` returned **`ship after fix`** with four must-fix findings — **all four were false claims in my own comments**, in a PR whose `Docs: n/a` explicitly elevates those comments to documentation. The judged list is posted above: **5 fix-now, 2 defer, 2 drop, 1 refuted.**

Three rulings are worth surfacing.

**F2's remedy was a scoping deletion, and the judge refused a replacement narrative it had itself verified as accurate.** I wrote that the count "read 21, which was correct at `f5af42c`". Measured, `f5af42c` carried **18**; `21` came from `4c23a93`, where the count was **already 22** — stale on arrival. AC5 requires only that the **count** be re-measured; the narrative carries **zero** enforcement value, because §156j doctrine leaves provenance counts unpinned so nothing can ever catch a story about it drifting. This line has now carried two successive wrong provenance stories, so the comment says *re-measure it, do not re-tell it*.

**One finding was REFUTED, and acting on it would have introduced a fourth false claim — through the review channel itself.** The reviewer said the MISSION citation was wrong because "The mechanism" is only about context narrowing. That section spans lines **10–22**, and **line 18** is an entire paragraph on enforcement. The reviewer read 10–16 and stopped immediately before it.

**One reviewer figure was not copied forward.** Its F1 text says "84% of the file is never reached"; measured, **74.5%**. Copying it would have made the repair of a false claim introduce the next one.

**The two pushed commit bodies are unamendable** and still carry the superseded "runs to EOF" mechanism and the wrong provenance history. Corrected forward here rather than by rewriting reviewed history.

## Round 2 (`f52ded4`) — the generator, measured

Round 2 returned `ship after fix` again, and **my round-1 repair had introduced a new false claim**: I wrote that a rename fails closed *"because the accept-check rejects an empty bound"*. The outcome is true; the mechanism is backwards — `s156o_accepts` short-circuits on an empty ceiling and falls through to `return 0`, so it **accepts**. The correct mechanism was already stated correctly **twice** in the same file, 140 lines below.

**The judge resolved the generator into something measurable.** Of the five edit clusters in `1ada0a0`, the **four that deleted or scoped** a claim are verified **true**; the **one that added an explanatory `because`** is verified **false** — 4-for-4 versus 0-for-1. The discriminator carried forward: *keep a claim only if it is verifiable against a cheap, stable oracle.* A fixed-commit fact is one `git show` away and can never go stale — keep it. A mechanism explanation needs a mutated-fixture run and has been wrong twice — delete it. The file already said *"Re-measure it; do not re-tell it"* **four lines above** the clause that violated it.

A second generator was also named: **corrections not reaching every surface** — round 2 reproduced it twice. The adopted counter is an **enumerated surface sweep**, and it worked: it found a **third** stale tense claim the review had not cited.

It also surfaced a **pre-existing** false claim of the same class in the §157 block (`157v–157x` described as "LOAD-BEARING RED at this commit"; all three are green today). It is from another PR and untouched here — deliberately **not** fixed, because that is the while-I'm-here move that produced both prior rounds. Recorded rather than repaired.

## Round 3 (`fb12283`) — the criterion fires, the swing is refused

Round 3's `code-reviewer` found **zero code defects for the third round running** and re-measured every fixed-commit claim round 2 installed as **true**. It found two more false/stale prose claims, both in text the two prior sweeps had passed over. The judge confirmed 5 of 5 findings, **upgraded** one the reviewer had marked no-edit, and **dropped** one the reviewer had also declined — after measuring all three rename variants rather than agreeing on sight.

**Round 2's stopping criterion is TRIGGERED**, and the reviewer missed it: it checked only the two findings in older text, never its own `A2`. The sentence *"the mechanism is stated once"* is **new in `f52ded4`**, is **false** (`grep` finds two sites), and contradicts the judged list that authorized the edit — round 2's own ruling said "stated correctly **twice**" while quoting both. So the generator survived a targeted counter, and the cheap-stable-oracle discriminator did not prevent it. The lesson the judge draws: *the discriminator selects which claims are keepable, but nothing forces the check to actually run.*

**The licensed consequence — "strip the remaining prose to assertions only" — was refused** as `unjustified — needs measurement`. Its only ground is the direction of the previous failure, which is not a position; it is an opposite-end move from round 2's deliberate `interior`; and a mass strip is itself the large-prose-edit class carrying the measured ~1-defect-per-round rate. It would delete text three rounds measured true and load-bearing, including the F4b **self-disclosure** at `:2497-2504`. Deleting honest self-disclosure to punish dishonest prose is a strictly worse artifact.

**The bounded rule adopted instead, decidable for round 4:** the prose is not stripped, but **no clause here may be corrected a fourth time — a clause measured false a third time is deleted, never rewritten.** All three round-3 fixes discharge under it already: two deletions and one token anchored onto the convention this PR had already adopted.

## Test plan

`§156o` and `§156p` are the **witnesses** — red at `1295/2` before Code, green after. `§156p` reads the guard's **behaviour** (does it reject an over-extended window?); `§156o` reads its **text** (does a failure name which side tripped?), which is AC1 made machine-readable.

`§156q` and `§156r` are **bounds, not witnesses**, labelled as such in-file under a `WITNESS HONESTY` block. `§156q` passes today only because the floor already rejects 0 — its job is to red **later** if a ceiling implementation blinds the floor. Its fixture is validated by its **marker** (renamed heading present, original absent) rather than by its count, because reading `0` as proof the fixture built would be circular when `0` is the very thing the arm asserts. `§156r` additionally pins that the fixture path still reproduces the live guard, by requiring its count to equal the live one.

The arms drive the live guard's **extractor** verbatim — but its **decision** is re-implemented from the two integers scraped out of the assertion messages, so an executed condition that *diverges* from the declared bounds is **not** caught by these arms. That is a real, measured vacuity vector: delete the ceiling from the executed condition only, leave the messages untouched, and all four arms stay green while the guard carries no effective ceiling. It is **deferred, not fixed here** — the proposed hoist would make the arms an unbound-variable **abort** against the pre-fix guard, degrading the witness reproduction from a clean `§156p` RED to a crash, and that RED is this PR's entire evidence the defect was repaired. **No Issue tracks it yet.**

**AC5 — re-measured, not re-typed.** The comment carried a stale `21`; the re-measured count is **22**. No history is reconstructed, deliberately: the count is **not** pinned by an arm (per `§156j`'s doctrine that provenance counts stay unpinned), so nothing can ever catch a narrative about it drifting — and this line had already carried two successive wrong provenance stories. *(Round 2: this paragraph still asserted the refuted narrative after the code comment was corrected — the second instance of a fix not reaching every surface.)*

**AC6 — phrases, not bare words.** `widened` and `re-anchor` were each individually satisfiable by unrelated prose anywhere in the window, so only the conjunction carried the lock. Both are now pinned to the forms §1.3.1 actually uses, verified against the live window, so each conjunct bites on its own.

**Measured, no live agent worktree:** `pass=1295 fail=2` → **`pass=1297 fail=0`**; `lint OK: 79 files clean`; real `SPEC.md` unmodified after a full run.

### DIFF
--- a/scripts/test/smoke.d/70-gates-contentlocks.sh
+++ b/scripts/test/smoke.d/70-gates-contentlocks.sh
@@ -2338,8 +2338,9 @@ elif [ "$s156_n" -lt 18 ] || [ "$s156_n" -gt 60 ]; then
   # mode rather than removing it (#643). ONE unbalanced ``` inside §1.3.1 does not
   # make the toggle stick — it INVERTS THE PARITY from that point on, so headings
   # sitting in real prose are ignored while `#`-prefixed lines inside genuine
-  # fenced blocks are honoured as terminators. Measured: the window then spans
-  # SPEC 269-739 and exits at 740 on a shell comment inside a ```bash fence — 368
+  # fenced blocks are honoured as terminators. Measured on the over-extension
+  # fixture (SPEC.md plus one injected marker, 2895 lines): the window spans
+  # 269-739 and exits at 740 on a shell comment inside a ```bash fence — 368
   # non-blank lines against a healthy 22, stable across six injection offsets.
   # Before fence-awareness an unbalanced
   # fence TRUNCATED and the floor caught it loud; after it, the window
@@ -2356,8 +2357,9 @@ elif [ "$s156_n" -lt 18 ] || [ "$s156_n" -gt 60 ]; then
   # renamed heading; ceiling=60 leaves §1.3.1 room to nearly triple while sitting
   # six-fold below the ~368 a desync produces today. The bounds are NAMED in both
   # messages so a failure says which side tripped, and §156o/§156p parse those two
-  # spellings — renaming them fails CLOSED (measured: both arms red, because the
-  # arms' accept-check rejects an empty bound), so it is not a silent hazard.
+  # spellings — renaming them fails CLOSED (measured: both arms red), so it is not
+  # a silent hazard. The mechanism is stated once, at §156o's accept-check below;
+  # it is deliberately not restated here.
   ng "156a: SPEC §1.3.1 window out of bounds (non-blank lines=$s156_n, floor=18, ceiling=60) — below the floor the content locks would pass VACUOUSLY; above the ceiling the window has over-extended past §1.3.1 and the locks would be asserted over unrelated sections (#640, #643)"
 else
   ok "156a: SPEC §1.3.1 window resolves within bounds (non-blank lines=$s156_n, floor=18, ceiling=60) — content locks below are load-bearing (#640, #643)"
@@ -2470,21 +2472,23 @@ fi
 # ---------- §156o–§156r: the §1.3.1 window is bounded on BOTH sides (#643) ----------
 # The count-guard above is a FLOOR only. The fence-awareness that closed the
 # truncation gap (#641) opened the opposite one: with an ODD number of ``` markers
-# inside §1.3.1 INVERTS the `f` parity from that point on, so headings in real prose
-# stop terminating the window and it runs on — measured, to line 740 of 2895,
+# inside §1.3.1 the extra marker INVERTS the `f` parity from that point on, so headings in real prose
+# stop terminating the window and it runs on — measured, to line 740 of the 2895-line
+# over-extension fixture (SPEC.md is 2894),
 # swallowing §1.4 through §3.2. A floor cannot see that; the
 # guard then reports the eleven content locks "load-bearing" over a window whose
 # scope it has not checked, which is anti-pattern #2 in the smoke.sh header wearing
 # the opposite sign.
 #
 # WITNESS HONESTY — only ONE of the four arms below is a witness:
-#   §156p (over-extension) is the WITNESS. It is RED at this commit and greens only
+#   §156p (over-extension) is the WITNESS. It is RED at e8aa0db (the Test commit,
+#   guard floor-only) and greens only
 #          when a ceiling exists.
 #   §156o (both bounds named), §156q (truncation), §156r (healthy) are BOUNDS, not
 #          witnesses. §156q and §156r pass BEFORE and AFTER the fix by construction —
 #          they exist to prove the ceiling does not blind the floor (§156q) and that
 #          the pair is not simply always-failing (§156r). Neither is evidence that the
-#          defect was repaired; only §156p is. (§156o is red today for the same reason
+#          defect was repaired; only §156p is. (§156o is red at e8aa0db for the same reason
 #          §156p is, but it reads the guard's TEXT, not its behaviour.)
 #
 # NO LIVE SPEC MUTATION. All three fixtures are copies under the preamble's $TMP,
@@ -2562,7 +2566,7 @@ else
 fi
 
 # §156p — AC2, THE WITNESS. One unclosed fence inside §1.3.1 must make the guard fail
-# LOUD. RED at this commit: the floor admits the over-extended window.
+# LOUD. RED at e8aa0db: the floor admits the over-extended window.
 if [ "$s156o_hn" -lt 0 ] || [ "$s156o_on" -lt 0 ]; then
   ng "156p: over-extension fixture not built (healthy n=$s156o_hn over n=$s156o_on) — the ceiling is UNTESTED, not satisfied (#643)"
 elif [ "$s156o_ofence" -ne $((s156o_hfence + 1)) ]; then
--- a/scripts/test/smoke.d/70-gates-contentlocks.sh
+++ b/scripts/test/smoke.d/70-gates-contentlocks.sh
@@ -2440,6 +2440,136 @@ else
   fi
 fi
 
+# ---------- §156o–§156r: the §1.3.1 window is bounded on BOTH sides (#643) ----------
+# The count-guard above is a FLOOR only. The fence-awareness that closed the
+# truncation gap (#641) opened the opposite one: with an ODD number of ``` markers
+# inside §1.3.1 the `f` toggle sticks at 1, the heading terminator never fires, and
+# the window runs to EOF — swallowing §1.4 through §3. A floor cannot see that; the
+# guard then reports the eleven content locks "load-bearing" over a window whose
+# scope it has not checked, which is anti-pattern #2 in the smoke.sh header wearing
+# the opposite sign.
+#
+# WITNESS HONESTY — only ONE of the four arms below is a witness:
+#   §156p (over-extension) is the WITNESS. It is RED at this commit and greens only
+#          when a ceiling exists.
+#   §156o (both bounds named), §156q (truncation), §156r (healthy) are BOUNDS, not
+#          witnesses. §156q and §156r pass BEFORE and AFTER the fix by construction —
+#          they exist to prove the ceiling does not blind the floor (§156q) and that
+#          the pair is not simply always-failing (§156r). Neither is evidence that the
+#          defect was repaired; only §156p is. (§156o is red today for the same reason
+#          §156p is, but it reads the guard's TEXT, not its behaviour.)
+#
+# NO LIVE SPEC MUTATION. All three fixtures are copies under the preamble's $TMP,
+# removed by its EXIT trap; $S156_SPEC is only ever read.
+#
+# The arms drive the LIVE guard, not a paraphrase of it: the extractor is lifted
+# verbatim out of the guard's own source line, and the bounds are read out of the
+# `156a:` assertion messages — i.e. the bounds the guard TELLS a failing reader about.
+# A bound the guard does not declare cannot bind anything, so an undeclared ceiling
+# reads here exactly as it behaves there: absent.
+S156O_SRC="$SHELL_ROOT/scripts/test/smoke.d/70-gates-contentlocks.sh"
+s156o_msgs=""
+s156o_floor=""
+s156o_ceil=""
+s156o_prog=""
+if [ -f "$S156O_SRC" ]; then
+  s156o_msgs=$(grep -E '^[[:space:]]*(ok|ng) "156a:' "$S156O_SRC")
+  s156o_floor=$(printf '%s\n' "$s156o_msgs" | grep -oE 'floor=[0-9]+' | head -1 | cut -d= -f2)
+  s156o_ceil=$(printf '%s\n' "$s156o_msgs" | grep -oE 'ceiling=[0-9]+' | head -1 | cut -d= -f2)
+  # Line-anchored so this arm's own quoted copy of the pattern cannot match first.
+  s156o_line=$(grep -m1 -E '^[[:space:]]*s156_win=\$\(awk ' "$S156O_SRC")
+  case "$s156o_line" in
+    *"awk '"*) s156o_prog=${s156o_line#*awk \'}; s156o_prog=${s156o_prog%\'*} ;;
+  esac
+fi
+
+# s156o_accepts <non-blank-line-count> — 0 when the guard's DECLARED bounds admit a
+# window of that size. Mirrors §156a's decision; an undeclared bound is inert.
+s156o_accepts() {
+  local n="$1"
+  [ -n "$s156o_floor" ] && [ "$n" -lt "$s156o_floor" ] && return 1
+  [ -n "$s156o_ceil" ] && [ "$n" -gt "$s156o_ceil" ] && return 1
+  return 0
+}
+
+# Fixtures. -1 means NOT MEASURED, so no arm can read a zero as agreement.
+S156O_DIR="$TMP/s156o"
+s156o_hn=-1; s156o_on=-1; s156o_tn=-1
+s156o_hfence=-1; s156o_ofence=-1
+if [ -f "$S156_SPEC" ] && [ -n "$s156o_prog" ] && mkdir -p "$S156O_DIR" 2>/dev/null; then
+  if cp "$S156_SPEC" "$S156O_DIR/healthy.md" 2>/dev/null && [ -s "$S156O_DIR/healthy.md" ]; then
+    # (b) over-extension: ONE unclosed fence opened immediately after the §1.3.1
+    # heading — the minimum mutation that desyncs the toggle. (c) truncation: the
+    # heading renamed, so the window never opens.
+    awk '{print} /^#### 1\.3\.1 /{print "```"}' "$S156O_DIR/healthy.md" > "$S156O_DIR/overextended.md" 2>/dev/null
+    awk '/^#### 1\.3\.1 /{sub(/^#### 1\.3\.1 /, "#### 1.3.1-renamed ")} {print}' \
+      "$S156O_DIR/healthy.md" > "$S156O_DIR/renamed.md" 2>/dev/null
+    s156o_hfence=$(grep -c '^```' "$S156O_DIR/healthy.md")
+    [ -s "$S156O_DIR/overextended.md" ] && s156o_ofence=$(grep -c '^```' "$S156O_DIR/overextended.md")
+    s156o_hn=$(awk "$s156o_prog" "$S156O_DIR/healthy.md" | grep -c .)
+    [ -s "$S156O_DIR/overextended.md" ] && s156o_on=$(awk "$s156o_prog" "$S156O_DIR/overextended.md" | grep -c .)
+    # The renamed fixture is validated by its MARKER, not by its count: a count of 0
+    # is the very thing §156q asserts, so reading 0 as proof the fixture built would
+    # be circular. Only if the rename actually landed is the count trusted.
+    if [ -s "$S156O_DIR/renamed.md" ] && grep -q '^#### 1\.3\.1-renamed ' "$S156O_DIR/renamed.md" \
+       && ! grep -q '^#### 1\.3\.1 ' "$S156O_DIR/renamed.md"; then
+      s156o_tn=$(awk "$s156o_prog" "$S156O_DIR/renamed.md" | grep -c .)
+    fi
+  fi
+fi
+
+# §156o — AC1: the guard names BOTH bounds in its own assertion message, so a failure
+# says WHICH SIDE tripped. Reads the guard's text; §156p reads its behaviour.
+if [ -z "$s156o_msgs" ]; then
+  ng "156o: cannot read this suite's own 156a assertion messages — the window-bound arms below would be vacuous (#643)"
+elif [ -n "$s156o_floor" ] && [ -n "$s156o_ceil" ]; then
+  ok "156o: the §1.3.1 window guard names both bounds in its assertion message (floor=$s156o_floor ceiling=$s156o_ceil) — a failure names which side tripped (#643)"
+else
+  ng "156o: the §1.3.1 window guard declares only one bound (floor=${s156o_floor:-<none>} ceiling=${s156o_ceil:-<none>}) — a failure cannot name which side tripped (#643)"
+fi
+
+# §156p — AC2, THE WITNESS. One unclosed fence inside §1.3.1 must make the guard fail
+# LOUD. RED at this commit: the floor admits the over-extended window.
+if [ "$s156o_hn" -lt 0 ] || [ "$s156o_on" -lt 0 ]; then
+  ng "156p: over-extension fixture not built (healthy n=$s156o_hn over n=$s156o_on) — the ceiling is UNTESTED, not satisfied (#643)"
+elif [ "$s156o_ofence" -ne $((s156o_hfence + 1)) ]; then
+  ng "156p: over-extension fixture did not take (fence markers healthy=$s156o_hfence over=$s156o_ofence, expected +1) — arm would be vacuous (#643)"
+elif [ "$s156o_on" -le "$s156o_hn" ]; then
+  ng "156p: over-extension fixture did not over-extend (over n=$s156o_on vs healthy n=$s156o_hn) — arm would be vacuous (#643)"
+elif s156o_accepts "$s156o_on"; then
+  ng "156p: an unclosed fence in §1.3.1 runs the window to $s156o_on non-blank lines (healthy=$s156o_hn) and the declared bounds ACCEPT it (floor=${s156o_floor:-<none>} ceiling=${s156o_ceil:-<none>}) — the content locks would be called load-bearing over §1.4–§3 (#643)"
+else
+  ok "156p: an unclosed fence in §1.3.1 over-extends the window to $s156o_on non-blank lines and the declared bounds REJECT it (floor=${s156o_floor:-<none>} ceiling=${s156o_ceil:-<none>}) (#643)"
+fi
+
+# §156q — AC3, a BOUND (not a witness; passes before and after). The ceiling must not
+# blind the floor: a renamed heading still collapses the window and still fails loud.
+if [ "$s156o_tn" -lt 0 ]; then
+  ng "156q: renamed-heading fixture not built or rename did not land — the floor is UNTESTED, not satisfied (#643)"
+elif [ "$s156o_tn" -ne 0 ]; then
+  ng "156q: renamed-heading fixture left a non-empty window (n=$s156o_tn, expected 0) — arm would be vacuous (#643)"
+elif s156o_accepts "$s156o_tn"; then
+  ng "156q: a renamed §1.3.1 heading collapses the window to 0 non-blank lines and the declared bounds ACCEPT it (floor=${s156o_floor:-<none>} ceiling=${s156o_ceil:-<none>}) (#643)"
+else
+  ok "156q: a renamed §1.3.1 heading collapses the window to 0 non-blank lines and the declared bounds REJECT it — the ceiling does not blind the floor (#643)"
+fi
+
+# §156r — AC4, a BOUND (not a witness; passes before and after). The two rejections
+# above are not an always-fail: the unmodified SPEC still passes. The count equality
+# is also the anti-drift check — a fixture path that no longer reproduces the live
+# window fails loud instead of quietly measuring something else.
+if [ ! -f "$S156_SPEC" ]; then
+  ng "156r: SPEC.md absent — the healthy-window bound is UNTESTED, not satisfied (#643)"
+elif [ "$s156o_hn" -lt 0 ]; then
+  ng "156r: healthy fixture not built or the guard's extractor could not be lifted from its source — bound UNTESTED (#643)"
+elif [ "$s156o_hn" -ne "$s156_n" ]; then
+  ng "156r: healthy fixture window ($s156o_hn) disagrees with the live §156a window ($s156_n) — the fixture path no longer reproduces the guard (#643)"
+elif s156o_accepts "$s156o_hn"; then
+  ok "156r: the unmodified SPEC §1.3.1 window ($s156o_hn non-blank lines) is ACCEPTED by both declared bounds (floor=${s156o_floor:-<none>} ceiling=${s156o_ceil:-<none>}) (#643)"
+else
+  ng "156r: the unmodified SPEC §1.3.1 window ($s156o_hn non-blank lines) is REJECTED by the declared bounds (floor=${s156o_floor:-<none>} ceiling=${s156o_ceil:-<none>}) — the bounds are always-failing (#643)"
+fi
+
 # §156j — POSITIVE parity assertion: the §1.8 lever row and the §1.9 posture row
 # added for the SSOT change sweep are shaped so §116's OWN derivations count them, and
 # §116's parity therefore still balances. The awk/grep expressions below are
--- a/scripts/test/smoke.d/70-gates-contentlocks.sh
+++ b/scripts/test/smoke.d/70-gates-contentlocks.sh
@@ -2333,12 +2333,28 @@ s156_fx() { printf '%s\n' "$s156_win" | grep -qF "$1"; }
 
 if [ ! -f "$S156_SPEC" ]; then
   ng "156: SPEC.md absent — cannot assert the §1.3.1 SSOT change sweep protocol (#640)"
-elif [ "$s156_n" -lt 18 ]; then
-  # 21 non-blank lines at this commit; a floor of 18 catches a collapsed/renamed
-  # heading or a gutted section while tolerating ordinary prose tightening.
-  ng "156a: SPEC §1.3.1 window empty or collapsed (non-blank lines=$s156_n, floor=18) — the content locks below would pass VACUOUSLY (#640)"
+elif [ "$s156_n" -lt 18 ] || [ "$s156_n" -gt 60 ]; then
+  # TWO-SIDED, because the extractor is fence-aware and that INVERTED the failure
+  # mode rather than removing it (#643). The window closes on a heading only while
+  # the fence toggle `f` is 0, so ONE unbalanced ``` inside §1.3.1 desyncs the
+  # toggle, no heading ever terminates it, and it runs to EOF — measured, 368
+  # non-blank lines against a healthy 22. Before fence-awareness an unbalanced
+  # fence TRUNCATED and the floor caught it loud; after it, the window
+  # over-extends and a floor alone reports the locks "load-bearing" over §1.4–§3.
+  #
+  # 22 non-blank lines at this commit — RE-MEASURED, not carried forward: this
+  # comment previously read 21, which was correct at f5af42c and went stale when
+  # that same commit added a prose line (#643 AC5). The count is deliberately NOT
+  # pinned by an arm, per §156j's neighbouring note.
+  #
+  # floor=18 tolerates ordinary prose tightening and catches a collapsed or
+  # renamed heading; ceiling=60 leaves §1.3.1 room to nearly triple while sitting
+  # six-fold below the ~368 a desync produces. The bounds are NAMED in both
+  # messages so a failure says which side tripped — and §156o/§156p parse those
+  # two spellings, so renaming them silently unbinds the arms that check them.
+  ng "156a: SPEC §1.3.1 window out of bounds (non-blank lines=$s156_n, floor=18, ceiling=60) — below the floor the content locks would pass VACUOUSLY; above the ceiling the window has over-extended past §1.3.1 and the locks would be asserted over unrelated sections (#640, #643)"
 else
-  ok "156a: SPEC §1.3.1 window resolves non-empty (non-blank lines=$s156_n) — content locks below are load-bearing (#640)"
+  ok "156a: SPEC §1.3.1 window resolves within bounds (non-blank lines=$s156_n, floor=18, ceiling=60) — content locks below are load-bearing (#640, #643)"
 
   # §156b: the two commit-trailer KEYS, anchored to the code form (line-start inside the
   # fenced grammar block), not a backticked prose mention — the later Execution Issues
@@ -2424,7 +2440,12 @@ else
   # §156m: the MONOTONICITY DIRECTION — widen-yes / narrow-never — plus the re-anchor
   # carve-out the parent Directive states. §156d locks the `inconclusive` machinery;
   # this locks the rule that machinery guards.
-  if s156_re 'widened' && s156_re 'never be narrowed' && s156_re 're-anchor'; then
+  # PHRASES, not bare words (#643 AC6). `widened` and `re-anchor` were each
+  # individually satisfiable by unrelated prose anywhere in the window; only the
+  # conjunction carried the lock. Pinned to the forms §1.3.1 actually uses, so
+  # each conjunct now bites on its own.
+  if s156_re 'may be \*\*widened\*\*' && s156_re 'never be narrowed' \
+     && s156_re '\*\*re-anchor\*\* accompanying a heading rename'; then
     ok "156m: §1.3.1 keeps widen-yes / narrow-never with the re-anchor carve-out (#640)"
   else
     ng "156m: §1.3.1 lost the monotonicity direction or the re-anchor carve-out (#640)"
--- a/scripts/test/smoke.d/70-gates-contentlocks.sh
+++ b/scripts/test/smoke.d/70-gates-contentlocks.sh
@@ -2342,9 +2342,8 @@ elif [ "$s156_n" -lt 18 ] || [ "$s156_n" -gt 60 ]; then
   # fixture (SPEC.md plus one injected marker, 2895 lines): the window spans
   # 269-739 and exits at 740 on a shell comment inside a ```bash fence — 368
   # non-blank lines against a healthy 22, stable across six injection offsets.
-  # Before fence-awareness an unbalanced
-  # fence TRUNCATED and the floor caught it loud; after it, the window
-  # over-extends and a floor alone reports the locks "load-bearing" over §1.4–§3.
+  # When the window over-extends like that, a floor alone reports the locks
+  # "load-bearing" over §1.4–§3.
   #
   # 22 non-blank lines at this commit — RE-MEASURED, not carried forward from the
   # stale 21 this comment used to carry (#643 AC5). No history is reconstructed
@@ -2358,8 +2357,7 @@ elif [ "$s156_n" -lt 18 ] || [ "$s156_n" -gt 60 ]; then
   # six-fold below the ~368 a desync produces today. The bounds are NAMED in both
   # messages so a failure says which side tripped, and §156o/§156p parse those two
   # spellings — renaming them fails CLOSED (measured: both arms red), so it is not
-  # a silent hazard. The mechanism is stated once, at §156o's accept-check below;
-  # it is deliberately not restated here.
+  # a silent hazard.
   ng "156a: SPEC §1.3.1 window out of bounds (non-blank lines=$s156_n, floor=18, ceiling=60) — below the floor the content locks would pass VACUOUSLY; above the ceiling the window has over-extended past §1.3.1 and the locks would be asserted over unrelated sections (#640, #643)"
 else
   ok "156a: SPEC §1.3.1 window resolves within bounds (non-blank lines=$s156_n, floor=18, ceiling=60) — content locks below are load-bearing (#640, #643)"
@@ -2470,7 +2468,7 @@ else
 fi
 
 # ---------- §156o–§156r: the §1.3.1 window is bounded on BOTH sides (#643) ----------
-# The count-guard above is a FLOOR only. The fence-awareness that closed the
+# The count-guard above was a FLOOR only at e8aa0db. The fence-awareness that closed the
 # truncation gap (#641) opened the opposite one: with an ODD number of ``` markers
 # inside §1.3.1 the extra marker INVERTS the `f` parity from that point on, so headings in real prose
 # stop terminating the window and it runs on — measured, to line 740 of the 2895-line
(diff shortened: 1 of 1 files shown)


ilgyu-yi:
## Finding triage (round 1)

`reviewed-head: 3a6297ae5eaf7373ea67d3c2d8fd4d96b03b388c` · `prior-round: none (first round)` · `swing: not-evaluated` on every fix-now.

**Dispatch.** One worktree-isolated `code-reviewer` (`ship after fix`), head resolved independently and matched — **valid**. No security tier: single test file, no enforcement code, no ADR. Judged by a fresh worktree-isolated `finding-judge` driving the §156 arms through a faithful **script-file** driver that sources the real `_preamble.sh` and the real section file.

**9 findings. 8 confirmed, 1 REFUTED. fix-now ×5, defer ×2, drop ×2.**

The four must-fix findings are all false claims **I wrote**, in a PR whose `Docs: n/a` explicitly elevates those comments to documentation.

### F1 — fix-now. The stated mechanism is false

I wrote that the toggle *"sticks"*, *"no heading ever terminates it"*, and the window *"runs to EOF"*. Measured: the window spans SPEC 269–739 and **exits at NR=740 of 2895**, on a `#`-prefixed shell comment inside a genuine ` ```bash ` fence. The toggle does **not** stick and a heading terminator **does** fire.

The real mechanism is **parity inversion**: from the injected fence onward every region's fence-state is flipped, so headings in real prose are ignored while `#`-lines inside real fences are honoured. Stable at 368 across **six** injection offsets.

**Floor**: remove the false clause at all **three** mutable occurrences (the §156a comment, the §156o header, the PR body) — a correction landing in one leaves it live in the others. **Ceiling**: do **not** touch the ceiling value, the instrument, or the adjacent clauses — every one was independently re-measured **true** (368 vs 22; the window really spans §1.4–§3.2; the six-fold margin). Replacing verified-true text with new unverified text is the exact generator #643 names.

### F2 — fix-now, and the remedy is a **scoping deletion**, not a rewrite

I wrote that the comment *"read 21, which was correct at f5af42c"*. Measured: at `f5af42c` it read **18**; `21` was introduced by `4c23a93`, where the true count was **already 22**. `21` was stale on arrival. The sentence is also internally incoherent — 21 both becomes correct and goes stale at one commit.

**`remedy-survives: no`.** The judge verified the reviewer's replacement narrative is *accurate* and still ruled it does not survive: AC5 requires only that the **count** be re-measured; the narrative carries **zero** enforcement value (§156j doctrine leaves provenance counts unpinned, so nothing can ever catch it drifting); and this exact line has a measured **2-of-2 false-history rate**. Writing a third historical narrative onto the line whose only defect *is* historical narrative re-arms the generator. **Floor**: keep the count and its anchor. **Ceiling**: no reconstruction of prior values or commits at all.

### F3 — fix-now. The disclosed risk is **inverted**

I wrote that renaming the bound spellings *"silently unbinds the arms"*. Measured with `ceiling=` → `upper=`: **two arms red loud**. `s156o_accepts` returns non-zero on an empty bound, so the design **fails closed**. **Ceiling: no code change** — a risk measurement shows does not exist needs no mitigation, only its false disclosure deleted.

### F4a — fix-now. Scope two overbroad claims

*"The arms drive the LIVE guard, not a paraphrase"* is true of the **extractor** and false of the **decision**, which re-implements the guard's condition from integers scraped out of message prose. And the WITNESS-HONESTY defence covers only the **omission** direction, not **divergence**. Scoping the first inherently discloses that the decision is a re-implementation. **Ceiling**: no change to the guard condition or the four arms.

### F4b — defer-to-issue. A real, measured, undisclosed vacuity vector

Delete the ceiling from the **executed condition only**, leave the messages untouched, and **all four arms stay green** while the guard carries no effective ceiling. That is the vacuous-guard anti-pattern §156a exists to convert.

**Deferred on scope, not doubt.** The proposed hoist touches the executed condition, both messages, and all four arms — and `_preamble.sh` sets `set -uo pipefail`, so arms reading `$S156_CEIL` become an **unbound-variable abort** against the pre-fix guard at `e8aa0db`. The witness reproduction would degrade from a clean `§156p` RED to a **crash**, and that RED is this PR's entire evidence the defect was repaired. Trading established witness evidence for a hardening the ACs do not require is the wrong order. The F4 mutant goes into the issue verbatim so the next round starts from the reproduction. ADV3/ADV4/ADV6 (the two idiom notes and the `head -1` AC1 nit) fold into the same issue — same arm-touch.

### ADV1 — **REFUTED**, and this one matters most

The reviewer said my MISSION citation was wrong: *"The mechanism" is about context narrowing and says nothing about audit-and-gate machinery.* Measured, the section spans **10–22**, and **line 18** is an entire paragraph on enforcement — *"negative enforcement refuses a wrong action … Which face a rule wears is decided by the asymmetry in the cost of being wrong."* The reviewer read 10–16 and stopped immediately before it.

Acting on this would have replaced a **correct** citation with a worse one — a fourth false claim, authored in direct response to a review finding, arriving through the review channel itself. Retained as `refuted`, excluded from the fix list, and **not** convertible to an issue; next round it reads back as `already-refuted (round 1)`.

### ADV2 — fix-now. One line

Ready-not-draft is disclosed nowhere, while `Docs: n/a` and `skip-changelog` both are — an internal inconsistency in the same document, and the third instance of this pattern in the drive. **Ceiling**: one line; specifically **not** converting back to draft, which would reset the review gate.

### Dropped (2)

**ADV5** — `S156O_SRC` hardcodes the section path. Measured on a relocated tree: the arm correctly read the copy's own file. Only a rename is exposed, and that fails closed. **ADV1** above.

### Constraints on the remedy

**The commit bodies are out of reach.** The false F1 text sits in **both** pushed commit bodies and the false F2 text in `3a6297a`. Amend-after-push is blocked and rewriting reviewed history is not on the table — so the remedy covers **mutable surfaces only**, with a forward correction in the PR body naming the two commit bodies as carrying the superseded claims.

**One figure the author must not copy forward.** The reviewer's F1 text says *"84% of the file is never reached."* Measured, it is **74.5%** (2156 of 2895 unreached). Copying it would make the repair of a false claim introduce the next one — the precise failure mode this finding set is about.

### Observation, not a judged record

F1's true mechanism means the over-extension magnitude depends on where the first `#`-in-fence line sits downstream, so a desync smaller than the ceiling would pass unbounded. That is inherent to **any** consequence bound and is what #643's Out of scope already settled ("bound the consequence, not parse the cause"). Measured stable at 368 across six placements, six-fold above the ceiling. Recorded so it is not rediscovered as a finding — it is **not** grounds to revisit the ceiling here.

ilgyu-yi:
## Finding triage (round 2)

`reviewed-head: 1ada0a0a1c3e25dec1acd1a365dae0fccd7d2fd0` · `prior-round: 1@3a6297a` · `swing: none` where evaluated.

**9 findings. 9 confirmed. fix-now ×6, drop ×3.** `code-reviewer`'s `ship after fix` is unchanged.

### R2-F1 — the remedy is a **deletion**, not a third correction

My F3 fix wrote *"fails CLOSED (measured: both arms red, **because the arms' accept-check rejects an empty bound**)"*. The **outcome** is true; the **because** is backwards. `s156o_accepts` short-circuits on an empty ceiling (`[ -n "$s156o_ceil" ] && …`) and falls to `return 0` — it **accepts**. The two arms red for two *different* reasons, neither of them the one written: `§156o` explicitly tests `-n "$s156o_ceil"`, and `§156p` treats an **accept** of a 368-line window as failure.

**The correct mechanism is already stated correctly twice in this same file** — *"an undeclared ceiling reads here exactly as it behaves there: **absent**"* and *"an undeclared bound is **inert**"*. Restating it at `:2359` duplicates already-correct text, and duplication across surfaces is the measured generator of R2-F2 and R2-F3 **in this very round**.

**Floor**: keep the measured outcome. **Ceiling**: add **no** mechanism statement. This clause has a **2-of-2 measured false rate** (round 1: "silently unbinds"; round 2: "accept-check rejects") and zero enforcement value. Same clause in the PR body's Risks bullet — same fix item.

### R2-F2 — round 1's F2 correction never reached the PR body

The Test-plan AC5 paragraph still carries the refuted narrative verbatim, **contradicting this same body's Round-1 review section eleven lines above**. Round-1 F1's floor named the PR body as one of three mutable surfaces precisely to prevent this. Same scoping deletion: keep the re-measured **22** and its anchor, reconstruct no history.

### R2-F3 — two stale RED claims, and here the remedy is a **correction**

`:2481` *"RED at this commit"* and `:2487` *"§156o is red today"* — both arms are **GREEN** at head. True when written at `e8aa0db`, false as shipped, and the round-2 commit edited the text immediately **above and below** them without touching them.

The judge explicitly **declined to swing this toward deletion**, on a principled discriminator (below). The replacement was verified, not passed through: `git show e8aa0db` confirms the guard was floor-only there.

### Idiom notes — 3 fix-now, 3 drop

**IN-1** (fix): the round-2 rewrite left *"with an ODD number of ``` markers inside §1.3.1 INVERTS the `f` parity"* — the verb has no subject. Restoring it asserts nothing new. **Ceiling: restore the subject only; do not restate the mechanism.**

**IN-2** (fix): `"740 of 2895"` / `"SPEC 269-739"` are the **mutated fixture's** coordinates while real `SPEC.md` is **2894** lines (the builder injects exactly one line). **The numbers are verified correct and must not change** — only the artifact they are attributed to is wrong. Same mislabel in the PR body.

**IN-3** (fix): the PR body's `## Round-1 review (1ada0a0)` header names the *correction* commit; round 1 reviewed `3a6297a`. One-token substitution read straight off the canonical marker.

**Dropped**: the explicit-discriminator note (**duplicate of round-1 F4b**, already deferred and correctly disclosed in-file); a three-process pipeline that could be one `sed` (style, and it touches working extraction code during a false-claim repair round); and ragged rewrap (cosmetic — and rewrapping verified-true text is exactly round-1 F1's stated ceiling violation; permitted only *incidentally* where a block is already being rewritten).

### Converging — and the pattern argues for stopping the **prose**, not the PR

**The enforcement surface is done and measured correct.** Two rounds produced **zero** code defects. The guard is two-sided, all four arms green, `§156p` rejects the 368-line over-extension, `§156q` shows the ceiling has not blinded the floor, `§156r` shows the pair is not always-failing. #643's actual deliverable has been stable since `3a6297a`.

**Every defect in both rounds is explanatory prose about that correct code**, and round 2 resolved the generator into two mechanisms:

- **(a) Added mechanism clauses fail.** Of the five edit clusters in `1ada0a0`, the four that **deleted or scoped** a claim are verified **true**; the one that **added an explanatory `because`** is verified **false**. **4-for-4 versus 0-for-1.**
- **(b) Corrections do not reach every surface.** Round 2 reproduced this twice — the F2 fix landed in the file but not the body, and the commit edited above and below the stale RED lines without touching them.

**The discriminator to carry forward:** *keep a claim only if it is verifiable against a cheap, stable oracle.* A fixed-commit fact (`git show e8aa0db`) is one command away and can never go stale — **keep it**. A mechanism explanation needed a mutated-fixture run and has been wrong twice — **delete it**, especially when a correct twin already exists 140 lines below. A provenance narrative needed archaeology across three commits and was wrong twice — **delete it**.

That doctrine was *already in the file* at `:2354` — *"Re-measure it; do not re-tell it"* — and `:2359` violated it **four lines later**.

**Stopping criterion, stated so round 3 is decidable:** if round 3 finds a new false claim in text that **round 2 added** under a deliberately subtractive ruling, the generator has survived a targeted counter and the remaining prose should be stripped to assertions only. We are not there — round 2's additions were unverified; round 3's are verified before landing.

**Counter to (b), adopted:** an **enumerated surface sweep**. R2-F1, R2-F2 and IN-2 each span two surfaces; no item is done until every surface carrying a tense, history, or mechanism claim about this guard has been checked — not just the one the finding cited.

### Observation, not a judged record

Round 1's `defer-to-issue` on the F4b divergence vector **was never executed** — no Issue exists, and the head's PR body correctly says *"No Issue tracks it yet."* Nothing is misstated, but the ruling's durable half is outstanding, and the three folded idiom notes ride with it.

ilgyu-yi:
## Finding triage (round 3)

`dispatch:` findings from a worktree-isolated `code-reviewer` (round 3); judgment from a fresh, worktree-isolated `finding-judge` with no shared context. Limb (a) is not self-observable — recorded as given, not asserted.

`reviewed-head: f52ded4d20b2cb250e2df2cd16c690ded60d0d57` · `prior-round: 2@1ada0a0`

**Artifact pin.** Head resolved independently. The judge's ambient worktree sits at the base `18dec09`, so every read was `git show f52ded4:<path>` and every run was in a scratch clone at `f52ded4`. `shasum -a 256` of `scripts/test/smoke.d/70-gates-contentlocks.sh` = `f6d826b251f27c76b4fe2c6f7e2ac7f38f0eaac5143fd53eb377f9aa90e4770e` — **identical to the reviewer's pin**, unchanged after all runs.

**5 findings judged. 5 confirmed, 0 refuted. fix-now ×3, defer ×1, drop ×1.** `code-reviewer`'s `ship after fix` is untouched — the judge converts no verdict.

Every reviewer command was re-run rather than trusted. **All of the reviewer's measurements reproduced exactly** — the artifact hash, `pass=1297 fail=0`, the `e8aa0db` positive control, and all three rename variants. This is the first round in which the reviewer's mechanism claims were right on every count.

---

### F1 — the truncation claim is false on both halves · **fix-now**

```
finding: F1 — "Before fence-awareness an unbalanced fence TRUNCATED and the floor caught it loud" is false in both halves
site: scripts/test/smoke.d/70-gates-contentlocks.sh#§156a-comment
axis-key: …#§156a-comment::history-claim
verdict: confirmed
evidence: executed=awk '/^#### 1\.3\.1 /{i=1;next} i&&/^(#|##|###|####) /{exit} i' over.md | /usr/bin/grep -c .
          result=23 vs healthy 22 — the fence-blind extractor GREW the window by the injected
          fence line; it did not truncate
evidence: executed=git show f5af42c:SPEC.md | awk 'NR==292{print "# …"} {print}' | awk '<fence-blind extractor>' | /usr/bin/grep -c .
          result=21 -> 13; floor at f5af42c measured -lt 10, so 13 >= 10 was ACCEPTED SILENTLY
evidence: executed=git show 4c23a93:… | /usr/bin/grep -n 's156_n" -lt'
          result=`-lt 18` — the 10->18 raise and fence-awareness landed in the SAME commit, so the
          floor was never loud "before fence-awareness"
action: fix-now
remedy-survives: yes
axis: fewer<->more   target-position: interior
prior-round: 2@1ada0a0   swing: none
```

**FLOOR** — the clause is deleted on **both** mutable surfaces (`:2345-2347` and the PR body's Key context). A deletion landing on one surface leaves it live on the other: generator (b), named in round 2 and reproduced twice.
**CEILING** — that clause only. The adjacent measured-true text stays (368-vs-22, the 269–739 span, exit at 740, the six-fold margin), and the true truncation history already sits at `:2321-2322` — re-measured accurate (21 → 13 at injection offset 292, matching the in-file figure exactly). **Author no replacement narrative**: this line already carries a 2-of-2 measured false-history rate, and a third narrative re-arms the generator.

The clause is doubly wrong — it misattributes truncation to an *unbalanced* fence when the real historical cause was a `# ` comment inside a **balanced** fence, and it credits a floor that measurably did not fire.

### F2 — stale present tense, inside the block round 2 swept · **fix-now**

```
finding: F2 — ":2473 `The count-guard above is a FLOOR only.` is present-tense and false at head"
site: scripts/test/smoke.d/70-gates-contentlocks.sh#§156o-witness-header
axis-key: …#§156o-witness-header::tense-anchor
verdict: confirmed
evidence: executed=/usr/bin/grep -n -m1 'elif \[ "\$s156_n"' scripts/test/smoke.d/70-gates-contentlocks.sh
          result=2336: `elif [ "$s156_n" -lt 18 ] || [ "$s156_n" -gt 60 ]` — two-sided at head
evidence: executed=git show e8aa0db:… | /usr/bin/grep -n 's156_n" -lt'
          result=2336: `elif [ "$s156_n" -lt 18 ]` — floor-only, so the proposed anchor is true
action: fix-now
remedy-survives: yes
axis: implicit<->explicit   target-position: high
prior-round: 2@1ada0a0   swing: none
```

Independent ground, not directional. (i) The author **adopted this exact convention** at `:2484` and `:2491` in this very commit ("RED at this commit" → "RED at `e8aa0db`") and missed only the block's opening sentence — this completes an adopted standard rather than choosing a new position. (ii) A fixed-commit anchor is round 2's measured keep-class: one `git show` verifies it and it can never go stale. The replacement was verified, not passed through.
**FLOOR** — the sentence must stop asserting a present-tense floor-only guard. **CEILING** — one token (`is a FLOOR only` → `was a FLOOR only at e8aa0db`). Do **not** delete the sentence (it sets up the paragraph; orphaning costs more than it saves), do not rewrap, do not touch §157's parallel claim.

Not a swing: round 2 faced this class (R2-F3) and **explicitly declined to swing it toward deletion** on a stated discriminator. F2 occupies the position round 2 chose.

### A2 — **upgraded** from the reviewer's "no edit". The round's most important finding. · **fix-now**

```
finding: A2 — ":2361 `The mechanism is stated once, at §156o's accept-check below`" — the count is false
site: scripts/test/smoke.d/70-gates-contentlocks.sh#§156a-comment
axis-key: …#§156a-comment::count-claim
verdict: confirmed
evidence: executed=/usr/bin/grep -n -E 'undeclared (bound|ceiling)|cannot bind anything|is inert' <file>
          result=TWO sites — :2505-2506 and :2524. Count is 2, not 1.
evidence: executed=git show 1ada0a0:… | grep 'stated once' ; git show 3a6297a:… | grep 'stated once'
          result=absent at both — the sentence is NEW IN f52ded4
evidence: executed=gh pr view 672 --json body -q .body | /usr/bin/grep -n -i 'stated once\|twice'
          result=line 47 "stated once in the file"; line 65 "stated correctly **twice** in the same
          file" — the body contradicts itself 18 lines apart
action: fix-now
remedy-survives: no
axis: fewer<->more   target-position: low
prior-round: 2@1ada0a0   swing: none
```

Independent ground: the sentence is a meta-claim about the file's own prose layout with **zero enforcement value**, its count is false under every reading, and a hand-synced pointer-plus-count has nothing that can ever catch it drifting — the identical `§156j` argument round 1 used to rule R1-F2 a scoping deletion.
**FLOOR** — the false count goes from **both** surfaces (`:2361` and PR body line 47). **CEILING** — that sentence/parenthetical only. Do **not** delete the two correct mechanism statements at `:2505` and `:2524`, and specifically do **not** "fix" the count to "twice": substituting a corrected count re-arms the exact generator, since the next edit that adds or removes a mechanism sentence makes it false again. **Deletion is the unique edit here that cannot go stale.**

`remedy-survives: no` because the reviewer's remedy was *no edit*, and it does not survive. A2 is in the reviewer's own set — this re-disposes and bounds it, it is not a new finding.

### A1 — **drop**. The reviewer's no-edit call is right, checked before agreeing.

```
finding: A1 — ":2360 renaming them fails CLOSED (measured: both arms red)" is over-general
verdict: confirmed
evidence: ceiling= -> upper=  : ✗156o ✗156p          pass=1295 fail=2
evidence: floor=   -> lower=  : ✗156o ✓156p ✗156q    pass=1295 fail=2
evidence: both renamed        : ✗156o ✗156p ✗156q    pass=1294 fail=3
          — under the PLURAL reading the sentence actually invites ("renaming THEM", both
          spellings), "both arms red" is TRUE
action: drop
```

The sentence says *"renaming **them**"* — plural, both spellings — and under that reading it measured **true**. It is false only under a floor-only single-spelling rename, which the sentence does not claim. The load-bearing half — *fails CLOSED, not a silent hazard* — held in **all three** variants (`§156o` red every time). A fourth pass over a clause whose claim is true as written would be the generator, not a repair.

### A3 — **defer-to-issue**, and the round-1 deferral is still unexecuted

```
finding: A3 — s156o_accepts re-derives the guard's decision from integers scraped out of assertion
        prose (the F4b divergence vector)
verdict: confirmed
evidence: executed=perl -i -pe 's/ \|\| \[ "\$s156_n" -gt 60 \]// if $.==2336' … ; smoke.sh
          result=✓156o ✓156p ✓156q ✓156r, pass=1297 fail=0 — the guard silently loses its ceiling
          and the suite reports fully green
action: defer-to-issue
```

The vector reproduces exactly. Deferral stands on round 1's reasoning (the hoist turns `§156p`'s clean RED at `e8aa0db` into an unbound-variable crash under `set -u`, trading this PR's only repair evidence for hardening the ACs do not require). **The in-file disclosure at `:2497-2504` is measurably TRUE and honest** — the file discloses the exact hole reproduced here.

**Carry-over, third round running:** round 1 ruled `defer-to-issue`; **no Issue exists**. Round 2 recorded it and it was still not executed. A deferral whose durable half never lands is indistinguishable from a drop. This is now the outstanding item most likely to be lost.

---

## Round 2's stopping criterion — **TRIGGERED** (the judge disagrees with the reviewer here)

The reviewer checked the criterion against F1 and F2 only, found both to be `3a6297a`/`e8aa0db`-era text, and concluded *not triggered*. That check was **incomplete — it was never applied to the reviewer's own A2.**

The criterion reads *"a new false claim in text that round 2 added under a deliberately subtractive ruling."* `1ada0a0` predates round 2's ruling, so it cannot be text added *under* it; the only coherent referent is `f52ded4`, and round 2's closing sentence confirms it (*"round 3's are verified before landing"*).

Measured: **"stated once" is new in `f52ded4`, and false.** Round 2's R2-F1 ruled *"add NO mechanism statement"* and said in the same breath that the mechanism *"is already stated correctly **twice** in this same file"* — quoting both sites. The repair then wrote **"once"**, on two surfaces, contradicting the judged list that authorized the edit and contradicting its own PR body 18 lines below. One `grep` falsifies it.

So the generator survived the targeted counter, and the cheap-stable-oracle discriminator did **not** prevent it. The real lesson: the discriminator selects *which claims are keepable*, but nothing forces the check to actually run.

### What triggering does NOT license — the swing refused

Round 2's stated consequence was *"the remaining prose should be stripped to assertions only."* That consequence is ruled **`unjustified — needs measurement` and declined**, under anti-swing rule 1.

Its only available ground is *"prose repair failed a third time, so go to the far end"* — the direction of the previous failure, which is not a position. Round 2 sat deliberately at `interior` on `fewer↔more`: it preserved verified-true fixed-commit prose and explicitly declined deletion for the tense class. Stripping to assertions-only is the `low` extreme and an **opposite-end move from round 2's own position**, justified by nothing but recurrence.

It is also self-defeating on the evidence. A mass strip is a *large prose edit* — the operation carrying the measured ~1-new-defect-per-round rate — and it would delete text three independent rounds measured **true and load-bearing**: the WITNESS-HONESTY block, the F4b self-disclosure at `:2497-2504`, and the `:2321-2322` truncation history. Deleting honest self-disclosure to punish dishonest prose is a strictly worse artifact.

**The bounded position ruled instead, on independent ground:** all three `fix-now` remedies are **deletions or a one-token anchor onto an already-adopted convention**. Deletion has a property no other edit class has — *an assertion that is not written cannot be false* — and that is definitional, not directional. Hence `swing: none` on every item: nothing moves to the opposite end of round 2's axis.

**Operative rule for round 4, stated so it is decidable:** the remaining prose is not stripped. But **no clause in this block may be corrected a fourth time** — a clause measured false a third time is **deleted, never rewritten**. F1 (2-of-2 false) and A2 (false on arrival) both discharge by deletion under that rule already.

## Observations — not judged records

- **§157 carries the identical defect class**, pre-existing and correctly out of scope: `:2662` *"ROUTING arms (157v–157x) are LOAD-BEARING RED at this commit"* while `✓157v ✓157w ✓157x` all pass at head. Not reopened here. It does show the unanchored-tense-claim class is **systemic rather than local to §156** — worth carrying into the same issue as A3.
- **The executable guard is correct; no code defect found.** Independently reproduced: `pass=1297 fail=0`; window 22 within floor 18 / ceiling 60; ceiling bites (368, loud, and `§156p` fails *honestly* rather than false-greening); floor bites (0 on a renamed heading); positive control at `e8aa0db` gives `✗156o ✗156p ✓156q ✓156r`, so the witness genuinely discriminates; `lint OK: 79 files clean`. **Three rounds have now produced zero code defects.** Every defect in all three rounds is prose about correct code.
- `§148i-cli` did not red in the judge's scratch clone, consistent with #642 being an agent-worktree artifact rather than a real failure.

ilgyu-yi:
## Finding triage (round 4)

`dispatch:` findings from a worktree-isolated `code-reviewer` (round 4); judgment from a fresh, worktree-isolated `finding-judge` with no shared context. Limb (a) is not self-observable — recorded as given.

`reviewed-head: fb122838c53ffbc2c3947014b6c3f5cc465f8be1` · `prior-round: 3@f52ded4`

**Artifact pin.** Head resolved independently. All reads via `git show fb12283:<path>`, all runs in a fresh scratch clone at that sha. `shasum -a 256` of `scripts/test/smoke.d/70-gates-contentlocks.sh` = `caa17215db109fba821f80f7a87c5c6dc9df45b0b9046c863316dbd8e36a6a4a` — **identical to the reviewer's pin**. Prior judged lists read under the trusted-author filter: 3 comments, rounds 1/2/3, each with both the literal marker and a matching `^## Finding triage \(round N\)$` first line, no duplicates.

**1 finding judged. 1 confirmed, 0 refuted. fix-now ×1.** `code-reviewer`'s `ship after fix` is untouched.

---

### F1 — "two lines away" is a false distance claim · **fix-now**

```
finding: F1 — PR body §"Round 3": "the convention this PR had already adopted two lines away".
         The nearest adoption is 11 lines away; the other is 96.
site: PR#672-body#Round-3-section  (mirrored, unamendably, in fb12283's commit body)
axis-key: PR#672-body#Round-3-section::distance-claim
verdict: confirmed
evidence: executed=/usr/bin/grep -n 'e8aa0db' <git show f52ded4:…70-gates-contentlocks.sh>
          result=anchored sentence 2473 ("is a FLOOR only"); adoptions 2484, 2491, 2569.
                 2484-2473=11, 2569-2473=96. No adoption at distance 2.
evidence: executed=/usr/bin/grep -n 'e8aa0db' <git show fb12283:…>
          result=2471 / 2482 / 2489 / 2567 -> 11 and 96 at head too. False on both blobs.
evidence: executed=git diff 1ada0a0 f52ded4 -- …70-gates-contentlocks.sh
          result=alternate ("count") reading tested and AVAILABLE: exactly TWO lines carried the
                 literal "RED at this commit" and were converted (2481->2484, 2565->2569); a third
                 (2491) came from "red today". So "the RED-at-this-commit anchors" = 2, arithmetically.
evidence: executed=/usr/bin/grep -nE '[0-9]+ lines? (below|above|away)' <PR body> <commit body>
                   + measured each: 2345-2321=24 ; 2353->2357=4 ; 2501-2361=140
          result=IDIOM TEST, decisive. The author uses "<n> lines <direction>" as a LINE-DISTANCE
                 idiom four times across these two surfaces; the other three measure TRUE exactly.
                 The count reading requires breaking a 3-for-3 idiom exactly once, precisely where
                 it is false. Plain reading stands; the claim is wrong by 5.5x.
action: fix-now
remedy-survives: yes (the deletion half; the reviewer's illustrative replacement is bounded out)
axis: fewer<->more   target-position: low
prior-round: 3@f52ded4   swing: none
```

Deletion is definitionally stale-proof — *an assertion that is not written cannot be false* — the same definitional (not directional) ground round 3 used for A2. And measured: the corrected value (11) is a hand-synced pointer with no cheap stable oracle, which goes stale on the next line shift. That is the A2 generator verbatim.

**FLOOR** — the false distance stops being asserted on the one mutable surface. Minimum edit: remove the words `two lines away` from the PR body. The bare excision is grammatical and leaves a wholly true sentence ("two deletions and one token" verified; "the convention this PR had already adopted" verified — three sites gained `e8aa0db` in `f52ded4`).

**CEILING** — that phrase only. Specifically **do not**: substitute a corrected number (`11 lines away`) — the A2 generator verbatim; adopt the reviewer's illustrative `in the same block` — a new positional claim, unnecessary when excision is grammatical; edit the commit body (unamendable, and it carries the plural qualifier that makes it defensible); **touch `scripts/test/smoke.d/70-gates-contentlocks.sh` at all** — the hash must stay `caa17215…`, because a file edit moves the head and would invalidate a head-pinned review; renumber, rewrap, or restructure the surrounding narrative.

**Alternate-reading check, run as instructed — and unlike round 3's A1 it comes out against the artifact.** The rescue reading is real: the commit body's *"at the RED-at-this-commit anchors"* is plural, and exactly two lines bore that literal phrase. Under a scalar-distance reading that plural is even self-inconsistent (11 and 96 cannot both be "two"). Three things kill it: the author's own 3-for-3 verified distance idiom; the rescue qualifier is **absent from the PR body**, the only mutable surface; and "two lines" is a small exact number, not the round approximation the author reserves for large distances (`140`).

**Not a reversal of round 3's F2.** The `was a FLOOR only at e8aa0db` token stays at `head:2471` — verified true (`git show e8aa0db:…` → `elif [ "$s156_n" -lt 18 ]`, floor-only; behaviourally `✗156o ✗156p ✓156q ✓156r`, `pass=1295 fail=2`, `ceiling=<none>`). Only the narrative *about* F2 is edited.

---

## Round 3's operative rule — **NOT triggered**, and the ruling lands on its side anyway

Two independent reasons, both measured:

1. **Out of scope.** The rule reads *"no clause **in this block**…"* — "this block" is the §156 prose block in the guard file. This clause is PR/commit-body narrative, not in that block.
2. **Wrong count.** Even by analogy, the rule fires on a clause *measured false a third time*. This clause is **new in `fb12283`** and false for the **first** time.

Its prescription (*deleted, never rewritten*) is nonetheless exactly what is ordered here, so the outcome is rule-consistent either way.

## Is a further review round warranted? — **No.**

- **Zero code defects, four rounds running.** Independently reproduced at head: `pass=1297 fail=0`; window 22 within floor 18 / ceiling 60; over-extension fixture 368 non-blank, span 269–739, exit at 740, swallowed headings §1.4 (310) → §3.2 (733) — so the one piece of genuinely new prose round 3 wrote (*"…over §1.4–§3"*) is **exact**. `§148i-cli` did not red in the clone (consistent with #642).
- **Round 3's repair is provably inert.** `f52ded4..fb12283` is 4 insertions / 6 deletions, **all comment lines, zero executable lines**. `§156o` scrapes its source pattern-anchored (`^[[:space:]]*(ok|ng) "156a:` and `^[[:space:]]*s156_win=\$\(awk `), which a `#`-leading comment cannot match — comment edits cannot perturb it.
- **Both round-3 deletions landed on both surfaces, no corrected-count substitute.** `/usr/bin/grep -niE 'TRUNCATED and the floor|Before fence-awareness|stated once|floor caught it'` on the head blob → **no match**. The PR body's one hit is correct *history* of round 3's ruling, not a live claim.
- **The fix needs no round.** Confirmed: a PR-body edit does not move the head. `headRefOid` is the branch ref tip; `gh pr edit --body` writes the PR's `body` field, which is not a git object and creates no commit. The guard evidence, the artifact hash, and any head-pinned review all survive the edit untouched.

**Merge-gate state, surfaced so it is not mistaken as discharged.** `gh pr view 672 --json reviews` returns **empty** and `.claude/ghjig-state/file-review/` is **empty** — a head-pinned `/file-review` at `fb12283` is still an outstanding prerequisite to merge, independently of this finding. Ordering-free: because the body edit does not move the head, the review may be posted before or after it and stays valid at `fb12283`. The converse is the reason for the ceiling — "fixing" this by touching the file would move the head and invalidate that review.

## Observations — not judged records

- **The generator has migrated, and the object-level artifact has converged.** Rounds 1–3 found defects in the *file's* prose; round 4's sole defect is in the *narrative about round 3's fix*. The file itself is now clean on every claim measurable. Four consecutive rounds of prose-about-correct-code is the strongest available signal to stop.
- The reviewer's no-action advisories are concurred with. The `+1` placeholders are genuinely stale for `fb12283` and can ride the same body edit at no extra cost. The line-79 pointer `:2497-2504` drifted (the block is **2495-2504** at head) but still lands inside and names the right text — note it is a **line-number pointer**, the anchor class this role forbids; worth carrying into the same issue as A3, not fixing here.
- **Round 1's `defer-to-issue` on `s156o_accepts` (A3) is now unexecuted for a fourth round.** No Issue exists. Round 3 flagged it as "the outstanding item most likely to be lost"; it is still lost. This is the only durable carry-over on the PR.

ilgyu-yi (commented):
**Verdict: APPROVE** — `code-reviewer` returned `ship` at this head after a five-round review.

`reviewed-head` blind-compared against the privately-computed PR head; artifact pin `shasum -a 256 scripts/test/smoke.d/70-gates-contentlocks.sh = caa17215db109fba821f80f7a87c5c6dc9df45b0b9046c863316dbd8e36a6a4a`, re-verified byte-identical after every run.

**Summary of the finding record.** Five rounds produced **zero code defects**. Every defect found in rounds 1–4 was a false or stale *prose* claim about correct code, and in two cases the prior round's own repair became the next round's defect. Rounds 3 and 4 therefore ruled subtractively — delete unverifiable prose rather than correct it again — and round 4 recorded that the artifact had converged. All findings are discharged.

**Guard verified by mutation, each with a same-input positive control** (a fail-closed answer is the expected one on these arms, so a broken fixture would otherwise green them):

- unmutated: `smoke: pass=1297 fail=0`, `lint OK: 79 files clean`; window `n=22` within `floor=18, ceiling=60`
- ceiling bites: injected fence → `156a` red at `n=368`; the eleven content locks sit in the `else` arm and do not fake-pass, and `156p` fails **loud** on its own fixture-integrity guard rather than green-passing
- floor bites: renamed heading → `156a` red at `n=0`, also loud
- non-vacuity: ceiling removed honestly → `156o`/`156p` red, `pass=1295 fail=2`
- real pre-fix control at the Test commit `e8aa0db` → `156o` and `156p` red, `156q`/`156r` green, `pass=1295 fail=2` — byte-for-byte the RED the PR body claims

All six acceptance criteria of the linked issue measured met. Scope is one file; Doc → Test → Code order holds; `Docs: n/a` is disclosed with the `fix` relaxation stated; `skip-changelog` present and `fragment-gate` green; all six CI checks pass at this head.

Remaining items are adjudicated and deliberately not touched: the `s156o_accepts` divergence vector (deferred by four consecutive judged rulings, disclosed in-file), a pre-existing false tense claim in the neighbouring routing block (out of scope by recorded decision), and one drifted line-number pointer. All three are carried to a follow-up issue rather than repaired here — a fifth prose repair is the exact move the record warns against.

ilgyu-yi:
Filed as #673 — the `defer-to-issue` ruling from rounds 1–4 now has its destination. The reproduction was re-run independently on `main` at `469d9fb` before filing (mutant: all four §156o–§156r arms green at `pass=1297 fail=0` with the ceiling silently absent; control: `pass=1295 fail=2`), rather than inherited from the judged lists.

PR #674, opened 2026-08-17
fix(#673): decide from the executed condition, not the declared bounds

### LINKED ISSUE #673: §156o checks the window guard's DECLARED bounds, not its executed condition — deleting the ceiling leaves the suite green (and §157:2660's tense claim is stale)
## What

`§156o`'s `s156o_accepts` (`scripts/test/smoke.d/70-gates-contentlocks.sh:2521-2528`) is the arm that checks §156a's window guard declares **both** bounds. It drives the live guard's **extractor** verbatim — lifted out of the guard's own source line — but it **re-implements the guard's decision** from two integers scraped out of the `156a:` assertion messages:

```sh
s156o_floor=$(printf '%s\n' "$s156o_msgs" | grep -oE 'floor=[0-9]+'   | head -1 | cut -d= -f2)
s156o_ceil=$( printf '%s\n' "$s156o_msgs" | grep -oE 'ceiling=[0-9]+' | head -1 | cut -d= -f2)
```

So the arm verifies what §156a **tells a failing reader**, never what §156a **executes**. A divergence between the two — the guard declaring a bound it does not apply — is invisible to every arm in the block.

Measured at `469d9fb` (i.e. with #643 merged), deleting only the executed clause and leaving both messages intact:

```
$ perl -i -pe 's/ \|\| \[ "\$s156_n" -gt 60 \]// if $.==2336' scripts/test/smoke.d/70-gates-contentlocks.sh
$ ./scripts/test/smoke.sh | grep -E '156[opqr]|^smoke: '
✓ 156o   ✓ 156p   ✓ 156q   ✓ 156r
smoke: pass=1297 fail=0
```

The window guard silently loses its ceiling and **the suite reports fully green** — including `§156o`, the arm whose entire purpose is to assert the ceiling exists. Control: the same deletion applied to the messages *as well* reds `§156o` (`pass=1295 fail=2`), so the arm is not dead — it is measuring the wrong surface.

**This is disclosed, not hidden.** `:2495-2504` states the limitation accurately and names it deferred. This issue is the deferral's destination.

## Why

**Primary — the failure mode is a guard that silently stops guarding.** #643 existed because a floor-only guard let the content locks be asserted over §1.4–§3. The ceiling that fixed it is now protected by an arm that cannot see the ceiling's removal. A future edit that touches the condition and not the message re-opens #643's exact hole with a green suite. That is a false negative in the enforcement tier, and the tier's value is entirely that a green arm means something.

**Why now — the deferral has been ruled four times and has never had a destination.** `finding-judge` ruled `defer-to-issue` at rounds 1, 2, 3 and 4 of PR #672. Rounds 3 and 4 both recorded it explicitly as *"the outstanding item most likely to be lost"* and *"the only durable carry-over on the PR."* A deferral whose durable half never lands is indistinguishable from a drop. #672 is now merged; if this is not filed the ruling is simply gone.

The deferral was correct each time and this issue does not reverse it: round 1 measured that hoisting the decision would turn `§156p`'s clean RED at `e8aa0db` into an unbound-variable crash under `set -u`, trading #643's only repair evidence for hardening its ACs did not require. That constraint is real and is carried into the acceptance criteria below rather than discarded.

**MISSION fit:** *The mechanism* — the cost-asymmetry rule reserves the hard negative face for silent wrong-allows. A test arm that greens while the property it locks is absent is that shape one tier up: it does not permit a bad change, it **fails to notice** one, and it reports success while doing so.

**Existing coverage:** none. #670 is the adjacent §156j/§116 window issue — a *window resolution* defect, not a decision-source defect; the two do not overlap. #655 is §6.0 P4 arm classification. #642 is the worktree false-positive. No open issue mentions `s156o_accepts`, the F4b vector, or decision-vs-message divergence.

## Acceptance criteria

- [ ] **A smoke arm reds when the executed condition diverges from the declared bounds.** Verified by mutation in **both** directions: deleting `|| [ "$s156_n" -gt 60 ]` from the executed condition **while leaving both `156a:` messages intact** must red; the unmutated tree must stay at **`fail=0`**. *(The pass **count** is deliberately not pinned. The fix may legitimately land in `§156o` or in a new sibling arm, and a new arm raises the count — pinning `pass=1297` would make this criterion false by construction on a branch it permits. `fail=0` is the oracle.)*
- [ ] **The arm derives the executed bound from the guard's source rather than restating it.** An arm that hardcodes `60` satisfies the letter of AC1 while creating a **third** hand-synced declaration of the bound — the same defect one surface over, which is the outcome this Issue exists to prevent. Verified by changing the guard's ceiling to a different value in both the condition and the messages together: a correctly-deriving arm stays green, a hardcoded one reds.
- [ ] **The omission direction is not regressed.** The current arms already catch an *undeclared* bound; re-measure that they still do — an undeclared ceiling must continue to read as absent rather than crash.
- [ ] **`§156p`'s RED at `e8aa0db` is preserved as a clean assertion failure.** This is round 1's measured deferral reason and the binding constraint on the design: the arms must remain runnable against the pre-#643 tree without an unbound-variable crash under `set -u`. If the chosen design cannot preserve it, say so explicitly and state what replaces that evidence — do not silently trade it away. **The disclosure has a named locus**, so "stated explicitly" is checkable: the PR body's Plan section **and** an in-file note at the arm, in the register the current `:2495-2504` note uses.
- [ ] **The decision source is stated once, where the arm reads it.** Whatever is chosen — parsing the condition, a shared discriminator, or an explicit declared-vs-executed comparison — the in-file note at `:2495-2504` is replaced by an accurate description of what the arm now covers. It must not survive as a stale disclosure of a limitation that no longer exists.
- [ ] **The full suite passes in a window with no live agent worktree** (see #642 — a live worktree reds `§148i-cli` for unrelated reasons).

## Also in scope — one adjacent prose defect of the same class, measured

Surfaced by PR #672's review rounds and recommended for carrying here by two independent `finding-judge` invocations rather than repaired inside #672, where a fifth prose repair was the measured generator of that PR's own defects.

- [ ] **`:2660`'s tense claim is false today.** It reads *"ROUTING arms (157v–157x) are LOAD-BEARING RED at this commit"*; all three measure **green** at `469d9fb` (`✓ 157v ✓ 157w ✓ 157x`, re-run before filing). Anchor it to the commit where it was true, on the convention #643 adopted (`RED at <sha>`), or delete it. This is the same defect class #643 spent four rounds removing from §156 — it is systemic in this file, not local to the block that happened to get attention.

Scoped deliberately: it is in the file this issue already edits, it is a one-line fix with a stated oracle, and it is the kind of item that never justifies its own issue and therefore never gets fixed.

**Not carried:** PR #672's body cites the in-file disclosure by line range (`:2497-2504`, now `2495-2504`). A merged PR body is a historical record like a commit body, not a maintained surface — re-anchoring it would be churn, not repair. Recorded so it is not rediscovered as a defect.

## Out of scope

- **Rewriting `§156p`/`§156q`/`§156r`'s fixture construction.** They are measured correct and were mutation-verified in three independent review rounds. This issue changes where the *decision* comes from, not how the fixtures are built.
- **A sweep of every smoke arm that re-implements a guard's decision.** If the sweep is warranted, the classification is its own unit of work with its own rationale; this issue owns the one site with a measured reproduction.
- **The §157 routing block beyond the single false tense claim named above.** Its arms are green and correct; only the comment is wrong.
- **Re-litigating #643's floor/ceiling values.** `floor=18`/`ceiling=60` are measured and locked; this issue is about whether the arm can see them.

## Notes

Raised by `finding-judge` at round 1 of PR #672 and re-ruled `defer-to-issue` at rounds 2, 3 and 4. The reproduction above was run independently by the round-3 judge, the round-4 reviewer and the round-4 judge, and reproduced identically each time (`pass=1297 fail=0` with the ceiling silently absent).

The in-file disclosure at `:2495-2504` is accurate and was verified true in rounds 3 and 4 — it names the limitation and calls it deferred, and it does **not** claim a tracking issue exists. Nothing false shipped; this issue makes the deferral's destination real.

ilgyu-yi:
Closes #673

## Plan

`§156a`'s window guard is the only thing standing between §1.3.1's eleven content locks and the state #643 fixed, where the locks were asserted over §1.4–§3. Its dedicated arm, `§156o`, lifts the guard's **extractor** verbatim out of the guard's own source — but then **re-implements the guard's decision** from two integers scraped out of the `156a:` assertion *messages*. So every arm in the block measures what the guard **says**, and nothing measures what it **does**.

Measured on `main` at `469d9fb`: deleting only `|| [ "$s156_n" -gt 60 ]` from the executed condition, leaving both messages intact, leaves the suite at **`pass=1297 fail=0`** with `✓156o ✓156p ✓156q ✓156r`. The guard silently loses its ceiling and the suite reports success.

**The fix: stop re-implementing the decision and run it.** `s156o_accepts` lifts `§156a`'s condition line from the guard's own source — the technique `§156o` already uses for the awk extractor, proven backward-portable — strips the `elif`/`; then` wrapper, sets a function-local `s156_n`, and evaluates it in a **subshell with a three-valued result** (accept / reject / unevaluable), so a condition referencing a later-defined variable cannot crash the parent under `set -u`.

**`fix` relaxation not taken.** Doc → Test → Code is kept in full. The reproduce-first relaxation was available, but the reproduction already exists — recorded in the Issue and re-run on `main` before filing.

## The contest — two challengers, both wrong in instructive ways

The plan went through `planner` + two mutually-blind `plan-challenger`s (axes `correctness` and `simplicity/maintainability`) + `plan-reviewer`. **Both challengers returned `dominates-A`. Only one survived measurement.**

| mutant | Plan A | B1 (correctness) | B2 (maintainability) | **winner (B1 + grafts)** |
|---|---|---|---|---|
| unmutated | green | green | green | **all green** |
| `\|\|` → `&&` @ the condition | **GREEN (miss)** | `✗156p ✗156q` | **GREEN (miss)** | **`✗156p ✗156q`** |
| M6 messages→40, condition 60 | red | **GREEN (miss)** | `✗156s` | **`✗156u`** |
| M3 `ok` message only→40 | red | `✗156t` | `✗156s` | **`✗156t ✗156u`** |
| M2 →40 everywhere | green (**needs prose deleted**) | green, prose intact | green | **green, prose intact** |
| M1 delete `-gt 60` | red | `✗156p` | `✗156p ✗156s` | **`✗156p`** |
| `(( ))` refactor | red (fail-open cost) | green + still discriminating | `✗` 3 arms | **green + still discriminating** |
| `e8aa0db` splice | n/a | clean | clean | **`✗156o ✗156p ✓156q ✓156r`, no crash** |

**`||` → `&&` is one token, and it makes *both* bounds inert.** Every numeric operand and every message numeral stays byte-identical, so Plan A's parity scanner is **provably** green on it — which refutes A's central claim that *"declared ≡ executed ⟹ deciding on declared is faithful."*

**B2 claimed to dominate A and does not.** B2 extracts the operands and re-applies its own comparison, so a connective flip is invisible to it too. B2 never ran that mutant; the reviewer built B2's design verbatim and did: `pass=1294 fail=4`, all arms green, byte-identical to B2 unmutated. B2 dominates on diff size and fixture reuse only — a fake-diff on the decisive axis.

**B1 is not a clean winner either.** It misses M6 — both messages *consistently* declaring `ceiling=40` while the condition executes `60` — which is AC1's own wording. Hence the graft: B1's decision source **plus** A/B2's declared-vs-executed parity arm.

## Invariant preserved by measurement, not by argument

AC3 requires `§156p`'s clean RED at `e8aa0db` to survive — this is round 1 of PR #672's *measured* deferral reason, which rejected the shared-symbol hoist precisely because it turned that clean RED into a `set -u` crash. Splicing the winner's block into a `git archive e8aa0db` tree: `✗156o ✗156p ✓156q ✓156r`, clean `ng` assertion failures, **zero** `unbound variable` or `command not found`.

**Scope ruling.** Touching `s156o_accepts` is admissible — the Issue's out-of-scope bars only `§156p`/`q`/`r`'s *fixture construction*, and says verbatim *"This issue changes where the decision comes from"*. Plan A's scope-out was self-imposed, and touching it was measured **not** to cost the invariant.

## Two corrections applied to the plan before implementation

1. **The reviewer refuted one of my own plan's claims.** Round 1 recorded that `empty-on-both = agreement` was the invariant-1 rescue. Round 2 re-measured: with a symmetric pipeline both ceiling sets come out empty at `e8aa0db` and naive equality already agrees. The rule is **defensive, not load-bearing** — still carried, but the plan no longer claims it saves the invariant.
2. **The broadened lift anchor needed bounding.** `^elif .*s156_n.*; then$` also matches `§156r`'s own `elif [ "$s156o_hn" -ne "$s156_n" ]` — against a guard refactored to a form not naming `s156_n`, `-m1` would lift *that* line and green `p`/`q`/`r`/`s` **with no bounds at all**. The Code phase bounds the lift to the region before the first `156a:` message.

## Disclosed, not hidden

**None of the three candidate designs ever executes `§156a` itself** — all remain source-derived simulations, B1 closest since it evaluates the real condition text. The gap is real and is stated in the in-file note rather than papered over.

## Measured kill matrix — every mutant on an isolated `git archive` copy, live tree never mutated

Isolated-copy floor is 4 (`status-cache` ×2, `133a`, `135e-2`) — artifacts of the copy lacking `.git`/the self-symlink, unrelated to §156. So `1296/4` **is** fully green there.

| mutant | measured | verdict |
|---|---|---|
| unmutated (archive) | `1296/4`, all §156 green | ✓ |
| **M-AND** `\|\|` → `&&` in the condition | `1294/6` — **`✗156p ✗156q`** | ✓ **the mutant that decided the contest** |
| **M1** delete `\|\| [ "$s156_n" -gt 60 ]`, messages intact | `1294/6` — **`✗156p ✗156u`** | ✓ *stronger than planned* |
| **M6** both messages → `ceiling=40`, condition executes `60` | `1295/5` — **`✗156u`** | ✓ |
| **M3** `ok` message only → `40` | `1294/6` — **`✗156t ✗156u`** | ✓ |
| **M2** ceiling → `40` in condition **and** both messages | **`1296/4` fully green**, block prose byte-untouched | ✓ AC2 |
| **M4** `(( s156_n < … \|\| s156_n > … ))` refactor | **`1296/4` green** | ✓ refactor-tolerant |
| **M4d** same form minus the ceiling | `1294/6` — **`✗156p ✗156u`** | ✓ still discriminating |
| **M-OMIT** ceiling gone from condition **and** both messages | `1294/6` — **`✗156o ✗156p`**, `✓156q ✓156r ✓156s ✓156t ✓156u` | ✓ AC3 — absent, not crashed |
| **`e8aa0db` true splice** | `1294/6` — **`✗156o ✗156p ✓156q ✓156r`**, clean `ng`s, **0** noise lines mentioning `156` | ✓ **invariant 1** |
| **M-ANCHOR** condition rewritten to a shape not naming `s156_n` | `1291/9` — **five arms red, all `UNTESTED`** | ✓ fail-closed, loud, no crash |

**AC6, live tree, no agent worktree live:** `lint OK: 79 files clean`; `smoke: pass=1300 fail=0`; the `^[[:space:]]*(ok|ng) "156a:` scrape still matches exactly **2** lines, so the additions do not perturb `§156o`; real `SPEC.md` unmutated.

## Four things the implementation found that the plan and the Issue had wrong

1. **The plan's kill matrix understated M1.** It recorded `✗156p` only. Measured, M1 also reds `§156u` (`declared={18 60} executed={18}`) — deleting the executed clause while leaving the messages intact *is* a declared-vs-executed divergence, so the parity arm necessarily fires. Same for M4d. The composition is strictly stronger than the plan claimed.
2. **The bound on the lift is load-bearing, and now measured directly rather than argued.** Counterfactual built: same M-ANCHOR mutant, lift *unbounded*. `§156p` greens **spuriously** — the unbounded `-m1` lifts `§156r`'s own `elif [ "$s156o_hn" -ne "$s156_n" ]` and, with `s156o_hn=22` / `s156_n=368`, that "condition" happens to reject. Only `§156u` reds, and only because the lifted line carries no numerals — exactly the *"do not rely on that"* the plan review warned of. With the bound in place, all five arms red `UNTESTED`.
3. **`§156t` is *not* subsumed by `§156u`** in the direction the plan assumed. Under M3 both red, but `§156u` reports `declared={18 40 60}` with no indication the *messages* disagree with *each other* rather than with the condition. The subsumption holds on the pass/fail bit only; `§156t` is the arm that localises the defect.
4. **`§156o` was never the arm AC1's reproduction needed.** Under M1 and M-AND it stays green; it reds only for the text reason it always did. The repair lands entirely in `§156p`/`§156q` (behaviour) and `§156s`/`§156t`/`§156u` (text-vs-behaviour). `§156o`'s WITNESS-HONESTY note, which already calls it a bounds-not-witness arm reading the guard's *text*, remains accurate and was left alone.

## Two implementation deviations from the approved plan

1. **Three-valued dispatch is via stored verdicts, not at the `elif` sites.** `if s156o_accepts "$n"; then` cannot distinguish `rc=1` from `rc=2`, so it would have folded UNEVALUABLE into REJECT and **greened `§156p`/`§156q`** — a direct invariant-3 violation. The three verdicts are taken once, initialised to `2`, and each arm branches on all three. This adds an `UNTESTED` arm to `§156p`/`q`/`r`: a branch change, not a fixture-construction change, so the Issue's out-of-scope is untouched.
2. **The lift moved up next to the extractor lift.** `s156o_accepts` is consumed by `§156p` ~90 lines before the `§156s` block where the plan placed the lift, so leaving it there would have made every arm UNEVALUABLE. Bound and anchor are exactly as planned.

## Round 1 review (`be5d256`) — one real finding, one refuted

`code-reviewer` re-ran every load-bearing row of the matrix above and reproduced them all, then returned `ship after fix` on **one** finding — and it was, again, *prose about correct code*: an in-file comment claimed an unbounded lift makes **every** arm green, where six green and `§156u` reds. The PR body stated it correctly; only the comment overstated.

**The judge refused the reviewer's remedy.** The proposed rewrite repaired one false clause by pinning five arm identities plus a contingent reason, and its own added clause — that `§156u`'s red is *"incidental and not a property to rely on"* — is self-undermining: it labels the property unreliable while depending on it. So the clause was **severed**, not re-specified. What replaces it is a statement about what gets *lifted* rather than what the arms then do — an analytic consequence with no staleness path — keeping the judge's floor (the anchor's breadth and the `§156r` dependency, which #673 AC5 binds) and its ceiling (no arm-outcome enumeration, no numeral-contingent clause, no bound literal).

**One finding was refuted outright.** The reviewer flagged `s156o_cline` as the lone un-initialised sibling. Measured: `s156o_line` is *also* assigned inside the guard block with no up-front declaration and is untouched by this PR. The real convention is not *declared vs forgotten* but **read outside the block vs not** — and `s156o_cline` follows it. Initialising it would make the file *less* consistent with its own rule.

**One cosmetic fix landed** (`§156t` rendered `floor={ floor=18 }` on the pass path, so every green run printed it), bounded to the two display assignments; the verdict variables are computed independently and untouched.

**The judge also caught an error neither the reviewer nor I did:** this body's M-ANCHOR row read `1290/10` where the measurement is `1291/9` — the same defect class as the finding, on a durable surface. Corrected above.

**No further adversarial round for behaviour.** The lift bound is line `2363`; both edits sit after it, outside the lift's search range, and the `§156t` edit touches only message interpolation. The kill matrix is inert with respect to them.

## Checklist

- [x] **Doc** (`9dd8529`) — delete the **fourth** bound copy; anchor §157's stale tense claim; point the deferral note at #673.
- [x] **Test** (`ffb3ff0`) — `§156s` (lift precondition) and `§156u` (parity) added while `s156o_accepts` still reads the messages → clean RED at `pass=1297 fail=2`, with three same-input positive controls.
- [x] **Code** (`b52a9ea`) — condition lift (bounded anchor) + subshell three-valued eval; `§156t`; `§156p`/`q`/`r` branch on all three verdicts; deferral note replaced wholesale (AC5).
- [x] **Code** (`be5d256`) — de-numeralise the last illustrative bound. **Measured across the entire PR diff: zero added lines contain a bound numeral.** The block went from **four** hand-synced declaration sites to **two** — both of them the `156a:` messages `§156o` must grep by contract.
- [x] **Review fixes** (`6413a98`) — sever the false lift-bound clause (F1); strip `§156t`'s doubled label (F3). F2 refuted, no change.

## Docs touched

- [~] `SPEC.md` — N/A. No SPEC clause governs a smoke arm; this block's in-file comment is its contract SSOT (`3a6297a` precedent).
- [~] Changelog fragment — N/A. Smoke-internal, no adopter-observable surface (§18.7); `skip-changelog` to be applied.

## Ship gate

- [x] `code-reviewer` round 1 at `be5d256` → `ship after fix`; judged, fixed in `6413a98`
- [x] **AC6** — full suite green in a worktree-free window (#642): `pass=1300 fail=0`, `lint OK: 79 files clean`. The judge ruled this **structurally undischargeable by any worktree-isolated agent**, so it was run by the main session at the final head.
- [x] AC closeout posted on #673
- [x] head-pinned review at the final head (`6413a98`)

## Review cost, recorded

Four review rounds; **zero code defects in all four.** Round 1 found one false clause in a code comment (real, fixed). Rounds 2, 3 and 4 found, in order: a wrong numeral in this body, a wrong numeral in a PR comment, and nothing. `s156o_cbound` is computed at runtime — so `2363` vs `2364` was only ever a claim *about* a measurement, and no code path could carry a stale value.

Recorded because it is evidence for the review-round-inflation work being raised in priority, not as a complaint about the reviewers: round 1's finding was genuine, and the `finding-judge` refuted a bad nit, refused a reviewer remedy that would have added six new claims, and caught a slip in this body that neither the reviewer nor the author saw.

### DIFF
--- a/scripts/test/smoke.d/70-gates-contentlocks.sh
+++ b/scripts/test/smoke.d/70-gates-contentlocks.sh
@@ -2494,21 +2494,64 @@ fi
 # NO LIVE SPEC MUTATION. All three fixtures are copies under the preamble's $TMP,
 # removed by its EXIT trap; $S156_SPEC is only ever read.
 #
-# The arms drive the live guard's EXTRACTOR verbatim, but its DECISION is
-# re-implemented here from the two integers scraped out of the assertion messages
-# — so an executed condition that diverges from the declared bounds is NOT caught
-# by these arms (#643 round-1 F4b, deferred; tracked at #673). The extractor is lifted
-# verbatim out of the guard's own source line, and the bounds are read out of the
-# `156a:` assertion messages — i.e. the bounds the guard TELLS a failing reader about.
-# This covers the OMISSION direction only — a divergence, where the guard declares
-# a bound it does not execute, is the deferred F4b vector.
-# A bound the guard does not declare cannot bind anything, so an undeclared ceiling
-# reads here exactly as it behaves there: absent.
+# DECISION SOURCE (#673). Every arm here drives the live guard: its EXTRACTOR is lifted
+# verbatim out of the guard's own `s156_win=$(awk …)` source line, and its DECISION is
+# taken by EVALUATING the guard's own condition line, lifted the same way (see the lift
+# below and §156s). Nothing in this block re-implements §156a's comparison, so the two
+# directions are covered together:
+#   OMISSION  — a bound the guard never declares and never executes cannot bind anything,
+#               and reads here exactly as it behaves there: absent (§156p reds).
+#   DIVERGENCE — a bound the guard DECLARES but does not EXECUTE, or executes but does not
+#               declare, is caught by the composition: §156s (the condition can be read at
+#               all) + §156t (the two `156a:` messages agree with each other) + §156u
+#               (what the messages declare equals the numerals the condition executes).
+#               This is #643 round-1's F4b vector; #673 is its repair, not its deferral.
+#
+# UNCOVERED BY CONSTRUCTION, deliberately: a bound changed WRONGLY but CONSISTENTLY — in
+# the condition and in both messages together — stays green here. It has to. #673 AC2 asks
+# for exactly that: the arms must DERIVE the bound rather than restate it, so a ceiling
+# retuned everywhere at once is indistinguishable from a legitimate retuning. Whether the
+# two values are the RIGHT ones is a judgement no arm in this file makes (#643, locked). No
+# bound numeral is repeated in this block, on purpose — a copy here would be one more thing
+# to keep in sync and would silently satisfy AC1 while re-creating #673's own defect.
+#
+# FAIL-CLOSED ON A REFACTOR. The lift anchor is broad (`^elif .*s156_n.*; then$`, and an
+# arithmetic `(( s156_n < … || s156_n > … ))` rewrite is measured to lift and evaluate
+# cleanly), but a rewrite it cannot recognise yields an empty condition, which is
+# UNEVALUABLE, not accepting: §156s reds and §156p/§156q/§156r red as UNTESTED rather than
+# reporting a bound they never applied.
+#
+# RUNNABLE AGAINST e8aa0db (the pre-#643 Test commit, guard floor-only) — #643 round 1's
+# measured deferral reason and #673 AC3's binding constraint. Measured for this design:
+# `✗156o ✗156p ✓156q ✓156r`, clean `ng` assertion failures, no `unbound variable` under
+# `set -u`. The lifted condition names only `s156_n`, which s156o_accepts binds as a
+# function-local before evaluating it, so a floor-only condition evaluates there exactly as
+# it does here.
+#
+# SHARED BLINDSPOT, disclosed rather than mitigated: no arm in this block ever executes
+# §156a itself. All of them SIMULATE it from its source text — the extractor and now the
+# condition are the guard's real bytes, but they are re-run over fixtures, not observed
+# firing. An edit that leaves both source lines intact and breaks §156a some other way
+# (its surrounding `if`/`elif` chain, the variable it feeds) is invisible here.
 S156O_SRC="$SHELL_ROOT/scripts/test/smoke.d/70-gates-contentlocks.sh"
 s156o_msgs=""
 s156o_floor=""
 s156o_ceil=""
 s156o_prog=""
+# s156o_cbound — the guard's own first `156a` assertion-message line number. It is the
+# LIFT BOUND, and the bound is load-bearing rather than tidy: the lift anchor
+# `^elif .*s156_n.*; then$` is deliberately broad enough to survive a rewrite of the
+# condition's shape, and unbounded it also matches §156r's own
+# `elif [ "$s156o_hn" -ne "$s156_n" ]; then` below. Against a guard refactored to a form
+# that no longer names s156_n, an unbounded -m1 lift picks up THAT line instead and — with
+# s156o_hn bound in-suite — every arm in this block greens with no bounds in force at all.
+# Searching only ahead of the first `156a` message cannot reach any arm of this block.
+#
+# s156o_cond — §156a's executed condition, lifted verbatim within that bound and stripped
+# of its `elif`/`; then` wrapper. Empty means the lift failed, which every consumer treats
+# as UNTESTED (never as agreement).
+s156o_cbound=""
+s156o_cond=""
 if [ -f "$S156O_SRC" ]; then
   s156o_msgs=$(grep -E '^[[:space:]]*(ok|ng) "156a:' "$S156O_SRC")
   s156o_floor=$(printf '%s\n' "$s156o_msgs" | grep -oE 'floor=[0-9]+' | head -1 | cut -d= -f2)
@@ -2518,15 +2561,37 @@ if [ -f "$S156O_SRC" ]; then
   case "$s156o_line" in
     *"awk '"*) s156o_prog=${s156o_line#*awk \'}; s156o_prog=${s156o_prog%\'*} ;;
   esac
+  s156o_cbound=$(grep -n -m1 -E '^[[:space:]]*(ok|ng) "156a:' "$S156O_SRC" | cut -d: -f1)
+  if [ -n "$s156o_cbound" ]; then
+    s156o_cline=$(head -n "$s156o_cbound" "$S156O_SRC" | grep -m1 -E '^elif .*s156_n.*; then$')
+    case "$s156o_cline" in
+      elif\ *\;\ then) s156o_cond=${s156o_cline#elif }; s156o_cond=${s156o_cond%; then} ;;
+    esac
+  fi
 fi
 
-# s156o_accepts <non-blank-line-count> — 0 when the guard's DECLARED bounds admit a
-# window of that size. Mirrors §156a's decision; an undeclared bound is inert.
+# s156o_accepts <non-blank-line-count> — THREE-VALUED, and the three values are the point:
+#   0 ACCEPT       — the guard's own condition, evaluated at that count, does NOT trip
+#   1 REJECT       — it trips
+#   2 UNEVALUABLE  — the condition could not be lifted, or evaluating it failed
+# It does not mirror §156a's decision any more; it RUNS it. The lifted text is EVALUATED in
+# a SUBSHELL with `s156_n` bound function-locally, so a condition naming a variable that is
+# unbound at this point in the suite kills the subshell under `set -u` and comes back as
+# UNEVALUABLE instead of killing the run. The two sentinel exits (10/11) are what separate a
+# genuine verdict from any failure mode of eval itself — a shell that died on the condition
+# exits 1/2/127, all of which land in UNEVALUABLE. Callers must branch on all three: folding
+# 2 into "not accepted" would green §156p/§156q on a condition they never managed to run.
 s156o_accepts() {
-  local n="$1"
-  [ -n "$s156o_floor" ] && [ "$n" -lt "$s156o_floor" ] && return 1
-  [ -n "$s156o_ceil" ] && [ "$n" -gt "$s156o_ceil" ] && return 1
-  return 0
+  # shellcheck disable=SC2034  # read by the lifted condition below, via eval
+  local s156_n="$1"
+  local rc=0
+  ( eval "if $s156o_cond; then exit 11; else exit 10; fi" ) >/dev/null 2>&1
+  rc=$?
+  case "$rc" in
+    10) return 0 ;;
+    11) return 1 ;;
+    *)  return 2 ;;
+  esac
 }
 
 # Fixtures. -1 means NOT MEASURED, so no arm can read a zero as agreement.
@@ -2555,6 +2620,15 @@ if [ -f "$S156_SPEC" ] && [ -n "$s156o_prog" ] && mkdir -p "$S156O_DIR" 2>/dev/n
   fi
 fi
 
+# Verdicts, taken ONCE and stored, because s156o_accepts is three-valued and
+# `if s156o_accepts "$n"; then` would silently fold UNEVALUABLE into REJECT — greening
+# §156p/§156q on a condition the arms never managed to run. 2 (UNEVALUABLE) is also the
+# initial value, so an unbuilt fixture is never mistaken for a verdict.
+s156o_hv=2; s156o_ov=2; s156o_tv=2
+if [ "$s156o_hn" -ge 0 ]; then s156o_accepts "$s156o_hn"; s156o_hv=$?; fi
+if [ "$s156o_on" -ge 0 ]; then s156o_accepts "$s156o_on"; s156o_ov=$?; fi
+if [ "$s156o_tn" -ge 0 ]; then s156o_accepts "$s156o_tn"; s156o_tv=$?; fi
+
 # §156o — AC1: the guard names BOTH bounds in its own assertion message, so a failure
 # says WHICH SIDE tripped. Reads the guard's text; §156p reads its behaviour.
 if [ -z "$s156o_msgs" ]; then
@@ -2573,10 +2647,12 @@ elif [ "$s156o_ofence" -ne $((s156o_hfence + 1)) ]; then
   ng "156p: over-extension fixture did not take (fence markers healthy=$s156o_hfence over=$s156o_ofence, expected +1) — arm would be vacuous (#643)"
 elif [ "$s156o_on" -le "$s156o_hn" ]; then
   ng "156p: over-extension fixture did not over-extend (over n=$s156o_on vs healthy n=$s156o_hn) — arm would be vacuous (#643)"
-elif s156o_accepts "$s156o_on"; then
-  ng "156p: an unclosed fence in §1.3.1 runs the window to $s156o_on non-blank lines (healthy=$s156o_hn) and the declared bounds ACCEPT it (floor=${s156o_floor:-<none>} ceiling=${s156o_ceil:-<none>}) — the content locks would be called load-bearing over §1.4–§3 (#643)"
+elif [ "$s156o_ov" -ge 2 ]; then
+  ng "156p: §156a's executed condition could not be evaluated at n=$s156o_on (lifted='${s156o_cond:-<none>}') — the ceiling is UNTESTED, not satisfied (#673)"
+elif [ "$s156o_ov" -eq 0 ]; then
+  ng "156p: an unclosed fence in §1.3.1 runs the window to $s156o_on non-blank lines (healthy=$s156o_hn) and the guard's EXECUTED condition ACCEPTS it (declared floor=${s156o_floor:-<none>} ceiling=${s156o_ceil:-<none>}) — the content locks would be called load-bearing over §1.4–§3 (#643, #673)"
 else
-  ok "156p: an unclosed fence in §1.3.1 over-extends the window to $s156o_on non-blank lines and the declared bounds REJECT it (floor=${s156o_floor:-<none>} ceiling=${s156o_ceil:-<none>}) (#643)"
+  ok "156p: an unclosed fence in §1.3.1 over-extends the window to $s156o_on non-blank lines and the guard's EXECUTED condition REJECTS it (declared floor=${s156o_floor:-<none>} ceiling=${s156o_ceil:-<none>}) (#643, #673)"
 fi
 
 # §156q — AC3, a BOUND (not a witness; passes before and after). The ceiling must not
@@ -2585,10 +2661,12 @@ if [ "$s156o_tn" -lt 0 ]; then
   ng "156q: renamed-heading fixture not built or rename did not land — the floor is UNTESTED, not satisfied (#643)"
 elif [ "$s156o_tn" -ne 0 ]; then
   ng "156q: renamed-heading fixture left a non-empty window (n=$s156o_tn, expected 0) — arm would be vacuous (#643)"
-elif s156o_accepts "$s156o_tn"; then
-  ng "156q: a renamed §1.3.1 heading collapses the window to 0 non-blank lines and the declared bounds ACCEPT it (floor=${s156o_floor:-<none>} ceiling=${s156o_ceil:-<none>}) (#643)"
+elif [ "$s156o_tv" -ge 2 ]; then
+  ng "156q: §156a's executed condition could not be evaluated at n=$s156o_tn (lifted='${s156o_cond:-<none>}') — the floor is UNTESTED, not satisfied (#673)"
+elif [ "$s156o_tv" -eq 0 ]; then
+  ng "156q: a renamed §1.3.1 heading collapses the window to 0 non-blank lines and the guard's EXECUTED condition ACCEPTS it (declared floor=${s156o_floor:-<none>} ceiling=${s156o_ceil:-<none>}) (#643, #673)"
 else
-  ok "156q: a renamed §1.3.1 heading collapses the window to 0 non-blank lines and the declared bounds REJECT it — the ceiling does not blind the floor (#643)"
+  ok "156q: a renamed §1.3.1 heading collapses the window to 0 non-blank lines and the guard's EXECUTED condition REJECTS it — the ceiling does not blind the floor (#643, #673)"
 fi
 
 # §156r — AC4, a BOUND (not a witness; passes before and after). The two rejections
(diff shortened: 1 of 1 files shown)


ilgyu-yi:
## Finding triage (round 1)

`dispatch:` findings from a worktree-isolated `code-reviewer` (round 1, second attempt — the first died on a transient 529 with no verdict); judgment from a fresh, worktree-isolated `finding-judge` with no shared context. Limb (a) is not self-observable — recorded as given.

`reviewed-head: be5d256f5413cd8a98791eee5540a17d886d6f90` · `prior-round: none (first round)`

**Artifact pin confirmed** — `341a6004774b120787eb67316654549a301bc7239f7a2725809cfc1e98daf159`, byte-identical to the reviewer's. Head re-read after judging: unmoved. Baseline reproduced independently: `pass=1296 fail=4` on an isolated copy, failures exactly the documented floor, all **20** §156 arms green.

**3 findings judged: 2 confirmed, 1 refuted. fix-now ×2, drop ×1.** `code-reviewer`'s `ship after fix` is untouched.

---

### F1 — the lift-bound note claims "every arm greens"; measured, six green and one reds · **fix-now**

```
finding: F1 — :2546-2547 claims "every arm in this block greens with no bounds in force at all"
site: scripts/test/smoke.d/70-gates-contentlocks.sh#s156o_cbound
axis-key: …#s156o_cbound::stale-prose-claim
verdict: confirmed
evidence: executed=bash scripts/test/smoke.sh (M-ANCHOR mutant + lift bound removed, isolated copy)
          result=pass=1295 fail=5 — ✓156o ✓156p ✓156q ✓156r ✓156s ✓156t ✗156u
                 (declared={ 18 60 } executed={ }). Six arms green, NOT every arm.
action: fix-now
remedy-survives: no
axis: shorter<->longer   target-position: interior
```

**FLOOR** — the note must still state that the anchor is deliberately broad **and** that unbounded it matches `§156r`'s own `elif [ "$s156o_hn" -ne "$s156_n" ]; then`. That dependency was measured directly: bounded lift returns empty (`rc=1`), unbounded returns `§156r`'s line (`rc=0`). Deleting it strips the bound of its stated rationale, which #673 AC5 binds.

**CEILING** — **no** enumeration of which arms green or red, and **no** clause resting on the "lifted line carries no numerals" property. Measured ground, not preference: `s156u_exe` greps operator-anchored numerals and came out empty *only* because `§156r`'s condition happens to be `-ne` against a variable. Give `§156r`'s condition any numeric literal and the reviewer's proposed clause becomes false — it encodes a fact with a **live staleness path**, in the file whose documented failure mode is exactly that. Also: no `18`/`60` literal (the block's own rule at `:2515-2516` and #673 AC2 both bind).

**Why the reviewer's remedy does not survive.** The finding is real and correctly measured, but the proposed replacement pins **five arm identities** plus a contingent reason — it repairs one false clause by adding six new re-measurable ones. Its own added clause (*"incidental and not a property to rely on"*) is self-undermining: it labels the property unreliable while depending on it. The false clause is **severable** — the two surrounding sentences measured true and carry the load-bearing justification on their own. The author owns the replacement; the judge bounds it and does not write it.

### F2 — **REFUTED.** The premise is false and the real convention is exceptionless · **drop**

```
finding: F2 — `s156o_cline` at :2566 is the one un-initialised sibling; convention drift
verdict: refuted
evidence: executed=/usr/bin/grep -nE 's156o_(msgs|floor|ceil|prog|cbound|cond|line|cline)' <head blob>
          ; gh pr diff 674 | /usr/bin/grep -E '^[+-].*s156o_line'  (rc=1)
          result=`s156o_line` (:2560) is ALSO assigned inside the `if [ -f "$S156O_SRC" ]` block
                 with no up-front declaration, and is UNTOUCHED by this PR — pre-existing
                 precedent, not drift.
action: drop
```

The split is not "declared vs forgotten" — it is **read outside the guard block vs not**:

- **Declared up front** (`:2537-2540`, `:2553-2554`): `s156o_msgs`, `s156o_floor`, `s156o_ceil`, `s156o_prog`, `s156o_cbound`, `s156o_cond` — all six are read *outside* the block (`:2588`, `:2601`, `:2634`–`:2760`). Under `set -u`, with `S156O_SRC` absent the block never runs, so the up-front init is **load-bearing**, not stylistic.
- **Not declared**: `s156o_line` **and** `s156o_cline` — both pure intermediates read only *inside* the block.

`s156o_cline` therefore **follows** the convention; initialising it would make the file *less* consistent with its own rule. This is precisely the #672 lineage's signature — a nit asserted about correct code. Refuted, not deferred. Retained as `already-refuted (round 1)`; not re-litigable without new evidence.

### F3 — `§156t`'s message doubles its own label · **fix-now**

```
finding: F3 — §156t's message renders `floor={ floor=18 }` instead of `floor={ 18 }`
site: scripts/test/smoke.d/70-gates-contentlocks.sh#s156t_floors
verdict: confirmed
evidence: executed=bash scripts/test/smoke.sh (isolated copy)
          result=live PASS-path line reads `✓ 156t: … (floor={ floor=18 } ceiling={ ceiling=60 })`
                 against §156u's clean `declared={ 18 60 }` on the SAME run. Observed in real
                 suite output, not simulated.
action: fix-now   remedy-survives: yes
axis: none — discrete (the prefix is either carried or stripped; no continuum)
```

**FLOOR** — touch only the two display assignments `s156t_floors`/`s156t_ceils` (`:2723-2724`), whose sole consumers are the two message strings (`:2730`, `:2732`). **CEILING** — do **not** touch `s156t_nf`/`s156t_nc` (`:2725-2726`) or the branch conditions; those compute the verdict independently, so no arm's pass/fail bit can move. The remedy is the pattern already used two lines up at `:2557-2558`. `§156t` ships **new** in this PR, so the wart is being introduced, not inherited.

Born observe-only per SPEC §6.0 P3 — recorded, not gated. Real (on the **pass** path, so every green run prints it), cosmetic, in scope, mechanically safe.

---

## Does the fix warrant a further review round? — **No, for behaviour.**

The change is **inert with respect to every measurement already taken**, verified by construction against this file's self-referential greps:

- `s156o_cbound` = **2363** (the first `156a:` message) *(corrected — this comment originally said 2364 in two places; the first correction reached only one of them, which is the same not-every-surface generator this PR keeps hitting).* F1's edit is at `2541-2548`, F3's at `2723-2724` — both **after** the bound, so neither can move it.
- The condition lift is bounded to lines 1–2363; both edits sit outside it. `s156o_line` and `s156o_msgs` (`^\s*(ok|ng) "156a:`) are likewise unreachable from a `#`-prefixed comment or a `156t:`-tagged string.
- F3 touches only message-interpolation variables; the verdict variables are computed separately.

So the kill matrix (M-AND, M1, M2, M3, M4, M4d, M6, M-OMIT, M-ANCHOR, the `e8aa0db` splice) does **not** need re-running.

**Two things do bind at the new head:** one unmutated full-suite run (AC6 requires it at the final head regardless), and a **human/author re-read of the rewritten comment for truth** — no run checks prose truth, and prose is the exact surface that failed five times in #672. The entire residual risk of F1's fix lives there, which is why the ceiling forbids new arm-pinned or numeral-contingent claims.

## AC6 — genuinely outstanding, and structurally so

#673's sixth criterion is explicit and unchecked: *"The full suite passes in a window with no live agent worktree (see #642)."*

- The reviewer could not discharge it (worktree-isolated → `§148i-cli` reds per #642).
- **The judge cannot either** — also worktree-isolated. **No worktree-isolated agent can ever discharge AC6**; it is unreachable from inside the review tier by construction.
- The author's `pass=1300 fail=0` is an unverified self-report. The arithmetic is *consistent* (1296 + the 4-failure archive floor = 1300 assertions), but consistency is not confirmation.
- Once F1/F3 land the head moves, so even that self-report no longer covers the merged artifact.

**Ruling: outstanding — must be run by the main session at the FINAL head, after agent worktrees are torn down.** A merge precondition, not an item for the author's fix list.

## Observations — not judged records

- **The PR body's M-ANCHOR row is off by one.** It states `1290/10`; the reviewer reported and the judge measured `1291/9` (five arms red + the 4-failure floor). Both sum to 1300, so it is a transcription slip — flagged because it is **the same defect class as F1**, a wrong measured claim on a durable surface, and the PR body is still editable.
- The reviewer's verdict token is untouched and unreinterpreted; F2's refutation does not alter it.

## Independence

Limb (b) holds by construction — the judge proposed no remedy of its own, only bounds on the author's. It emits no `ship`/`block`/`approve` token and enters no tally.

ilgyu-yi (commented):
**Verdict: APPROVE** — `code-reviewer` returned `ship` at this head after four review rounds.

`reviewed-head` blind-compared against the privately-computed PR head; artifact pin `sha256 8b4bd5b47268526ace2a8db06a05d7b3e29aab3e06cb9af9bc743230f6e6cbc4`, re-verified MATCH after every run.

**What was fixed.** `§156a`'s window guard is the only thing keeping §1.3.1's eleven content locks from being asserted over §1.4–§3. Its dedicated arm re-implemented the guard's *decision* from integers scraped out of the guard's own assertion *messages* — so it verified what the guard said and never what it did. On `main` at `469d9fb`, deleting only the executed ceiling clause left the suite fully green. `s156o_accepts` now lifts the guard's condition line from its own source within a bounded anchor and evaluates it in a subshell with a three-valued result, so an unevaluable condition fails closed and loudly rather than folding into REJECT.

**Measured, every mutant on an isolated copy, live tree never mutated:**

- `||` → `&&` in the condition (both bounds inert, every operand and message numeral byte-identical) → `✗156p ✗156q`. This is the mutant that decided the plan contest: **both** losing designs measured green on it.
- delete the executed ceiling clause, messages intact → `✗156p ✗156u`
- both messages declare a ceiling the condition does not execute → `✗156u`
- one message only → `✗156t ✗156u`
- ceiling moved in condition **and** both messages together → **fully green**, block prose untouched
- `(( ))` refactor → green **and still discriminating**; the same form minus the ceiling still reds
- condition renamed off the counter variable → five arms red, all `UNTESTED`, fail-closed and loud
- splice onto the pre-fix tree → clean `ng` assertion failures, **no** `unbound variable` crash

Every arm has a same-input positive control, so no green is fail-closed-by-accident. All six acceptance criteria met; AC6 was discharged by the main session in a worktree-free window (`pass=1300 fail=0`, `lint OK: 79 files clean`) — the judge ruled it structurally undischargeable by any worktree-isolated agent.

**The bound-numeral invariant holds at zero.** No added line in the entire PR introduces a bound numeral, and the block went from four hand-synced declaration copies to two — both of them the assertion messages the sibling arm must grep by contract.

**Review record, stated plainly.** Four rounds, **zero code defects in all four**. Round 1's finding was genuine — a false clause in a code comment. Rounds 2 and 3 each found a wrong numeral in PR metadata; round 4 found nothing. The `finding-judge` refuted one nit outright on a false premise, refused a reviewer remedy that would have repaired one false clause by adding six new re-measurable ones, and caught a transcription slip in the PR body that neither the reviewer nor the author had seen.

PR #675, opened 2026-08-17
fix(#670): make §156j's SPEC windows agree with §116's, and enforce it

### LINKED ISSUE #670: smoke §156j's §1.8 window still collapses on a renumber, and nothing enforces that it agrees with §116's
## What

`scripts/test/smoke.d/70-gates-contentlocks.sh` (§156j) re-derives §116's window expressions under an `s156_` prefix. **The two have diverged in both halves, nothing enforces that they agree, and §156j's copy still carries the §1.8 defect that §116's no longer does.**

| expression | §156j has | §116 has at `main` | diverged at |
|---|---|---|---|
| §1.9 posture window | `/^## 2\. /{exit}` | `$0~endre{exit}` (shared `S116_END_RE`) | #644 / PR #667 |
| §1.8 lever window | `/^### 1\.8 /{i=1;next} /^### 1\.9 /{exit}` | title anchor + shared `S116_END_RE` | #668 / PR #669 |

Measured at `main`, on a SPEC copy with §1.8 renumbered:

```
§156j lever window :  1 → 0     ← still collapses
§116  lever window :  6 → 6     ← hardened by #669
code references from 70-gates-contentlocks.sh to §116's helpers : 0
```

On that renumber §156j emits *"SSOT change sweep rows mis-shaped or absent — §1.8 lever rows=0"* while the row is **present and correctly shaped** — only the window failed to open.

**The re-derivation itself is legitimate** — `smoke.sh`'s header reserves cross-section symbols for `_preamble.sh`, so §156 genuinely cannot read `s116_*`. What is missing is any mechanism that makes the two agree, or notices when they stop.

### What already changed, and why this Issue is narrower than when it was filed

This Issue was filed while PR #669 was open, and §156j then carried a comment asserting its expressions were *"§116's verbatim"*. **That assertion is gone**: #669's own commit `e01e8ae` — authored within eight minutes of this Issue being filed (measured 7m27s) — replaced it with an honest disclosure that they are no longer verbatim, citing this Issue.

So the *false claim* is already remediated. What remains is the **substance**: the windows still disagree, the disagreement is still unenforced, and §156j's §1.8 window still collapses on a renumber. The narrowing is recorded rather than quietly absorbed, because the decay is itself the pattern this Issue is about — a claim about another file's contents went stale within minutes, and no arm noticed.

### The exposure is the window, not the predicate

§156j is green against a new `### 1.8.5` sibling — but **not** because its window is sound. Measured, the sibling **is** admitted into §156j's window exactly as it was into §116's, and counting every `| **` row there would inflate 6 → 7. §156j survives only because its **predicate** counts one specific row (`^\| \*\*SSOT change sweep\*\*`), so an admitted sibling contributes nothing to *that* count. The immunity is a property of the predicate; **the window is equally fragile**, and any future arm that widened the predicate would inherit the terminator half too.

## Why

**MISSION fit:** *The mechanism* — the cost-asymmetry rule makes a wrong nudge cheap only when it is ignorable. §156j's failure names a cause ("mis-shaped or absent") that is not the actual cause, which is the non-cheap kind: an operator repairs a row that was never broken.

**Why now.** The claim's decay is the point. It went half-false at #644/PR #667 and fully false at #668/PR #669 — twice, in PRs that had no reason to look at this file, with no arm noticing. A claim that decays silently by design is worth fixing as a *class*, not just patching the two instances, which is why this Issue is scoped to the claim's enforceability rather than only to the window.

PR #669's round-1 review found this; its `finding-judge` ruled the window repair out of that PR's scope — repairing it there would have been an untested cross-file change in a PR whose Test commit was already landed, inverting Doc→Test→Code — and required this Issue be filed **before** #669 merges, so no artifact asserts a tracking item that does not exist.

**Existing coverage:** none. #643 targets a **different** window in this same file (§1.3.1's content-lock extractor, its floor/ceiling asymmetry). #668 scoped `s116_levers` only, in the other file, and its ACs do not reach here. No Issue covers §156j or the verbatim claim.

## Acceptance criteria

- [ ] **§156j's window survives a §1.8 renumber.** Verified by mutation: a temp SPEC copy with §1.8 renumbered must leave §156j green. It **reds against current code** — measured today it yields `s156_lever=0` against an expected 1.
- [ ] **The two files state the *contract* they share, not a claim about each other's bytes.** Dropping the false "verbatim" assertion is **already done** (`e01e8ae`) and does **not** discharge this AC. What is required is the positive half: a statement of what the two windows must agree *about* — the terminator rule and the start-anchor rule — sited where a reader of either file will find it, so a future editor of one knows the other exists.
- [ ] **The agreement is enforced by an arm that reds when the two stop agreeing.** This is the point of the Issue. Non-enforcement is admissible **only** if all three hold: (i) the arm was actually built, (ii) its measured cost is recorded as a number — runtime, or false-red count over the suite's existing fixtures — and (iii) that number exceeds a threshold **stated in the same commit**, so the bar is fixed before the result is known rather than after. The record lives **in the arm file at the site**, not in a PR comment, because PR comments are not where the next editor of these two windows will look. An unmeasured "not worth it", or a threshold chosen after seeing the measurement, does **not** discharge this AC — that is the prose-shaped escape that let the previous claim decay unnoticed across two PRs and then go stale within eight minutes of being filed.
- [ ] **A paired arm asserts the direction that must NOT change**: the SSOT-change-sweep row, present and correctly shaped, still yields `s156_lever=1` and `s156_posture=1` on an unmodified SPEC. Measured today it passes, so it is a **bound, not a witness**, and must be labelled as such in-file.
- [ ] **The real `SPEC.md` is never mutated** by any arm.
- [ ] The §1.9 posture half is addressed on the same terms as the §1.8 half — it diverged first and is equally unenforced.

## Out of scope

- **`s116_levers` / `s116_posture_rows` themselves.** `s116_posture_rows` is #644, landed. `s116_levers` is #668, whose PR #669 is **open** at the time of filing — this Issue is deliberately filed before that merge, and must not re-open either.
- **The `smoke.sh` cross-section symbol reservation.** Lifting it so §156 could read `s116_*` directly is a different and larger change to the suite's structure; this Issue takes the reservation as given.
- **§156's other arms**, including the §1.3.1 content-lock window — that is #643.
- **Widening §156j's count** from the one SSOT-change-sweep row to all lever rows. It counts one row deliberately (a mis-shaped row would otherwise cancel out and keep §116 looking balanced); this Issue does not change what it counts, only which window it counts within.

## Notes

Surfaced by the `code-reviewer` on PR #669 and confirmed by its round-1 `finding-judge`, which measured that the posture half had **already** diverged at that PR's merge base — so the claim was half-false before #669 touched anything.

The measurements above were re-run by the author before filing. One correction from that run, recorded because it narrows the Issue: an earlier reading described §156j as carrying "the exact defect" #668 removes. Measured, it carries only the start-anchor half — the sibling case is green, because §156j counts one specific row rather than every lever row.

ilgyu-yi:
Closes #670

## Plan

`scripts/test/smoke.d/70-gates-contentlocks.sh` (§156j) re-derives §116's two SPEC-window expressions under an `s156_` prefix — legitimately, since `smoke.sh`'s header reserves cross-section symbols for `_preamble.sh`. The copies had diverged, **nothing enforced that they agree**, and §156j's windows collapsed on a renumber while reporting a cause that is not the actual cause: *"SSOT change sweep rows mis-shaped or absent"* when the row is present and correctly shaped.

Measured at the base, on a SPEC copy with §1.8 / §1.9 renumbered:

| window | healthy | renumbered |
|---|---|---|
| §156j lever | 1 | **0** |
| §156j posture | 1 | **0** |
| §116 lever | 6 | 6 |

**A correction to the Issue's own table, measured.** #670 reads as if only this copy's posture window were fragile. It is not — §116's `s116_posture_rows` opens on the literal `### 1.9 ` as well, and collapses **66 → 0** on the same renumber. Only the *lever* window was hardened to a title anchor (#668). Hardening §116's posture is explicitly out of scope for #670, so §156j is brought to **agree** with §116 on the lever half and held to a renumber **bound** on the posture half (its own §156x), not to posture-vs-posture agreement — and the shared residual is disclosed at both sites instead of being silently inherited.

**The fix.** State the contract as two rules the windows agree *about* — a **start rule** (open on the section's title, never its number) and a **terminator rule** (close on the shared `^## |^### `) — then bring §156j's lever and posture windows to satisfy both, and add arms that red when the two stop agreeing (§156v) or when a renumber collapses either window (§156w/§156x), with a paired unmodified-SPEC bound (§156y). The enforcing arms lift both files' window definitions from their own source and compare behaviour on `$TMP` SPEC copies, reusing the source-lift technique #673 established in this same file rather than adding a third copy of the rule.

**Rebase note (this session).** Rebased onto current main. #676 (merged since filing) added a real `### 1.10` section, which collided with the §156v–y fixture's renumber target `1.10`/`1.11` (the fixture guards that the target number is unused) — so the arms initially red as *UNTESTED* rather than window-collapse. The Test commit `4a63cac` moves the renumber targets to genuinely-unused numbers (`1.42`/`1.43`) and keeps the fail-loud guard. #676's §1.10 also sat between §1.9 and §2, so §156j's old posture terminator `/^## 2\. /` over-ran §1.10; the shared-terminator fix closes it correctly at §1.10's heading (count unchanged on the unmodified SPEC — §156y green).

**Why no planner contest.** Two files, no schema / API / migration — under the 3+-file threshold for a required `planner` dispatch.

## Checklist

- [x] **Doc** (`cf5be28`) — the shared window contract stated in both files, with the §116-posture residual disclosed.
- [x] **Test** (`d7cdc71` + `4a63cac`) — §156v (agreement), §156w/§156x (renumber survival), §156y (unmodified-SPEC bound); red against pre-fix code for the right reason (window collapse), and the fixture-target rebase fix so the arms actually take.
- [x] **Code** (`5f9fc0c`) — §156j's lever and posture windows open on the section title and close on the shared `^## |^### ` terminator.

## Docs touched

- [~] `SPEC.md` — N/A. No SPEC clause governs a smoke window; these two files are each other's SSOT here and both carry the contract.
- [~] Changelog fragment — N/A per §18.7: test-only change, no adopter-observable surface; `skip-changelog` applied.

## Test plan

- [x] §156v/§156w/§156x red against pre-fix code as window-collapse (not UNTESTED), §156y green — intermediate state pinned during Code phase.
- [x] After the fix: §156v/§156w/§156x/§156y all green, §156j still green, full smoke `pass=1323 fail=0` in a worktree-free window (#642).

## Ship gate

- [x] `code-reviewer` clean at head — ship at 5f9fc0c (head blind-compare matched)
- [x] full smoke green in a worktree-free window — pass=1323 fail=0 at the Code commit (#642)
- [~] CI green — N/A at ready time; verified at the merge gate
- [x] AC closeout posted on #670

### DIFF
--- a/scripts/test/smoke.d/50-perproject-recall.sh
+++ b/scripts/test/smoke.d/50-perproject-recall.sh
@@ -2670,8 +2670,23 @@ S116_END_RE='^## |^### '
 # terminator rule but still opens on the literal `### 1.9 `, so it collapses to 0 on
 # a §1.9 renumber (measured: 66 -> 0). That is a KNOWN residual, out of scope for
 # #670 and unhardened here on purpose — recorded so the next editor does not read
-# the posture helper as already conforming. §156j is held to AGREEMENT with these,
-# not ahead of them, and §156v reds if the two stop agreeing.
+# the posture helper as already conforming.
+#
+# WHERE THE AGREEMENT IS ENFORCED, and its one deliberate asymmetry. §156v reds if the
+# two stop agreeing; it compares the two sides' window OPENNESS on shared $TMP fixtures,
+# on the LEVER half, where both conform. The POSTURE half is held instead by §156x, as a
+# renumber bound on §156j alone — not as agreement with the helper below. That is forced,
+# not chosen: #670 requires §156j's posture to survive a §1.9 renumber and puts hardening
+# `s116_posture_rows` out of scope, so once §156x is satisfied §156j's posture is AHEAD of
+# this helper by exactly that residual, and a posture-vs-posture comparison would red on
+# the two legitimately differing. The lever half is held to agreement, not ahead.
+#
+# THE LIFT DEPENDS ON THESE THREE LINE SHAPES, not on their contents: `S116_END_RE=` at
+# column 0, `s116_lever_rows() {` / `s116_posture_rows() {` at column 0, and a closing `}`
+# at column 0. §156v/§156x lift them by source text (they must not CALL them — the
+# reservation above), so a reshaping of the definitions is fail-closed there: the lift
+# returns empty and every arm reds as UNTESTED rather than reporting agreement it never
+# measured. Reshape freely; just expect to re-anchor §156v.
 
 # NEITHER END IS A SECTION NUMBER (#668). The window used to open on the literal
 # `### 1.8 ` and close on the literal `### 1.9 `, so it broke in both directions:
--- a/scripts/test/smoke.d/50-perproject-recall.sh
+++ b/scripts/test/smoke.d/50-perproject-recall.sh
@@ -2656,6 +2656,23 @@ s116_spec="$SHELL_ROOT/SPEC.md"
 # tracking is deliberately NOT attempted here.
 S116_END_RE='^## |^### '
 
+# THE SHARED WINDOW CONTRACT (#670). `70-gates-contentlocks.sh` §156j re-derives
+# these same two windows under an `s156_` prefix — it cannot call the helpers below,
+# because smoke.sh's header reserves cross-section symbols for _preamble.sh. The two
+# copies must agree, and what they agree ABOUT is stated here and at §156j, not as a
+# claim of byte-identity (that claim decayed twice unnoticed):
+#
+#   START RULE — a window opens on the SECTION'S TITLE, never on its number (#668).
+#   TERMINATOR — a window closes on $S116_END_RE above, never on a hand-picked
+#                next-heading literal (#644).
+#
+# `s116_lever_rows` below satisfies both. `s116_posture_rows` satisfies the
+# terminator rule but still opens on the literal `### 1.9 `, so it collapses to 0 on
+# a §1.9 renumber (measured: 66 -> 0). That is a KNOWN residual, out of scope for
+# #670 and unhardened here on purpose — recorded so the next editor does not read
+# the posture helper as already conforming. §156j is held to AGREEMENT with these,
+# not ahead of them, and §156v reds if the two stop agreeing.
+
 # NEITHER END IS A SECTION NUMBER (#668). The window used to open on the literal
 # `### 1.8 ` and close on the literal `### 1.9 `, so it broke in both directions:
 # a `### 1.8.5` sibling sat INSIDE the window and inflated the count (6 -> 7),
--- a/scripts/test/smoke.d/70-gates-contentlocks.sh
+++ b/scripts/test/smoke.d/70-gates-contentlocks.sh
@@ -2823,9 +2823,9 @@ fi
 if [ ! -f "$S156_SPEC" ]; then
   ng "156j: SPEC.md absent — cannot assert §1.8/§1.9 SSOT-change-sweep parity (#640)"
 else
-  s156_lever=$(awk '/^### 1\.8 /{i=1;next} /^### 1\.9 /{exit} i' "$S156_SPEC" \
+  s156_lever=$(awk '/^### .*[Ii]n-session narrowing levers/{i=1;next} i&&/^## |^### /{exit} i' "$S156_SPEC" \
                | grep -cE '^\| \*\*SSOT change sweep\*\*')
-  s156_posture=$(awk '/^### 1\.9 /{i=1;next} /^## 2\. /{exit} i' "$S156_SPEC" \
+  s156_posture=$(awk '/^### .*Harness-overlap classification/{i=1;next} i&&/^## |^### /{exit} i' "$S156_SPEC" \
                  | grep -E '^\|.*SSOT change sweep' \
                  | grep -cE '`(cede-to-harness|keep-as-policy|keep-as-safety-redundancy)`')
   # §156j: both new rows exist AND match the exact shapes §116 counts — a lever row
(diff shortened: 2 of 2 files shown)


ilgyu-yi (commented):

`code-reviewer` (worktree-isolated, artifact-only) reviewed PR #675 at the head below and returned **ship**. The head was blind-compared to the privately-held PR head before this post.

Summary of what was verified empirically at the head: §156j's hardened lever and posture SPEC windows open on the section title (not its number) and close on the shared `^## |^### ` terminator, still counting exactly 1 lever + 1 posture row on the unmodified SPEC (the §156y / §156j bound preserved); the four enforcing arms §156v–§156y red against pre-fix §156j for the right reason (window collapse, not UNTESTED) and green after, with the source-lift confirmed not to call the live definitions (a failed lift reports UNTESTED, never the leaked healthy value); the rebased fixture renumber targets (1.42/1.43) are collision-free against current SPEC and keep the fail-loud guard. Doc→Test→Code phase discipline holds; scope is the two smoke files only (the 50-perproject change is a comment-only contract statement). Test-only change, no adopter-observable surface — skip-changelog per §18.7. No blockers, no fix-needed findings (three advisory nits only: the pre-rebase SHA citations, now corrected in the body; a cosmetic position-3-vs-4 comment wording; illustrative measured figures in comments).

PR #685, opened 2026-08-21
fix(#642): exclude .claude/worktrees/ from smoke grep -r scans

### LINKED ISSUE #642: smoke false-positives when a worktree-isolated subagent is running

ilgyu-yi:
Closes #642

## Goal

A smoke gate must not flip on whether an unrelated worktree-isolated subagent is alive. §148i-cli (and the same-shaped `grep -r` siblings) recurse over the whole `$SHELL_ROOT/.claude` tree, which now includes `.claude/worktrees/agent-*/` full repo copies — so a file-counting assertion double-counts while a subagent runs, reddening a clean suite (observed on PR #641's own /ship review step).

## Plan

`fix`, reproduce-first: the failing arm IS the reproduction (Test commit lands before Code). No planner contest — the change is smoke-internal, under 3 files, no schema/API/contract surface (norm: planner required at 3+ files or schema/API/migration). No Doc phase — no SSOT clause governs smoke internals (the issue confirms this); `Docs: n/a — smoke-internal test-harness scoping`.

- **Test** (`8040824`) — §148i-decoy: a decoy `ghjig_state_dir_cli()` def-form under the real `.claude/worktrees/agent-decoy/...` must be excluded (count stays 1); reds pre-fix at count=2. §148i-neg: a genuine second def OUTSIDE worktrees still counts>1, so the exclusion cannot blind the original check.
- **Code** (`d8b0088`) — a single shared `ghjig_grep_claude_defs` helper in `_preamble.sh` (portable post-filter dropping `/.claude/worktrees/` paths — `--exclude-dir` diverges BSD/GNU), routed through by §148i-cli's detection (AC1). §148i-neg also routes through it so it genuinely guards against an over-broad exclusion pruning real source.

## Out of scope

Changing worktree isolation or relocating `.claude/worktrees/`; auditing every `grep -r` beyond the four named call sites.

## Checklist

- [x] **Test** (`8040824`) — §148i-decoy (red pre-fix) + §148i-neg (paired negative)
- [x] **Code** (`d8b0088`) — shared `ghjig_grep_claude_defs` in `_preamble.sh` (portable `-v '/.claude/worktrees/'` post-filter, not `--exclude-dir`); §148i-cli/decoy/neg routed through it. Suite pass=1325 fail=0.

## Ship gate

- [x] `code-reviewer` clean at head — ship at d8b0088 (head blind-compare matched)
- [x] full smoke green in a worktree-free window — pass=1325 fail=0 at d8b0088
- [~] CI green — N/A at ready; verified at merge gate
- [x] AC closeout posted on #642

### DIFF
--- a/scripts/test/smoke.d/70-gates-contentlocks.sh
+++ b/scripts/test/smoke.d/70-gates-contentlocks.sh
@@ -938,6 +938,53 @@ else
   ng "148i-cli: ghjig_state_dir_cli must be a single hookrt.sh definition the wrapper calls, never re-defines (count=$s148i_defcount hookrt=$s148i_in_hookrt redef=$s148i_wrap_redef calls=$s148i_wrap_calls) (#602, #633)"
 fi
 
+# §148i-decoy (LOAD-BEARING RED until Code, #642): §148i-cli's tree-wide def-form scan greps the
+# WHOLE $SHELL_ROOT/.claude tree, which includes .claude/worktrees/agent-*/ — full repo COPIES a
+# worktree-isolated subagent makes (SPEC §1.5, §4.5). A copy of hookrt.sh there carries a genuine
+# `ghjig_state_dir_cli()` def-form line, so while any such subagent is live the naive `grep -r`
+# double-counts and §148i-cli reds — count=2 though the source tree is fine. This arm plants
+# exactly that decoy at the real worktrees path, re-runs §148i-cli's OWN detection expression
+# verbatim, and asserts it STILL counts exactly 1: the worktree copy must be EXCLUDED. Pre-fix
+# (no exclusion) it counts 2 and reds; Phase C's shared worktrees-exclusion greens it. Anti-
+# vacuity: the decoy must actually exist (made=1) and a real non-worktree hookrt.sh def must still
+# be found (realfound=1), so a broken fixture reds as untested rather than passing on count=1.
+S148I_DECOY_WT="$SHELL_ROOT/.claude/worktrees/agent-decoy"
+S148I_DECOY_F="$S148I_DECOY_WT/.claude/hooks/hookrt.sh"
+mkdir -p "$S148I_DECOY_WT/.claude/hooks"
+printf '%s\n' 'ghjig_state_dir_cli() { :; }' > "$S148I_DECOY_F" 2>/dev/null
+s148i_decoy_made=0; [ -f "$S148I_DECOY_F" ] && s148i_decoy_made=1
+s148i_decoy_defs=$(grep -rlE '^[[:space:]]*ghjig_state_dir_cli[[:space:]]*\(\)' "$SHELL_ROOT/.claude" "$SHELL_ROOT/scripts" 2>/dev/null | sort -u)
+s148i_decoy_count=$(printf '%s\n' "$s148i_decoy_defs" | grep -c . )
+s148i_decoy_realfound=0
+printf '%s\n' "$s148i_decoy_defs" | grep 'hookrt\.sh$' | grep -qv '/worktrees/' && s148i_decoy_realfound=1
+rm -rf "$S148I_DECOY_WT"
+if [ "$s148i_decoy_made" = 1 ] && [ "$s148i_decoy_realfound" = 1 ] && [ "$s148i_decoy_count" = 1 ]; then
+  ok "148i-decoy: §148i-cli's def-form scan excludes .claude/worktrees/agent-*/ copies — a live worktree-isolated subagent can no longer false-red the single-def lock (count=$s148i_decoy_count) (#642)"
+else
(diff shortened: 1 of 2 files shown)


ilgyu-yi (commented):

`code-reviewer` (worktree-isolated, artifact-only) reviewed PR #685 at the head below and returned **ship**. The head was blind-compared to the privately-held PR head before this post.

Verified at the head: the shared `ghjig_grep_claude_defs` helper in `_preamble.sh` post-filters `/.claude/worktrees/` paths with a portable `grep -v` (component-scoped by the trailing slash, so it prunes only worktree copies and never legitimate source; no `--exclude-dir`, portable BSD/GNU); §148i-cli routes through it and still catches a genuine second def outside worktrees (not blinded — §148i-neg is the standing proof); §148i-decoy is anti-vacuous (greens only on made=1 ∧ realfound=1 ∧ count=1; the Test commit's real count=2 RED confirms a genuine reproduction, not a crash); §148i-neg's $TMP fixture cannot self-prune. `fix` reproduce-first, Doc n/a (no SSOT clause governs smoke internals), test-only so skip-changelog per §18.7. Full suite pass=1325 fail=0. No blockers, no fix-needed findings.

Issue #683, opened 2026-08-20
the clock-over-length guard in escape.sh is load-bearing against a sign-dependent fall-open and no smoke arm locks it


ilgyu-yi:
## What

`_escape_token_honored`'s clock validation (`.claude/hooks/helpers/escape.sh:163-165`) guards `now=$(date +%s)` with three arms: `clock-malformed` (empty / leading-zero / non-digit), `clock-over-length` (`[ "${#now}" -le 11 ]`), then the TTL comparison. The over-length arm is **load-bearing**: with it removed, an all-digit `now` longer than 11 digits reaches `if [ "$t_created" -gt "$now" ] || [ "$(( now - t_created ))" -gt 60 ]`, where `[` rejects the non-integer operand **non-fatally** (rc=2, control crosses the `||` into the arithmetic) and the arithmetic **wraps sign-dependently**.

Measured at `a8c4b77` — a mirror of the :165 comparison with the :164 length guard removed, script-file mode, `set -uo pipefail`, `t_created=1755640000` (a past-TTL epoch):

```
$ bash clock-overlen-probe.sh
[: 99999999999999999999: integer expression expected
now=99999999999999999999 len=20 delta=7766279629696601919  -> BLOCK (stale-or-future-dated)
[: 10000000000000000000: integer expression expected
now=10000000000000000000 len=20 delta=-8446744075465191616 -> HONORED - FALL-OPEN
[: 18446744073709551617: integer expression expected
now=18446744073709551617 len=20 delta=-1755639999          -> HONORED - FALL-OPEN
```

The wrap is **sign-dependent**: the `1e20-1` class lands positive and blocks anyway, while the `1e19` and `2^64+1` classes land negative and honor a past-TTL token. A fixture built on the wrong member greens while measuring nothing.

**Nothing locks the arm.** `scripts/test/smoke.d/60-escape-identity.sh` mutation-locks the `created` twin — `§125-10` covers leading-zero octal and `>=2^63` overflow — and the clock's **empty** (`§125-11`) and **leading-zero** (`§125-11b`) arms, with `§125-11-ok` as the positive control. No arm feeds an over-length all-digit clock reading, so deleting or loosening `escape.sh:164` leaves the full suite green.

## Why

[...]

Issue #656, opened 2026-08-15
relocate escape.sh's TTL rationale to SPEC, correcting four measured prose defects as the payload


ilgyu-yi:
## What

`_escape_token_honored` in `.claude/hooks/helpers/escape.sh` carries a hand-maintained rationale block — a measurement-mode declaration, a three-signature fall-open taxonomy, and a threat model — several times the length of the guard it documents. Relocate the durable contract to SPEC and leave a pointer, and correct four measured prose defects as the payload of that move rather than as four more clause patches.

## Why this is a relocation and not a fourth round of edits

PR #651 corrected this block in rounds 3, 4 and 5 — three correction commits (`407bddd`, `97bc1ea`, `f9fa2dd`) after the initial guard, and **none in round 6**, which the table's empty round-6 repair cell records. *(An earlier revision said "rounds 3, 4, 5 and 6" and then added the very table row that contradicts it.)* **Each correction introduced or left a new false claim at the same axis-key:**

| round | the repair | what it left wrong |
|---|---|---|
| 3 | merged the hex row into the empty/non-numeric row | listed `abc` as sharing that signature; under `set -u` `abc` exits 2 (the hook's block signal), so it is not a fall-open at all |
| 4 | corrected the `abc` row | asserted the abort "stops at the enclosing `if`, so sibling arms still run"; measured, it unwinds to the next **top-level** statement and `commit-format` never fires |
| 5 | corrected the abort scope | grouped `0<epoch>` and `%s` in one row and attached a **wrapper** consequence true only of the octal member |
| 6 | — | the round-5 clause is the LOW-1 finding below |

Three of the four are defects *in the previous round's repair*. `finding-judge` ruled at round 6: *"A 55-line hand-maintained taxonomy embedded in a helper body cannot be kept true by patching, and five rounds are the measurement."*

There is a structural reason. A helper-body comment reaches `main` through a **code-file edit**, which no SPEC-edit review path guards. SPEC §7 already owns the normative TTL contract and §5.29 the arm inventory; the taxonomy is the one piece of durable doctrine living where nothing reviews it as doctrine.

**MISSION fit:** MISSION "The mechanism" — durable doctrine parked in a working surface is a misplaced archive: it grows the context every reader of that function must page through, while sitting where no doc-review path inspects it. SPEC §9's reference-don't-restate discipline is the same principle applied to documents; this Issue applies it to a helper body.

[...]

PR #689, opened 2026-08-21
fix(#655): complete P4 arm classification + harden the §148h-p4 lock

### LINKED ISSUE #655: §6.0 P4's arm classification is incomplete and §148h-p4 locks a weaker rule than it claims
## What

SPEC §6.0 P4's arm-scoped recovery obligation sorts reject arms into two buckets — honest-mistake arms carry a recovery clause, hostile-input arms stay deliberately terse. Measured against `scripts/ghjig_file_review_post.sh`, **three arms with a message surface fall into neither bucket**, and the smoke lock that enforces the rule (`§148h-p4`) is weaker than its own comment claims, on three independent limbs.

> **Citation convention (revised after activation round 1).** Everything below is cited by **stable anchor** — SPEC section number, arm name, smoke section number — never by line number. The first draft of this issue cited `SPEC.md:1826` and `SPEC.md:1862`; PR #651 and PR #657 merged after filing and shifted both **+8**, so AC5's stated reproduction would have mutated the wrong paragraph *and still gone green*. That is the false-confirming measurement this issue exists to complain about, committed by the issue itself. The repo's own rule is `.claude/agents/finding-judge.md`: *"**Never a line number** — lines move."*

## Measured (at `2d6bf1e`)

`scripts/ghjig_file_review_post.sh` has **20 `deny` call sites over 17 distinct arm names**; `mtime-unresolvable`, `mtime-malformed` and `now-malformed` each appear **twice**. Classified by whether the message carries an em-dash recovery clause:

| arm | recovery clause? |
|---|---|
| `symlink-dir`, `symlink-leaf` | bare — **exempt by contract** (hostile input) |
| **`staging-absent`** | **bare, and in neither bucket** |
| `staging-irregular`, `staging-unreadable`, `staging-empty` | yes |
| **`mtime-unresolvable`** (both sites) | **bare, and in neither bucket** |
| `mtime-changed` | em dash present — but see limb 2 |
| `mtime-malformed` | first site yes; **over-length twin bare** |
| `now-malformed` | first site yes; **over-length twin bare** |
| **`mtime-future`** | **bare, and in neither bucket** |
| `stale` | yes (the shape template) |
| `marker-*`, `local-head-*` | yes |

SPEC §5.29 names exactly four honest-mistake arms (`staging-irregular`, `staging-unreadable`, `staging-empty`, `now-malformed`) plus `stale` as the template. `§148h-p4` checks exactly the same five.

### The three unclassified arms are reachable by accident, not by hostility

- **`staging-absent`** — the caller wrote the body to a different path than the wrapper reads. `.claude/commands/file-review.md` devotes a paragraph to that divergence, so it is an anticipated operator error, not an attack.
- **`mtime-unresolvable`** (both sites) — "could not read the staging file mtime on this platform", i.e. a platform with neither `stat -c` nor `stat -f`. An environment condition with no adversary at all — **the same argument #647 used to admit `now-malformed` into the honest-mistake set.**
- **`mtime-future`** — clock skew.

Each has a message surface (unlike `.claude/hooks/helpers/escape.sh`'s arms, which return a bare status code and whose exemption §6.0 P4 states explicitly). So the P4 bound does not cover them, and neither does the hostile-input exemption.

### Limb 1 — the lock is per-arm-NAME, not per-call-site

`§148h-p4` greps `deny <arm> ` and requires **one** matching line to carry an em dash. `mtime-malformed` and `now-malformed` each have one clause-bearing site and one bare site, so each bare twin is covered by its sibling and the lock is satisfied.

### Limb 2 — the lock tests em-dash presence, not the shape its comment claims

`§148h-p4`'s case comment says the required shape is "an em dash followed by an **imperative**". The predicate is `… | grep -q '—'`. `deny mtime-changed` reads *"staging file mtime changed during the read … — the freshness check would not cover the posted bytes"*: an **explanation**, not a recovery. It says why the arm fired, not what to do. The predicate cannot discriminate the two. **No false green today**, because `mtime-changed` is not in the checked set; the point is that the predicate would not catch it if it were.

### Limb 3 — the inventory is fixed

`§148h-p4` hard-codes five arm names. A newly added honest-mistake arm carries no obligation until someone remembers to extend the list — the same drift shape #647 repaired in the §5.29 inventory by replacing a count with names.

### Limb 4 — §6.0 P4's normative paragraph is locked by nothing

`grep -rn 'P4' scripts/test/smoke.d/` returns **4 hits, all inside comments**. No assertion reads P4's prose, so the paragraph can be reworded — or reversed — without a single arm going red. This belongs here rather than in its own issue **because unlocked P4 prose is the mechanism by which the classification drifts unnoticed**: it is the same defect seen from the enforcement side. Raised and dropped in three consecutive review rounds on PR #651; filing it is what stops a fourth raise.

## Why

**This is observability and doctrine consistency, not a security hole.** Every arm listed fails closed; nothing here can produce a wrong *allow*. The cost of being wrong is an operator who hits an environment condition and gets a bare refusal with no named next step — precisely the "block that only says no is half a rule" case §6.0 P4 exists to prevent.

**Why now.** #647 generalized P4 from gate-scoped to arm-scoped and shipped the bound that exempts message-less arms. That change is what makes the three unclassified arms visible: before it, "the gate names one alternative somewhere" covered them. The rule is now precise enough for the gap to be a gap.

**MISSION fit:** *The mechanism* — "Enforcement is the action-side counterpart of this dual." A negative gate that names no alternative is the one-sided regression MISSION warns about, and P4 is that principle's arm-level form.

**Existing coverage:** no duplicate, and two adjacencies that must be named.

- **#656 shares an edit site.** #656's LOW-2 edits the SPEC §5.29 parenthetical about `mtime-malformed`; this issue's AC1/AC2 edit the classification in **that same paragraph**. The activation reviewer confirmed the two issues are genuinely distinct work — #655 owns the wrapper's arm classification and the `§148h-p4` lock, #656 owns relocating `escape.sh`'s rationale plus four TTL-taxonomy prose corrections, and neither subsumes the other — but **whichever lands second must reconcile that paragraph.** The first draft of this body said "Existing coverage: none", which was already stale when written: #656 was filed four minutes later and links back here.
- **#647 / PR #651** own the clock guards and explicitly scoped this out across two judged rounds (round 4 D1, round 5 D-c), ruling "file it; do not fix it here". #647's own AC widened `§148h-fc`'s `head -1` to every matching line; AC3 below asks a related change to the sibling `§148h-p4` limb in the same `case` block. Not a duplicate — `§148h-fc` checks arm *presence*, `§148h-p4` checks the recovery *clause* — but the same edit shape in the same place.
- **#653** covers audit silence in `escape.sh`'s reject arms — a different file, and about records rather than messages.

## Acceptance criteria

- [ ] **AC1** — SPEC §6.0 P4 (or §5.29) classifies `staging-absent`, `mtime-unresolvable` and `mtime-future` explicitly: either they gain a recovery clause, or the doctrine gains a stated third category with the reason they belong to it. Silence is not an outcome. **Both branches are pinned, not just the cheaper one.** Branch 1 is caught by AC2's derived source feeding `§148h-p4`. Branch 2 lands as prose, so it ships with its own phrase lock naming the third category, verified by mutation: reword or delete the category sentence and an arm must RED. *(Round-2 activation finding: as first written, branch 2 could be discharged by prose a later reword silently deletes with no arm going red — which is limb 4's own thesis applied to this issue's headline remedy, and the same unpinned-cheap-arm shape round 1 required closed at AC3 and AC4.)*
- [ ] **AC2** — the §5.29 honest-mistake set and `§148h-p4`'s checked list are derived from **one** source rather than hard-coded twice, so they cannot drift apart. Verified by mutation: adding a name to the single source must change what the lock checks, with no second edit.
- [ ] **AC3** — the lock's per-arm-name weakness is closed, and the over-length twins are **exempted in the SSOT, not only in test code**. PR #651 round 1's F1 excluded the twins from the obligation and rounds 3, 4 and 5 each declined to re-raise it; `now-malformed` is in the checked five and its twin is bare, so a naive "check every call site" rule REDs the real script today. Therefore: (a) the exemption is stated in SPEC §5.29 or §6.0 P4 where a future reader of the contract will find it — recording it only in a smoke comment is not sufficient, because that is the same "narrow the rule where nobody reading the rule can see it" failure this issue is about; and (b) a **fixture** (not the live script) carrying one clause-bearing and one bare site for a *non-exempt* arm must RED. This AC is discharged by implementing (a) and (b); re-opening round 1's ceiling instead is a separate decision that does not discharge it.
- [ ] **AC4** — either the case comment is corrected to state what the predicate actually tests (em-dash presence), or the predicate is strengthened to discriminate a recovery from an explanation. **Whichever is chosen is pinned by a fixture:** if the predicate is strengthened, `deny mtime-changed`'s current explanation-only message must RED against it; if the comment is corrected instead, an assertion pins the corrected comment against the predicate it describes, so the two cannot drift apart again.
- [ ] **AC5** — **§6.0 P4's normative prose is locked** by a phrase lock in the shape `§148h-ttl2` already uses. **Name the target unambiguously**: PR #659 added a third paragraph to P4 (the shared-helper *calling surface* rule), so "the sentence" no longer identifies one thing. The lock pins the sentence that states the **honest-mistake / hostile-input classification rule** specifically, and if AC1 takes branch 2 the third-category sentence is pinned as well. Measured today: `grep -rn 'P4' scripts/test/smoke.d/` returns only comments, and greps for `never ship a bare gate`, `half a rule`, `bounded to arms`, `Pair the faces` and `calling surface` return **zero** locks — so #659's own new paragraph is unlocked too. Verified by mutation: replacing that sentence with one asserting the opposite ("gates may ship bare") must RED. Today it does not — `grep -rn 'P4' scripts/test/smoke.d/` returns 4 hits, all comments. This AC exists because AC2 **cannot** cover it: AC2 derives the *arm inventory* from one source, and P4 deliberately does not restate the inventory (#647 removed those names from P4 on purpose), so no inventory-derived check can see a P4 mutation.
- [ ] **AC6** — the two symlink arms stay bare and stay exempt. A fixture asserts they are not swept into the obligation.
- [ ] **AC7** — the full smoke suite passes in a window with **no live agent worktree** (see #642 for why that qualifier is load-bearing; a live worktree reds `§148i-cli` for unrelated reasons).

## Out of scope

- **The clock guards themselves, and anything in #647's AC set.** This Issue changes no guard behaviour and adds no reject arm.
- **`escape.sh`'s reject arms.** They have no message surface at all — §6.0 P4 states that exemption explicitly — and their audit silence is #653.
- **Adding a `now-unresolvable` arm** (or any new arm) to the wrapper. PR #651's round-5 judgment refused that remedy because it would change a fail-closed arm set whose security review is keyed to the current inventory. Same reasoning applies here.
- **Rewording arms that already carry a recovery clause.**

ilgyu-yi:
Closes #655

## Goal

SPEC §6.0 P4 sorts `/file-review` wrapper reject arms into honest-mistake (em-dash recovery clause) vs hostile-input (terse, exempt). Three arms with a message surface — `staging-absent`, `mtime-unresolvable` (both sites), `mtime-future` — fell into neither bucket, and the `§148h-p4` lock that enforces P4 was weaker than its comment on four limbs (per-arm-name not per-site; em-dash presence not "recovery"; hardcoded 5-name list; P4 prose locked by nothing). All arms already fail closed — this is observability/doctrine, not a wrong-allow.

## Plan

Doc → Test → Code STRICT (contract + lock change). Full planner contest ran (design latitude: AC1 branch + single-source mechanism).

- **Doc** (`71a16b0`) — SPEC §5.29 single-source honest-mistake enumeration ADDS the three arms (Branch 1, #647 `now-malformed` precedent); a separate anchored sentence exempts the over-length twins (kept clear of the `mtime-malformed` parenthetical #656 owns); the recovery paragraph references the enumeration (one source, not two). §6.0 P4 already carries the AC5 lock target strings.
- **Test** (`bb54a60`) — `§148h-p4` reworked: derives the checked set by parsing §5.29 (no hardcoded list), iterates per call site, twin-exemption via a parsed sentence (≥1 site), AC4(a) comment corrected to em-dash presence + comment↔predicate self-pin, AC5 phrase-lock, and **B1's anti-vacuity sentinel floor** (`stale`+`now-malformed`, distinct `<…degenerate>` token) so a de-backticked enumeration cannot vacuously GREEN. Fixtures: AC3b (per-site RED), AC6 (symlink GREEN), B1 vacuity (floor RED). RED pre-fix naming the three bare arms.
- **Code** (`1bd6b56`) — the three arms (4 sites) gain an em-dash recovery clause naming the live fix, in the `stale`-arm shape. Message-string-only; no control-flow/keying change; twins + `mtime-changed` untouched.

## Alternatives considered

Contest {A, B1, B2}, axes {robustness/risk, simplicity/maintainability}, plan-reviewer judged.
- **A** (base): Branch 1 recovery clauses + single-source §5.29 parse + per-site + AC4(a) + AC5 lock.
- **B1** (robustness): **narrow dominates-A** — A's single-source parse could vacuously GREEN on a de-backticked/empty derivation (§148h-doc stays safe only via its two-source hardcoded floor). Graft: an anti-vacuity sentinel floor + vacuity fixture. **Carried.**
- **B2** (simplicity): non-dominating — fold the two §5.29 sentences into one; rejected (collides more with #656's parenthetical, Inv 6; A's two-sentence form reads cleaner). Conceded AC4(a) + strict phases.
- **Verdict**: `refine: carry B1 forward` → the graft is in the Test commit. B2 rejected.

## Out of scope

The clock guards themselves / #647's AC set (no guard-behavior change, no new arm); `escape.sh`'s status-code reject arms and its TTL relocation (#656) / clock-over-length lock (#683); rewording arms that already carry a clause (incl. `mtime-changed`); #659's shared-helper calling-surface paragraph (not the AC5 target).

## Checklist

- [x] **Doc** (`71a16b0`) — §5.29 single-source honest-mistake enumeration + twin-exemption; P4 AC5 target present
- [x] **Test** (`bb54a60`) — §148h-p4 rework (derived-set, per-site, twin-exempt, AC4 comment-pin, AC5 lock, B1 floor) + AC3b/AC6/vacuity fixtures; RED pre-fix
- [x] **Code** (`1bd6b56`) — 3 arms / 4 sites gain the recovery clause; §148h-p4 + siblings GREEN, suite pass=1334 fail=0

## Ship gate

- [x] `code-reviewer` clean at head — ship at 62c9c0d
- [x] full smoke green in a worktree-free window — pass=1334 fail=0 (#642)
- [x] security-reviewer: approve (message-string-only; merge-review producer surface unchanged, no injection)
- [~] CI green — N/A at ready; verified at merge gate
- [x] AC closeout posted on #655


ilgyu-yi (commented):

`code-reviewer` reviewed PR #689 at the head below and returned **ship** (confirm-the-delta re-review after the fixture shellcheck-disable fix; the substance — wrapper message-string edits, SPEC §5.29/§6.0 P4 doctrine, the §148h-p4 rework — was approved at 62c9c0d by code-reviewer ship + security-reviewer approve and is byte-unchanged). Head blind-compared to the privately-held PR head before this post.

Delta since 62c9c0d is exactly two `# shellcheck disable=SC2154,SC2034 # <reason>` comment lines in the two wrapper-fragment fixtures (`ac3b-partial-clause.sh`, `ac6-symlink-bare.sh`) — scoped, reason verified accurate ($sf/$frdir/mt1/mt2 are assigned by the real wrapper; §148h-p4 greps the fixtures as text and never executes them). §148h-p4 greps `$S148_WRAP_FILE` (the real wrapper), not the fixtures, and the added comment carries no `deny`/em-dash, so the arm's counts cannot move — re-verified green locally (§148h-p4 + -vac/-ac3b/-ac6/-cmt/-p4lock, suite pass=1334 fail=0). lint.sh clean over the fixtures. Security surface unchanged (message-string-only wrapper edits, confirmed at 62c9c0d). CI green at head. No blockers.

PR #693, opened 2026-08-21
docs(#656): relocate the TTL fall-open taxonomy to SPEC §7.1, correct four claims

### LINKED ISSUE #656: relocate escape.sh's TTL rationale to SPEC, correcting four measured prose defects as the payload

ilgyu-yi:
Closes #656

## Goal

Relocate `escape.sh`'s hand-maintained TTL fall-open rationale into SPEC, and land four measured prose corrections as the payload of that move rather than as a fifth round of in-place clause patches.

MISSION "The mechanism". A 55-line taxonomy of durable doctrine parked in a helper body is a misplaced archive: it grows the context every reader of `_escape_token_honored` must page through, while sitting where no doc-review path inspects it as doctrine. PR #651 patched it in rounds 3, 4 and 5, and **each repair introduced or left a new false claim at the same axis-key** — the measurement that says patching cannot keep it true. SPEC §9's reference-don't-restate discipline is this principle applied to documents; this PR applies it to a helper body.

## Plan

- [x] **Doc** (`667ff95`) — SPEC §7.1 created; `escape.sh` block → pointer + 3 shape-names; the LOW-2 parenthetical gains the over-length limb in SPEC §5.29 + `file-review.md`; the CR-minor-3 `hookrt.sh` caller-comment corrected; ToC regenerated.
- [x] **Test** (`50fb84a`) — smoke `§148h-ttl3` in `70-gates-contentlocks.sh`: unique `### 7.1` heading anchor (fail-closed `<ttl3-anchor-absent>` / `<ttl3-anchor-not-unique>`), awk-cut scope to the §7.1 region only, four named substrings (the LOW-1 pair + the INFO-4 pair), name-which-reddens diagnostics.
- [x] **Chore** (`82fe3f5`) — changelog fragment `changelog_unreleased/changed/693.md` (§18.1).
- [x] **Review fixes** (`407e71c`) — round-1 judged list, F1–F4 (below).
- [x] **Review fixes** (`ef7b37b`) — round-3 judged list, pure subtraction: SPEC §7.1 bullet 2 keeps only its operand-naming sentence (the judged floor; the hex-`created` clause the round-2 repair added was measured false at `created=0x689b2cb6`, a hex-rendered epoch inside the TTL window, which is HONORED), and the `§148h-ttl3` INFO-4 comment line now mirrors the LOW-1 line — names the pinned literals, points at §7.1 as owner, glosses nothing.
- [x] **Review fixes** (`7ffc582`) — round-2 judged list: the `§148h-ttl3` comment points at §7.1 instead of restating it, SPEC §7.1 bullet 2 names its operand, the changelog fragment describes the correction rather than quoting the superseded claim.
- **Code phase: n/a** — prose-only relocation. Doc→Test→Code is relaxed here per CLAUDE.md; the invariant that replaces it is stated and re-verified under *Test plan*: no functional line changed.

## Alternatives considered

- **A fifth in-place patch of the `escape.sh` comment block** (the base plan's null alternative). Rejected on the measurement in *Goal*: three of the four defects being corrected here are defects *in a previous round's repair*. `finding-judge` ruled at PR #651 round 6 that "a 55-line hand-maintained taxonomy embedded in a helper body cannot be kept true by patching, and five rounds are the measurement."
- **Relocate but leave the four claims as-is, filing them separately.** Rejected: the corrections are the payload that justifies the move, and splitting them would land a relocation whose text is known-false on arrival.
- **Plan contest** — base Plan A vs two `plan-challenger` rivals on distinct axes. `plan-reviewer` verdict: `refine → A+B1+B2` (composed winner). Both challenger mandates are in the shipped plan: the INFO-4 lock alongside LOW-1 (not LOW-1 alone), and the anti-vacuity anchor discipline (a unique heading anchor failing closed, rather than a bare whole-file `grep -qF`).

## Key context

Prose-only: no functional line changed, no guard arm or `125-*`/`148j-*` lock touched; #655's twin-exemption/enumeration sentences are byte-untouched; #683's cross-operand analysis is deliberately **not** absorbed (that is #683's scope).

### The four corrections, reproduced

**LOW-1 (symptom reach set).** Not "every leading-zero epoch" falls open — within the leading-zero shapes the reaching set is exactly those **containing an 8 or 9**, because `[` accepts the operand as a base-10 string where `$(( ))` rejects it as invalid octal:

```
$ bash -c '[ 0888888888 -gt 0 ] && echo "[ ok: 0888888888"'
[ ok: 0888888888
$ bash -c '[ 0177777777 -gt 0 ] && echo "[ ok: 0177777777"'
[ ok: 0177777777
$ bash -c 'echo $((1755000000 - 0888888888))'
bash: 1755000000 - 0888888888: value too great for base (error token is "0888888888")
$ bash -c 'echo "arith ok: $((1755000000 - 0177777777))"'
arith ok: 1721445569
```

The split-vs-join is a **separate** mechanism, and it separates the two *sites* rather than narrowing the reach set: the `/file-review` wrapper splits the two comparisons into two `|| deny` statements (catching a `[`-rejecting operand one statement early), while `escape.sh` joins them inside one `if` (so the same operand falls through to the arithmetic). That is why `%s` shares the row at `escape.sh` but denies under `mtime-future` at the wrapper. *(The first-round text welded these two into one `because`; round-1 finding F3 measured the connector false and it was rewritten in `407e71c`.)*

**INFO-4 (measurement mode).** The `bash -c` framing that cleared this class twice (#635) named the wrong reason. The discriminator is per-line parsing, not `bash -c` versus a script file: a one-line `bash -c` dies as a single parse unit (rc=1, no POST — the SAFE-looking answer that misled), a multi-line `bash -c` abandons only the current compound and runs the next top-level line to the POST (identical to a script file):

```
$ bash -c 'set -uo pipefail; x=$((1755000000 - 0888888888)); echo POST'
bash: 1755000000 - 0888888888: value too great for base (error token is "0888888888")
rc=1   # one-line: no POST
$ bash -c $'set -uo pipefail\nx=$((1755000000 - 0888888888))\necho POST'
bash: line 1: 1755000000 - 0888888888: value too great for base (error token is "0888888888")
POST
rc=0   # multi-line: POST reached
```

**LOW-2 (over-length limb).** The `mtime-malformed`/`now-malformed` parenthetical in SPEC §5.29 and `file-review.md` omitted the over-length reject the code enforces; both now read `…, or longer than 11 digits`, matching the `-le 11` bound:

```
$ grep -n 'le 11' .claude/hooks/helpers/escape.sh
120:  [ "${#t_created}" -le 11 ] || { _escape_reject "$cat" "$tok" "created-over-length"; return 1; }
125:  [ "${#now}" -le 11 ] || { _escape_reject "$cat" "$tok" "clock-over-length"; return 1; }
$ grep -rc 'or longer than 11 digits' SPEC.md .claude/commands/file-review.md
SPEC.md:1
.claude/commands/file-review.md:1
```

**CR-minor-3 (caller comment).** `hookrt.sh`'s resolver comment claimed the review-post wrapper was its "ONLY /file-review-side caller". The claim is now stated at the extent its own search returns — **only script caller on the runtime path**, with the test caller named:

```
$ git grep -l 'ghjig_state_dir_cli' HEAD -- scripts .claude/hooks .claude/commands
HEAD:.claude/commands/file-review.md
HEAD:.claude/hooks/hookrt.sh
HEAD:scripts/ghjig_file_review_post.sh
HEAD:scripts/test/smoke.d/70-gates-contentlocks.sh
```

Four paths: the definition site, one **doc-prescribed agent invocation** (`file-review.md:79` instructs the agent to source `hookrt.sh` and call the resolver itself — an executed call, not a passive mention), one runtime script caller (`ghjig_file_review_post.sh`), and one **test** caller — smoke `§148i-clibranch`, which sources `hookrt.sh` and calls `ghjig_state_dir_cli` to pin this resolver's live branch. *(The first-round body pinned the equivalent `grep -rln` and enumerated three of the four — SPEC §1.10(b)'s stated-count-does-not-reproduce fixture. Round-1 finding F2; corrected in `407e71c`, and the comment narrowed to what is true rather than the count refreshed alone.)*

## Checklist

- [x] SPEC §7.1 carries the relocated taxonomy; `escape.sh` retains a pointer plus the three shape-names only (AC1)
- [x] All four corrections land, each with its reproduction on a durable carrier (AC2)
- [x] The relocated contract is mutation-locked by `§148h-ttl3` (AC3)
- [x] Full suite run in a window with no live agent worktree (AC4 — see *Test plan*)
- [x] ToC regenerated (`scripts/build_toc.sh --check` → rc 0)

## Out of scope

- **The cross-operand over-length fall-open and its missing smoke lock** — #683 owns the load-bearing analysis and the lock; this PR carries only the two prose parentheticals.
- **The §6.0 P4 arm-classification defect and the `§148h-p4` lock** — #655. The two collide on SPEC §5.29's `malformed` parenthetical; whichever lands second reconciles it.
- **Any change to the guards themselves.** The functional lines are correct at both sites and mutation-locked (`§125-10`, `§125-11`, `§125-11b`, `§148j-oct-mt`, `§148j-oct-now`).
- **The escape-token channel's audit silence** — #653.
- **The repo-wide `MEASURED`-label rule** — deferred; it is unscoped to `escape.sh` and has no gate behind it.

## Risks

- **A relocation can silently drop a claim.** Mitigated by `§148h-ttl3`, which pins two load-bearing sentences and fails closed on a missing or duplicated anchor.
- **A restatement drifts from the SSOT it restates.** This is the failure #656 exists to end, and it recurred inside the PR: the `§148h-ttl3` rationale comment glossed the LOW-1 reach set and its cause, and after `407e71c` corrected §7.1 that gloss contradicted the very sentence the arm pins. Round-2 finding F1; the comment now names which literals are pinned and points at §7.1 as their owner, asserting no mechanism of its own.
- **A lock whose region is too wide is a lock in name only.** This was real, not hypothetical: at the first head the `### 7.1` heading sat mid-§7 and the awk region ran 58 lines, so a decoy planted in unrelated §7 prose greened the arm. Round-1 finding F1; the block now sits at the end of §7 and the region is 17 lines.

## Open questions

None outstanding.

## Decisions

No ADR. The relocation is reversible (the text is editable and git-recoverable), so it does not meet the §5.5 irreversible-decision bar.

## Test plan

**The lock (`§148h-ttl3`).** Green at HEAD; red under every probed mutation, naming the offender:

- reword LOW-1 (`8 or 9` → `high digit`) → `missing: <LOW-1:leading-zero-epochs-containing-an-8-or-9>`
- reword INFO-4 (`set -uo pipefail` → `set -euo pipefail-ish`) → `missing: <INFO-4:set--uo-pipefail>`
- heading removed / duplicated → `<ttl3-anchor-absent>` / `<ttl3-anchor-not-unique>`
- phrase deleted from §7.1 and injected as decoy narration in §8 → still red naming LOW-1
- **in-region decoy (the case F1 exposed)** — phrase deleted from the taxonomy bullet and replanted in §7's `Categories:` prose. At the pre-fix head, with the region at 58 lines, this returned **GREEN**. At `7ffc582` (re-verified unchanged at `ef7b37b`), with the region at 17 lines, the identical mutation returns:
  ```
  RED:   148h-ttl3: SPEC §7.1 must carry both the LOW-1 symptom and the INFO-4 measurement-mode clauses (missing: <LOW-1:leading-zero-epochs-containing-an-8-or-9>) (#656)
  ```

**The prose-only invariant, re-verified after the review fixes.** Comment-stripped blobs of both touched guard files are byte-identical to `origin/main`:

```
$ for f in .claude/hooks/helpers/escape.sh .claude/hooks/hookrt.sh; do
    diff <(git show origin/main:"$f" | grep -vE '^[[:space:]]*(#|$)') \
         <(grep -vE '^[[:space:]]*(#|$)' "$f") >/dev/null && echo "$f: IDENTICAL"; done
.claude/hooks/helpers/escape.sh: IDENTICAL
.claude/hooks/hookrt.sh: IDENTICAL
```

**Full suite (AC4).** One run, at `ef7b37b` (the current head), `git worktree list` showing the main checkout only — no live agent worktree:

```
$ bash scripts/test/smoke.sh | tail -1
smoke: pass=1332 fail=1
```

The single failure is **`§174h`, a pre-existing wall-clock bound unrelated to this PR** — it asserts `scripts/lint_citations.sh` finishes 250 spans in under 20s and this machine measures 20s. The file is untouched here (same blob at both ends), and the same arm reds identically on the base:

```
$ git rev-parse origin/main:scripts/lint_citations.sh HEAD:scripts/lint_citations.sh | uniq -c
   2 b51091807a1482818c298a03c37af2e0e246b8c9
$ (cd <detached worktree at origin/main> && bash scripts/test/smoke.sh | tail -1)
smoke: pass=1331 fail=1     # ✗ 174h … spancount=20s/200
```

CI's `smoke` job is green on both `ubuntu-latest` and `macos-latest`, which is the clean-window evidence AC4 asks for. The tight bound is a follow-up candidate, not a defect of this change.

## Docs touched

- `SPEC.md` — new §7.1 (relocated taxonomy, at the end of §7); §5.29 parenthetical gains the over-length limb; ToC regenerated.
- `.claude/commands/file-review.md` — same over-length limb.
- `.claude/hooks/helpers/escape.sh` — rationale block → thin pointer + three shape-names (comments only).
- `.claude/hooks/hookrt.sh` — resolver caller-comment narrowed to the measured extent (comments only).

## Ship gate
- [x] All tests pass — the single full-suite run pinned in *Test plan* (at `ef7b37b`, same commit, same run); its one failure is `§174h`, proven pre-existing and PR-independent there
- [x] code-reviewer — rounds 1–3, each `ship after fix`, each round re-verifying the prior round's fixes discharged by execution before raising its own: round 1 F1–F4 → `407e71c`; round 2 F1–F4 → `7ffc582`; round 3 F1/F3 (subtraction only) → `ef7b37b`, F2 = this body's suite-pin reconciliation. Every round `finding-judge`-triaged, each judged list posted to this PR before the fix.
- [x] (conditional) security-reviewer — n/a: the §4.6 trigger does not fire (both guard-file diffs are comment-only; no auth/input/deps/crypto/IO-boundary change), so the §4.11 `merge-security-surface` tier does not apply either
- [x] Docs in sync
- [x] PR body curated
- [x] CI green — 6/6 pass at `ef7b37b` (fragment-gate, bash -n ×2, smoke ×2, mutation kill-rate)


ilgyu-yi:
## Finding triage (round 1)

`reviewed-head: 82fe3f50a85f28d11c657f43a29ba584eb3457ff`

**dispatch** — findings from one `code-reviewer` invocation (worktree-isolated, single-reviewer path; the §4.11 high-asymmetry tier was evaluated as NOT applicable: the guard-file diff is comment-only, no `docs/ADRs/` file, no force-push in the motion). Judgment from one separate `finding-judge` invocation, worktree-isolated, no shared context. The judge cannot self-observe this limb, so the caller records it.

**prior-round read** — trusted-author comment scan returned zero prior `<!-- finding-judge: … -->` markers; round = 1. Every `swing:` is therefore `not-evaluated`, never a vacuous `none`.

`code-reviewer`'s verdict token is **`ship after fix`** and is untouched — the judge converts no verdict and casts no vote of its own.

### Summary

| # | Verdict | Action | Reviewer's remedy |
|---|---|---|---|
| F1 | confirmed | **fix-now** | survives — relocate §7.1 to the end of §7, regenerate the ToC |
| F2 | confirmed | **fix-now** | survives — narrow to the measured extent, or re-pin at 4 |
| F3 | confirmed | **fix-now** | **does not survive** — the author writes the replacement |
| F4 | confirmed | **fix-now** | survives — delete both parentheticals |
| F5 | confirmed | **fix-now** | survives — restore the template sections, `## Ship gate` at minimum |
| F6 | confirmed | drop | n/a — no change requested, correctly |

Nothing refuted this round. Six confirmed, five to fix, one dropped.

### The records

```
finding: F1 — `### 7.1` inserted mid-§7 mis-nests ~40 lines of unrelated §7 body under it, widening the §148h-ttl3 lock region until an in-region decoy greens it
site: SPEC.md#§7.1-TTL-operand-validation
axis-key: SPEC.md#§7.1-TTL-operand-validation::lock-region-scope
verdict: confirmed
evidence: executed=bash <§148h-ttl3 extracted verbatim from head smoke L1009-1032, SHELL_ROOT pinned to a scratch SPEC copy> mode=script-file matches-artifact-mode=yes result=control GREEN; region measured 58 lines (the awk cut runs heading 2350 -> `## 8.` at 2409, NOT to the end of the taxonomy); the mutation deleting `leading-zero epochs containing an 8 or 9` from the taxonomy bullet (2358) and replanting it as decoy narration at 2386 (in-region `Categories:` prose) => still GREEN. The reviewer's load-bearing claim reproduces exactly.
action: fix-now
remedy-survives: yes
axis: narrow<->wide
target-position: low
justification: Both endpoints measured, independent of any prior failure's direction. FLOOR = 17 lines — the taxonomy block alone; the judge built the reviewer's relocation (block moved to just before the `---` ending §7) and the arm is GREEN at 17 lines, so the floor is satisfiable with all four locked literals intact. CEILING = anything admitting non-taxonomy §7 prose; measured at 58 lines the decoy 28 lines downrange greens the arm. Under the relocation the identical decoy REDs, naming `<LOW-1:leading-zero-epochs-containing-an-8-or-9>`. Constraint the author owns: the ToC row `| §7.1 | ... | 2350 |` (SPEC.md:107) carries a line number — `scripts/build_toc.sh` must be regenerated.
swing: not-evaluated
```

```
finding: F2 — hookrt.sh's "the review-post wrapper is now this resolver's only script caller" is false, and the PR body's pinned grep does not reproduce its stated count (SPEC §1.10(b))
site: .claude/hooks/hookrt.sh#ghjig_state_dir_cli
axis-key: .claude/hooks/hookrt.sh#ghjig_state_dir_cli::census-claim
verdict: confirmed
evidence: executed=`git grep -l 'ghjig_state_dir_cli' 82fe3f5 -- scripts .claude/hooks .claude/commands` mode=static-corpus-query matches-artifact-mode=yes result=FOUR paths (file-review.md, hookrt.sh, ghjig_file_review_post.sh, scripts/test/smoke.d/70-gates-contentlocks.sh); the PR body pins the equivalent `grep -rln` and enumerates THREE. The smoke caller is real: `§148i-clibranch` sources hookrt.sh and calls `ghjig_state_dir_cli` (head smoke L1163-1165).
action: fix-now
remedy-survives: yes
axis: narrow<->wide
target-position: interior
justification: Both endpoints measured. FLOOR = the measured caller set — 4 paths at head; any claim narrower than that is false by measurement. CEILING = the extent the pinned command actually searches (`scripts/ .claude/hooks/ .claude/commands/`); a claim wider than its own search is §1.10(b)'s narrower-than-its-search fixture, which is half of this defect. The reviewer's suggested narrowing lands inside that bound and reproduces true: `scripts/ghjig_file_review_post.sh` IS the only script caller outside the smoke suite. Ground is the measurement, not any prior failure's direction. Correction to the reviewer: the comment is at hookrt.sh **:102**, not :79 — the anchor, not the line, is authoritative.
swing: not-evaluated
```

Recurrence note for F2: this is a claim-class finding (a census of a mutable set). The rule is **not armed** — it requires a confirmed occurrence in a *resolved prior judged round*, and this PR's prior round is `none (first round)`. Instance correction therefore survives.

```
finding: F3 — SPEC.md:2358's `because` misattributes the cause of the 8-or-9 narrowing to the split-vs-join
site: SPEC.md#§7.1-TTL-operand-validation
axis-key: SPEC.md#§7.1-TTL-operand-validation::causal-attribution
verdict: confirmed
evidence: executed=bash <function-framed repro of escape.sh:126's joined `if`, matching the real function frame> mode=script-file matches-artifact-mode=yes result=`%s` -> `[: %s: integer expression expected` then `%s: syntax error: operand expected`, the FUNCTION FRAME UNWINDS (neither the honored nor the blocked branch prints), execution resumes at the next top-level statement, exit 0. `0888888888` -> `value too great for base`, IDENTICAL outcome. `0177777777` -> reject, rc=1. So `%s` traceless-allows at escape.sh exactly as the octal shape does, though it is neither leading-zero nor 8-or-9; the split-vs-join is what separates the two SITES, not what narrows the reach set HERE. The preceding sentence already states the real cause (`[` base-10 accept vs `$(( ))` octal reject).
action: fix-now
remedy-survives: no
axis: none — discrete (a causal attribution names one mechanism or another; the measurement selects between two candidate causes, and there is no interior position between them)
target-position: n/a
justification: n/a
swing: n/a
```

Why F3's remedy does not survive — two independent grounds, both measured:
1. **It REDs the suite as written.** The reviewer's replacement phrasing drops the locked literal `splits the two comparisons into two`. Applied literally: `RED: 148h-ttl3 … (missing: <LOW-1:splits-the-two-comparisons-into-two>)`. Any replacement must keep both locked literals in-region.
2. **It is internally contradictory.** The proposed text retains *"only leading-zero epochs containing an 8 or 9 fall open here"* and then adds *"which is why `%s` shares this row at `escape.sh`"*. If `%s` shares the row at `escape.sh` — measured, it does, identically — then "only … 8 or 9 … fall open here" is false in the same breath. The finding is real; the author owns writing the replacement.

```
finding: F4 — `(INFO-4-corrected)` / `(LOW-1-corrected)` are prior-round provenance tags in a durable artifact (SPEC §1.10(c))
site: SPEC.md#§7.1-TTL-operand-validation
axis-key: SPEC.md#§7.1-TTL-operand-validation::provenance-tag
verdict: confirmed
evidence: executed=`grep -nF -e '(INFO-4-corrected)' -e '(LOW-1-corrected)' <head SPEC copy>` then the extracted arm over a tag-stripped copy mode=script-file matches-artifact-mode=yes result=both tags present at SPEC.md:2354 and :2356; after `sed` removal the count is 0 and §148h-ttl3 is still GREEN — the lock pins the sentence bodies, not the parentheticals, so removal carries no lock cost.
action: fix-now
remedy-survives: yes
axis: none — discrete (a provenance parenthetical is present or absent; §1.10(c)'s remedy is deletion, which admits no intermediate position)
target-position: n/a
justification: n/a
swing: n/a
```

```
finding: F5 — the PR body departs from the §9.3/§9.3.1 body contract
site: SPEC.md#§9.3.1
axis-key: PR#693-body::section-contract
verdict: confirmed
evidence: executed=`gh pr view 693 --json body -q .body | grep -nE '^## '` and `git show <HEAD>:.claude/templates/pr_body.md | grep -nE '^## '` mode=static-corpus-query matches-artifact-mode=yes result=the body has 4 sections (What, Plan, The four corrections reproduced, Test evidence); the template declares 13. §9.3.1's table row reads `Ship gate | fixed (template); /ship verifies and ticks` — that surface is absent.
action: fix-now
remedy-survives: yes
axis: fewer<->more
target-position: interior
justification: Both endpoints from §9.3.1's table, not from any prior failure's direction. FLOOR = `## Ship gate` — §9.3.1 names it as the surface `/ship` verifies and ticks, and `/ship` is executing right now, so its absence is an operational gap in this very run, not a cosmetic one. CEILING = the template's 13 sections; beyond that is invention. The judge is **upgrading the reviewer's "5 optional" to fix-now** on that floor alone — the reviewer rated it before the `/ship` context was in view.
swing: not-evaluated
```

```
finding: F6 — AC4 (full suite green in a window with no live agent worktree) is not verifiable from a worktree-isolated reviewer
site: <issue #656>#AC4
axis-key: issue#656::ac4-verifiability
verdict: confirmed
evidence: executed=`gh issue view 656` + reading the reviewer's own dispatch conditions mode=static-corpus-query matches-artifact-mode=yes result=AC4 requires a no-live-agent-worktree window; a worktree-isolated reviewer is itself such a violation, so the `pass=1335 fail=0` figure is not re-derivable from that position. The reviewer flags it as unverified, explicitly NOT as a defect, and explicitly declines to ask for a refresh.
action: drop
remedy-survives: n/a
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
```

Why F6 is `drop`, not `defer-to-issue`: it names no defect and requests no change. The reviewer's refusal to refresh the figure is **correct** and the judge affirms it — §1.10 says *remove or pin, never refresh*, and re-running that measurement to update the number is exactly the refresh the rule forbids. AC4's discharge remains a `/ship` gate obligation on the author in a clean window; that is a gate step, not a fix-list item.

### Closing observations (not judged records)

- **Cross-PR recurrence at F2's axis key, recorded as an observation only.** Issue #656's body documents that `hookrt.sh`'s caller claim was *already* corrected once ("ONLY /file-review-side caller" -> "only script caller", CR-minor-3), and that correction has landed inexact again. The recurrence rule arms on a resolved prior judged round **on the PR under review**, and this PR's is `none (first round)` — so F2's instance correction is ruled surviving rather than stretching the rule across PR boundaries to reach a stronger conclusion. The pattern is flagged because it is the exact loop #656 exists to end.
- **F1 region width:** the reviewer said 56 lines; measured **58**. Immaterial to the finding — noted for accuracy, since the number is the finding's own evidence.
- **F2 line number:** the reviewer cited `hookrt.sh:79`; the comment is at **:102**. The stable anchor `ghjig_state_dir_cli` is authoritative.
- **A coupling worth stating once:** F1's relocation and F3's rewrite touch the same lines and the same two locked literals. Whichever lands second must re-verify `§148h-ttl3`, because F1 narrows the region to 17 lines — after which any locked literal that drifts out of the taxonomy block REDs immediately instead of being silently absorbed by 40 lines of unrelated §7 prose. That is F1 doing its job, but it will surprise an author who lands F3 first and F1 second.

ilgyu-yi:
## Finding triage (round 2)

`reviewed-head: 407e71c6729102f1970634854593308fff53a873`

**dispatch** — findings from one fresh `code-reviewer` invocation, distinct from the round-1 reviewer; judgment from one separate `finding-judge` invocation, worktree-isolated, no shared context. Neither agent can self-observe this limb, so the caller records it. The §4.11 high-asymmetry tier was evaluated as NOT applicable: both guard-file diffs are comment-only and functionally byte-identical to base, no `docs/ADRs/` file, no force-push in the motion.

**prior-round read** — the judge's own trusted-author comment scan found exactly one canonical marker, `round=1 head=82fe3f5`, with a matching header line; round = 2, prior resolves to `1@​82fe3f5`. Every `swing:` below is genuinely evaluated, never a vacuous pass.

`code-reviewer`'s verdict token is **`ship after fix`** and is untouched — the judge converts no verdict and casts no vote of its own.

**Round-1 fix-now list: all five discharged**, verified by execution rather than by reading the diff (F1 region 17 lines with the in-region decoy now REDding, F2 caller claim true at head, F3 internally consistent under the new scope qualifier, F4 tags at count 0, F5 all 13 template sections with every pinned command reproducing byte-for-byte).

### Summary

| # | Verdict | Action | Reviewer's remedy |
|---|---|---|---|
| F1 | confirmed | **fix-now** | survives — replace the gloss with a pointer to §7.1 |
| F2 | confirmed | **fix-now** | survives — name the `now` operand in bullet 2 |
| F3 | confirmed | **fix-now** | survives — restore the dropped scope qualifier |
| F4 | confirmed | **fix-now** | survives — "one doc-prescribed agent invocation" |

Four confirmed, four to fix, none refuted. All four are prose-only; **none touches a locked literal** — proven by execution, not by inspection.

### The records

```
finding: F1 — the §148h-ttl3 comment restates §7.1's claim on its own authority, unscoped, and its causal connector now contradicts SPEC.md:2399
site: scripts/test/smoke.d/70-gates-contentlocks.sh#§148h-ttl3
axis-key: scripts/test/smoke.d/70-gates-contentlocks.sh#§148h-ttl3::causal-attribution
verdict: confirmed
evidence: executed=`bash f2.sh "%s" 1755000000 0` / `0888888888 …` / `0177777777 …` (escape.sh's TTL tail lifted verbatim from the head blob into a scratch script, driven from top level) mode=script-file matches-artifact-mode=yes result=`%s` -> traceless-allow (NEITHER branch prints, frame unwinds, `RESUMED-NEXT-TOP-LEVEL`, rc=0) though it is neither leading-zero nor 8-or-9; `0888888888` -> identical traceless-allow; `0177777777` -> BLOCKED. Both octal cases share one statement structure and differ in outcome, so the statement structure cannot be what narrows the set — the comment's `because` is false by measurement, and SPEC.md:2399 states the opposite in terms ("It is **not** caused by the statement structure"). Separately `bash f2.sh 1755000000 "" 0` -> HONORED, a third fall-open shape, so "only … 8 or 9 fall open in escape.sh" is false unscoped.
action: fix-now
remedy-survives: yes
axis: fewer<->more
target-position: low
justification: Both endpoints stated, each with the measurement it comes from, and independent of any prior failure's direction. CEILING = zero mechanism claims asserted on the comment's own authority — set there not by taste but by a measurement inside this PR: the gloss landed at `50fb84a` (Test) and was contradicted by `407e71c`'s SPEC three commits later, so a hand-maintained restatement drifted from its SSOT within a single PR. SPEC §9's reference-don't-restate is the independent definitional ground. FLOOR = the comment must still name which two literals are locked and point at §7.1 as their owner; deleting it outright would strip the anti-vacuity rationale that makes the arm auditable, so `low` is the floor, not zero comment.
prior-round: 2 (this round) — prior 1@​82fe3f5
swing: none
```

**Recurrence (check 6) on F1 — the caller flagged the question; answered three ways, all pointing the same direction.**
1. **Scope gate, the decisive one.** Check 6 governs *claim-class* findings — a live count, a census of a mutable set, a self-referential measurement, or review-round narrative. F1's subject is a **mechanism claim** about how a bash arithmetic error propagates. It does not expire, counts nothing, surveys nothing. Behaviour findings are out of scope by construction. **Not armed.**
2. **Key gate.** The kind was deliberately named `::causal-attribution` to *match* round-1 F3 rather than to differentiate it, so the non-match rests entirely on the site half — and that half is a real difference: round-1 F3 was `SPEC.md#§7.1-TTL-operand-validation::causal-attribution`, and that copy **was** corrected (measured: `SPEC.md:2399` now carries the negation). The surviving copy is in a different file. **Not armed.**
3. **And it is moot.** The reviewer's remedy is *already* the removal/restructure class an armed recurrence rule would compel. The honest answer is: no, and it would not matter if it were.

Constraint the author owns (a constraint, not a remedy): the replacement must assert **no** independently-verifiable mechanism, or it re-creates the defect at the same key next round.

```
finding: F2 — SPEC.md:2400's "empty / `0x<hex>` below `created`" never names its operand and is false under the `created` reading bullet 1 sets up
site: SPEC.md#§7.1-TTL-operand-validation
axis-key: SPEC.md#§7.1-TTL-operand-validation::operand-ambiguity
verdict: confirmed
evidence: executed=`bash f2.sh "" 1755000000 0` / `1755000000 "" 0` / `0xff 1755000000 0` / `1755000000 0xff 0` / `1755000000 0x1755000000 0` mode=script-file matches-artifact-mode=yes result=the two readings have OPPOSITE truth values. created=EMPTY -> BLOCKED(parse-empty-value); created=0xff -> BLOCKED(stale-or-future-dated); now=EMPTY -> HONORED(escape/skip); now=0xff -> HONORED. Hex ABOVE created -> BLOCKED, and `$(( 0x1755000000 - 1755000000 ))` = 98455311168, reproducing the bullet's own figure exactly. The bullet is true of `now` and false of `created`.
action: fix-now
remedy-survives: yes
axis: implicit<->explicit
target-position: interior
justification: Both endpoints, from the measurement and from §7.1's own deference, not from any prior failure's direction. FLOOR = the operand named once. Below it the sentence is not merely terse: its two readings are BLOCKED and HONORED — measured — so leaving it implicit is a coin-flip on a false normative statement, not a cheaper equivalent. CEILING = explicitness that restates the operand-range rule itself, which §7.1's opening paragraph defers to the **Binding weight** bullet as SSOT; past the floor that is a §9 second copy. The reviewer's one-clause replacement lands between them.
prior-round: 1@​82fe3f5
swing: none
```

Two things the author needs on F2.

**It is lock-safe, proven by execution.** The judge applied the reviewer's replacement clause verbatim to a scratch SPEC copy and re-ran §148h-ttl3 lifted verbatim from the head smoke file: **GREEN**, region still **17 lines** (round-1 F1's floor holds). Control over unmodified head SPEC: also GREEN at 17.

**A scope boundary, because #683 is adjacent.** Issue #656 item 2 explicitly defers the *cross-operand* scope claim to **#683**. F2 is not that claim — it is a one-clause prose accuracy fix in text this PR is authoring into SPEC. The remedy is to name the operand in bullet 2; it is **not** to add a cross-operand analysis. On the in-scope question: yes, the wording is inherited near-verbatim from `origin/main`'s helper comment (confirmed, `escape.sh:132` at base carries the identical elliptical bullet), but this PR is the change that promotes it into SPEC as normative doctrine, so it belongs here rather than in its own filing.

**One correction to the reviewer, which strengthens the finding.** The reviewer's stated mechanism for an empty `created` — "arithmetic treats the unset name as 0 -> delta = `now` -> stale" — is not where it actually blocks. Measured, it blocks one clause **earlier**, at `parse-empty-value` (`escape.sh:94`), before the TTL comparison is reached. The reviewer's mechanism is the *secondary* one that would fire if that clause were absent. Verdict unchanged, and `created` turns out to be blocked by two independent clauses, not one.

```
finding: F3 — the changelog fragment describes the pre-F2 state of the hookrt.sh comment
site: changelog_unreleased/changed/693.md#four-corrections-gloss
axis-key: changelog_unreleased/changed/693.md#four-corrections-gloss::ssot-restatement
verdict: confirmed
evidence: executed=`git show 407e71c:changelog_unreleased/changed/693.md` + `grep -n 'only script caller' <head hookrt.sh>` + `git log -S … origin/main..407e71c` mode=static-corpus-query matches-artifact-mode=yes result=the fragment ends "…now claims only-script-caller"; at head the comment (hookrt.sh:102, anchor `ghjig_state_dir_cli`) reads "only script caller **on the RUNTIME path**", with the smoke test caller named on the next line. Bare "only script caller" is precisely what round-1 F2 measured false. Fragment authored at `82fe3f5`, not refreshed by `407e71c` — confirmed by log.
action: fix-now
remedy-survives: yes
axis: none — discrete (a restatement of another artifact's text is accurate at head or it is not; the remedy selects among three discrete forms: add the qualifier, describe the correction instead of quoting the claim, or drop the clause)
target-position: n/a
justification: n/a
prior-round: 1@​82fe3f5
swing: n/a
```

Recurrence on F3: **not armed** — not claim-class, and its site differs from round-1 F2's (`.claude/hooks/hookrt.sh#ghjig_state_dir_cli::census-claim`). Author latitude worth naming: the more drift-resistant of the three forms is to **describe the correction** rather than to quote the corrected claim — quoting it is what put the fragment one reword behind in the first place.

```
finding: F4 — the PR body's four-item gloss classifies file-review.md as "a doc mention"
site: PR#693-body#F2-enumeration
axis-key: PR#693-body::caller-enumeration-role
verdict: confirmed
evidence: executed=`gh pr view 693 --json body -q .body | grep -n 'doc mention'` + `sed -n '79p' <head file-review.md>` + `gh issue view 656 -q .body | grep -n 'file-review.md'` mode=static-corpus-query matches-artifact-mode=yes result=body line 79 assigns a role to each of the four paths (definition site / doc mention / runtime script caller / test caller); file-review.md:79 is the literal `. ".claude/ghjig-root/.claude/hooks/hookrt.sh"; printf '%s\n' "$(ghjig_state_dir_cli)/file-review/staging"`, an instruction the agent executes. Issue #656 item 4 records exactly that. Within the body's own role-classification scheme, "doc mention" is the wrong role.
action: fix-now
remedy-survives: yes
axis: none — discrete (a role label in a four-item enumeration is one role or another; there is no continuum between "doc mention" and "doc-prescribed agent invocation")
target-position: n/a
justification: n/a
prior-round: 1@​82fe3f5
swing: n/a
```

The judge nearly ruled F4 `drop` — "doc mention" is not literally false, and the reviewer rated it INFO. What decided `fix-now`: this enumeration is the PR body's **own evidence for F2's discharge**, and F2's defect was an undercounted caller set. Repeating the undercount's classification error inside the correction's evidence is the same imprecision one level over, in a body being edited for the ship gate anyway. It does **not** falsify the `hookrt.sh` comment, which correctly qualifies with "script caller".

### Lock exposure — checked directly, not assumed

For **F2** re-running §148h-ttl3 is the right check, and it was run green. For **F1** the arm cannot see the edit at all: §148h-ttl3 reads `$SHELL_ROOT/SPEC.md` only. The two arms that *do* read `70-gates-contentlocks.sh` itself — §156O (`S156O_SRC`, line 2747) and §156V (`S156V_JSRC`, line 3122) — are bounded to §156a/§156j constructs, none of which live at lines 1005-1006. A repo grep for the comment's distinctive strings finds them only in the comment itself. Scope of that claim: three arms measured, not all ~1335 assertions.

### Closing observations — not judged records, no action requested

- **A pattern across three of the four findings.** F1, F3 and round-1's F2 are the same failure: a **scope qualifier dropped from a restatement**. §7.1 says "within the leading-zero shapes"; the smoke comment drops it. `hookrt.sh` says "on the runtime path"; the changelog drops it. Each restatement lost the exact clause that made the original true. That is the drift #656 exists to end, and it is arriving through restatement sites rather than through the original claims.
- **Bullet 1's operand framing is also `created`-centric while the phenomenon is symmetric** — measured, `%s` traceless-allows at *both* operands. Deliberately **not** filed: it is adjacent to the cross-operand scope claim #656 item 2 routes to **#683**, and adding findings is not the judge's to do. Flagged only so the author does not read F2's remedy as licence to open that seam here.
- The reviewer's line-number citations all resolve at head (smoke 1005-1006, SPEC 2399/2400, file-review.md 79); the round-1 correction that `hookrt.sh`'s comment is at **:102** rather than :79 still holds.

ilgyu-yi:
## Finding triage (round 3)

`reviewed-head: 7ffc58204101d8f5c7902146b594affbae6bf02e`

**dispatch** — findings from one fresh `code-reviewer` invocation, distinct from the round-1 and round-2 reviewers; judgment from one separate `finding-judge` invocation, worktree-isolated, no shared context. Neither can self-observe this limb, so the caller records it.

**prior-round read** — the judge's own trusted-author scan found exactly two canonical markers (`round=1 head=82fe3f5`, `round=2 head=407e71c`), no duplicates; round = 3, prior resolves to `2@​407e71c`. Every `swing:` is genuinely evaluated.

`code-reviewer`'s verdict token is **`ship after fix`** and is untouched. The judge casts no vote.

### Summary

| # | Verdict | Action | Reviewer's remedy |
|---|---|---|---|
| F1 | confirmed | **fix-now** | survives — **preferred (subtraction) form only**; the qualify alternative does not |
| F2 | confirmed | **fix-now** | survives — direction left to the author |
| F3 | confirmed | **fix-now** | survives — implied subtraction, floor-constrained |

Three confirmed, three to fix, none refuted. Round-2's four fix-now items were re-verified discharged.

### The records

```
finding: F1 — the round-2 fix commit authored a new false claim into SPEC §7.1 bullet 2: a `0x<hex>` `created` does not universally BLOCK, so "the two readings have opposite outcomes" is false
site: SPEC.md#§7.1-TTL-operand-validation
axis-key: SPEC.md#§7.1-TTL-operand-validation::operand-ambiguity
verdict: confirmed
evidence: executed=`bash ./ttl.sh <created> <now>` — a structural copy of escape.sh:126-128 at head 7ffc582, PRE-GUARD, driven from top level mode=script-file matches-artifact-mode=yes result=created=0xff -> BLOCKED; created=0x689b2cb6 (= 1754999990, measured, a hex-rendered epoch inside the window) -> HONORED; now=0x689b2cb6/created=1755000000 -> HONORED; created=EMPTY -> BLOCKED; now=EMPTY -> HONORED; now=0x1755000000 -> BLOCKED (delta +98455311168, reproducing the bullet's own figure). `[ 0x689b2cb6 -gt 5 ]` -> rc=2 on either side. All three reviewer rows reproduce exactly.
action: fix-now
remedy-survives: yes
axis: implicit<->explicit
target-position: interior
justification: Both endpoints, each from a measurement taken this round, independent of any prior failure's direction. FLOOR = the operand named once — measured, now=EMPTY HONORS while created=EMPTY BLOCKS, so for the empty class the two readings genuinely differ; deleting the naming clause re-opens round-2 F2's measured defect, so `low` is excluded on evidence, not caution. CEILING = no claim about hex-`created` behaviour in this bullet — measured, created=0x689b2cb6 and now=0x689b2cb6 BOTH honor, so no hex-operand statement is true unqualified, and the only true form carries the magnitude qualifier, which is the TTL rule §7.1's opening paragraph defers to the Binding weight bullet as SSOT. Past the floor it is a §9 second copy. Interior = the floor exactly, nothing more.
prior-round: 2@​407e71c
swing: none
```

On F1, three rulings the author owns:
1. **Only the preferred remedy survives.** Keep *"The operand here is the **clock reading** `now`, not `created`."* and stop. The reviewer's alternative ("…a `0x<hex>` `created` far enough below `now` BLOCKS as stale") is *true* as measured — but it lands past the ceiling, and it is a fourth mechanism claim at a site that has produced a false one in three consecutive rounds. The reviewer's own closing line forbids it; the two offered remedies are in tension, resolved against the alternative.
2. **Lock-safe, proven by execution.** The subtraction applied verbatim to a scratch SPEC copy: §148h-ttl3 **GREEN**, region unchanged, `wc -l` identical both sides — **no `build_toc.sh` regeneration needed**.
3. **The subtraction leaves no orphaned falsity.** The bullet's surviving sentences were re-measured: "Hex *above* `created` blocks" reproduces exactly; now=EMPTY -> HONORED holds.

Recurrence (check 6) on F1: **not armed** — F1 is a mechanism claim, not claim-class (no live count, census, self-referential measurement, or round narrative); the same ruling round 2 made, not stretched because the pattern is irritating. And moot: the surviving remedy is already the removal class an armed rule would compel. What terminates the loop is the **ceiling** — no hex-`created` claim in this bullet, at all; a fourth-round reworded mechanism claim at this key is what that ceiling was written against.

```
finding: F2 — the PR body pins one full-suite measurement to two different commits: `## Test plan` says `7ffc582`, `## Ship gate` says `407e71c`
site: PR#693-body#Test-plan-and-Ship-gate
axis-key: PR#693-body::suite-run-pin
verdict: confirmed
evidence: executed=`gh pr view 693 --json body -q .body | grep -nE '7ffc582|407e71c|pass=1332'` + `git show -s --format=%cI` on both commits + `gh issue view 656 --json comments` mode=static-corpus-query matches-artifact-mode=yes result=body:135 "Full suite (AC4). Run at `7ffc582`" over `smoke: pass=1332 fail=1`; body:161 "full suite at `407e71c`…" — the Ship gate cites the Test-plan block AS its source ("above") while pinning it to a different commit. Timeline: 407e71c 04:20:20Z < issue #656 AC4 comment 04:33:23Z < 7ffc582 05:12:41Z; that comment reports the identical figures and narrative attributed to `407e71c`.
action: fix-now
remedy-survives: yes
axis: none — discrete (a commit pin names one SHA; two surfaces citing one measurement either agree on it or they do not)
target-position: n/a
justification: n/a
prior-round: 2@​407e71c
swing: n/a
```

One correction to the reviewer, sharpening F2: "At most one is true" is an overreach — if the suite genuinely ran at both commits, both sentences could be true. The confirmed defect is narrower and stronger: the body presents **one** block of evidence under **two** commit pins, and body:161 explicitly points at body:135 as its source. A single cited measurement cannot have occurred at two commits.

What neither reviewer nor judge can resolve, stated plainly: the timestamp chain is *consistent with* a relabel but not dispositive — a genuine second run over a prose-only delta would return identical numbers. **Only the author knows whether a second run happened.** Constraints on the remedy: both surfaces state the run that actually happened, at the commit it actually happened on; and this must **not** be discharged by re-running solely to make the `7ffc582` label come true (§1.10 remove-or-pin-never-refresh, per round-1 F6). A genuine new run in a clean window, reported truthfully at its own commit, is a new measurement and is fine.

Recurrence (check 6) on F2: **in scope but not armed** — F2 is claim-class (a pinned self-referential measurement), but the key `PR#693-body::suite-run-pin` was confirmed in neither prior round (round 1: `::section-contract`, `::ac4-verifiability`; round 2: `::caller-enumeration-role`). Instance correction survives; the class is already correct (a pin is not an expiring claim; the defect is mis-attribution, and correcting an attribution is not a refresh).

```
finding: F3 — the §148h-ttl3 comment block states a discipline ("never why they hold") that its INFO-4 line does not follow
site: scripts/test/smoke.d/70-gates-contentlocks.sh#§148h-ttl3
axis-key: scripts/test/smoke.d/70-gates-contentlocks.sh#§148h-ttl3::ssot-restatement
verdict: confirmed
evidence: executed=`git show 7ffc582:scripts/test/smoke.d/70-gates-contentlocks.sh | sed -n '994,1012p'` + `git log -S` on both clauses over origin/main..7ffc582 mode=static-corpus-query matches-artifact-mode=yes result=the discipline sentence's subject is "**this comment**", so its scope is the block, not the LOW-1 bullet alone; the INFO-4 line carries "the meta-claim #635 got wrong twice", restating §7.1's "cleared this class twice (#635)" — a live second copy at head. Authorship: the discipline sentence was authored by `7ffc582`, the INFO-4 clause by `50fb84a` — the inconsistency was CREATED by the round-2 repair.
action: fix-now
remedy-survives: yes
axis: fewer<->more
target-position: low
justification: Both endpoints — the same continuum round-2 F1 measured at this same site; this is that remedy applied to the half it missed, not a new position. CEILING = zero claim asserted on the comment's own authority (SPEC §9; round 2 measured the concrete failure). FLOOR = the block must still NAME the two INFO-4 literals (`script-file mode`, `set -uo pipefail`) and point at §7.1 as owner — a reader auditing the arm's RED needs the comment to say which. So `low` is the floor, not deletion of the INFO-4 sub-bullet.
prior-round: 2@​407e71c
swing: none
```

Why `fix-now` on an INFO, without touching the verdict: the reviewer's INFO rating and "not a merge condition" are severity statements and stand — F3 is not upgraded into a merge condition. `fix-now` is simply the right triage cell: `drop` is for a finding naming no defect (round-1 F6); `defer-to-issue` would mint durable memory for a two-clause deletion in a file this PR is already editing. Grounds: a self-inconsistency newly created by the round-2 fix commit, subtraction, measurably lock-free (§148h-ttl3's only file operand is `$SHELL_ROOT/SPEC.md` at both the anchor count and the region cut — editing this comment cannot red it; scope of that claim: this one arm).

Adjacent surface, deliberately not filed: the clause "locked alongside LOW-1 by plan-reviewer mandate" is review-round process narrative in a durable artifact (the class round-1 F4 removed from SPEC). The reviewer filed F3 on half-conformance, not §1.10(c), and adding findings is not the judge's to do — flagged so the author can fold it into F3's subtraction.

### Anti-swing check

All three prior-round comparisons resolved against `2@​407e71c`; no vacuous values.

- **F1** — same axis and site as round-2 F2, judged target `interior` with floor "the operand named once". The round-3 remedy lands **on that floor** — back inside the same interior band. What failed in round 2 was the delivered text overshooting the stated ceiling, not the judged position; a retreat to the ceiling is not the opposite end. `swing: none`. The directional reading ("last round added words and they were false, so remove words") was specifically rejected as rule-1 reasoning.
- **F2** — discrete, `swing: n/a`.
- **F3** — same axis, site, and the same `low` position as round-2 F1: a continuation onto the half it missed. `swing: none`.

### Closing observations — not judged records

- Region line count: round 1 pinned 17, the judge's own count of the awk-cut region is 16 content lines — a trailing-newline counting artifact, immaterial (F1's remedy is a within-line subtraction), recorded because the figure was made load-bearing.
- The reviewer's evidence held up under re-measurement without exception, including the exact decimal of `0x689b2cb6` and `+98455311168`. The only corrections: "at most one is true" (replaced with a stronger, checkable ground) and the region count.

ilgyu-yi (commented):
**Verdict: approve** — `code-reviewer` round-4 verdict `ship` at this head, zero findings at blocking or fix-required severity.

Summary: prose-only relocation of the TTL fall-open taxonomy into SPEC §7.1 with four measured claim-corrections; the reviewer re-verified by execution that all three round-3 judged items are discharged (the §7.1 bullet-2 subtraction lands exactly at the judged floor with no orphaned falsity, the `§148h-ttl3` comment conforms to its own reference-don't-restate discipline, the suite-pin surfaces agree on one run at one commit), that the comment-stripped guard files are byte-identical to `origin/main`, that the `§148h-ttl3` lock is green with the in-region decoy mutation redding, and that every pinned command in the PR body reproduces at this head. Review round history (4 rounds, three `finding-judge` triage comments) is durable on the PR.

Issue #707, opened 2026-08-24
the activation arm has no triage layer behind its findings, unlike the code arm


ilgyu-yi:
## What

Issue #645 diagnosed that the shell *"has a judge for plans and none for reviews"* and shipped `finding-judge` as the remedy. That remedy reached one reviewer arm.

`.claude/agents/finding-judge.md` scopes itself to the code-review arm — its description reads ``Post-`code-reviewer` finding judge. Called by `/review` (step 3.5) and `/ship` (step 1.5)``. `.claude/commands/activate.md` has no equivalent step: step 3 parses the `activation-reviewer` verdict and hands it straight to step 4's filer matrix. Nothing rules whether a finding is real before the author acts on it, nothing grades severity, and nothing is independent of the reviewer that produced it. The one categorization on that path is step 3's reject-audit reason-class token, which is observability by its own statement and grades nothing.

Two consequences, and the second is the expensive one.

**The disposition for a finding that is real but not body-blocking is improvised, not absent.** `activation-reviewer`'s grammar is `pass` / `revise` / `reject` (`.claude/agents/activation-reviewer.md:150-152`) and names no place for such a finding. One was nonetheless carried forward: #705's `pass` comment opens a paragraph **"Carried to the PR body, not fixed here."**, hands the finding to the implementation cycle, and states why it sat below the revise bar. So the arm's forward token *can* discharge a below-bar finding without returning the body — the gap is not that the outcome is unreachable but that reaching it is **uncontracted**. No step produces it, nothing grades what falls above or below the bar, no artifact records the ruling except the reviewer's own prose, and nothing is independent of the reviewer that raised the finding.

The same is true one arm over, which is the point rather than a counterexample. `plan-reviewer` on #705's plan contest returned, per PR #706's `## Alternatives considered`, *"Second pass on the composition: `ship`, with three residual defects converted into binding implementation constraints rather than another authoring round."* — and `plan-reviewer.md:57` documents only `VERDICT: ship — <one-line naming the winning candidate…>`, with no constraints-on-ship affordance, so that too extended its grammar. **On both arms the forward disposition is improvised; on the code arm alone a layer exists that makes it durable** — `finding-judge` produces a judged, ranked list, posted before the author acts, with an anti-swing check against the prior round. Whether the remedy for the activation arm is that same layer or something else is this issue's question. It is **not** a proposal to change any reviewer's verdict grammar, which Directive #638 rules out by name (see Boundaries).

[...]

PR #706, opened 2026-08-24
docs(#705): deliver SPEC 1.10 where a durable body is written

ilgyu-yi:
Closes #705

## How this serves the mission

`MISSION.md`'s *Success looks like* names the channel: **"Each contributes friction back into the SPEC"**, under its multiple-repos item.

The friction repaid is the review round spent re-deriving a body's claims after it was filed — the one-sided pair `MISSION.md`'s enforcement dual names: the negative face (the reviewer gate) exists at every filing surface, the positive face (the authoring discipline that would let a body pass first time) was delivered at two of them.

Secondarily, `MISSION.md`'s *The mechanism* judges a proposal against that surface — **"a proposal that bloats context past the task is regression,"**

The always-on surface **shrinks** rather than grows: §1.10 binds only at the instant a durable body is authored, so the binding belongs at the writing surfaces and `.claude/CLAUDE.md`'s statement contracts to a pointer.

## Plan

Three moves, no new mechanism.

1. **Deliver at the writing surfaces.** One byte-identical clause constant at seven carriers — four command surfaces that write an Issue, Directive, or PR body, and the three skeletons a shell step composes a body from (the latter as a head comment). `planner` carries its own line, because it stages no file and performs no write. **Three distinct maintained strings**, not eight.
2. **Shrink the always-on surface.** `.claude/CLAUDE.md`'s Evidence-discipline bullet becomes a scope-correct, `(a)`–`(d)`-labelled pointer, paid for by a subtraction in the same file.
3. **Lock it.** Smoke §182, shaped after §179 (#698).

The clause anchors on **"before the `gh …` call that writes the body"**, never on a step number: each carrier has exactly one body-flagged write site, and step numbers in these files churn (`4.5`, `4.6`, `5.5`, `5.6` are all later insertions) while the write verb does not.

It carries the **fail-open surfacing obligation** in `.claude/commands/file-issue.md:41`'s form. `scripts/lint_citations.sh` prints its declines to stderr and `exit 0`s unconditionally, so without that clause a report relayed at a new carrier makes "never ran" indistinguishable from "clean".

Three insertion points sit where the body actually exists: `/work-on` **inside step 8 after `git push`** (so the staged bytes are the bytes filed and the Doc commit is on the branch), `/revise-directive` **step 2 now states the write** it previously left implicit, and `/consume-initiative` **creates its proposal files at extraction time** rather than at filing time.

[...]

PR #712, opened 2026-08-25
docs(#711): state the comment shape where the caller composes it

### LINKED ISSUE #711: the judged-list comment shape finding-judge tests for is stated in no caller, so which composition passes is left to the caller
## What

`.claude/agents/finding-judge.md` tests a judged-list comment for **two** things — the literal marker and a header line — and the instruction to write the second one exists in no caller. At merged `main` (`45d497b`):

```
$ command grep -c 'Finding triage' .claude/commands/review.md .claude/commands/ship.md \
    .claude/agents/finding-judge.md SPEC.md
.claude/commands/review.md:0
.claude/commands/ship.md:0
.claude/agents/finding-judge.md:1
SPEC.md:0
```

The consumer's rule, at `.claude/agents/finding-judge.md:26`:

> A canonical judged-list comment carries the literal HTML marker `<!-- finding-judge: round=<N> head=<sha> -->` and a first line matching `^## Finding triage \(round [0-9]+\)$`. A loose lookalike — a comment that merely mentions the header text, or carries the marker without the header — **does not count**.

Both callers instruct only the marker half — `.claude/commands/review.md:17` and `.claude/commands/ship.md:13` say *"carrying the canonical marker"* — and both deliberately decline to restate the schema, deferring to the agent file as contract SSOT under §9. So the party that writes the comment is never told the shape the party that reads it tests for.

**Measured on PR #706, which ran three judged rounds.** The first line of each judged-list comment:

```
$ gh pr view 706 --json comments \
    --jq '.comments[]|select(.body|test("finding-judge: round"))|(.body|split("\n")|map(select(length>0))|.[0])'
<!-- finding-judge: round=1 head=e36d90680f7036c8a036fa01a4ab26fc653da795 -->
## Finding triage (round 2)
## Finding triage (round 3)
```

Round 1 carries the marker with a `## Judged finding list — round 1` header, so it satisfies the marker half and fails the shape half. Rounds 2 and 3 are canonical, and each recorded the consequence from the consumer side, differently:

- **Round 2** had round 1 as its prior round and resolved it **against its own rule** — as `1@e36d906` rather than `unresolved`, on the ground that reporting `not-evaluated` over a demonstrably real, sha-verified prior round would be the vacuous pass its rules forbid. It named the producer-side gap as real and separable.
- **Round 3** had round 2 as its prior round and resolved it **by** the test, since that comment satisfies both halves. What it had to override instead was the recipe's round-number derivation: a literal `1 + canonical-count` yields **2**, colliding with the existing canonical round 2, because round 1 does not count.

So the check degrades in two different ways depending on which prior round it lands on, and both stem from one comment that satisfies half the test.

**What the caller is left to decide, and why nothing steers it.** The judge's own reply opens with `reviewed-head: <sha>` (`.claude/agents/finding-judge.md:114`), and the canonical marker appears nowhere in that output schema — tree-wide the literal occurs only in the consumer rule at `:26`. So the comment is necessarily **composed** by the caller rather than pasted, and the caller alone decides where the header, the `reviewed-head:` line and the marker each land, with no instruction anywhere naming the composition the consumer tests for. On PR #706 two compositions resulted: round 1 opened with the marker under a non-canonical header, while rounds 2 and 3 opened with the header, carried `reviewed-head:` on the following content line, and placed the marker last — satisfying both halves of `:26`. The gap is that the rule deciding between those compositions is stated in one file the producer never reads.

The failure mode is not a wrong verdict. It is that the **anti-swing check silently degrades to judgment exactly when a prior round exists to check against** — which is the only case the check exists for. A shape test the live producer cannot be expected to satisfy is a test that keeps producing that degradation.

## Why

`MISSION.md`'s enforcement dual names this shape directly: *"negative-only enforcement walls the agent in with no door (blocked, never guided to the alternative)"*. Here the consumer holds the negative half — a shape test that rejects — while no surface supplies the positive half of telling the producer what to write. The same file's isolation model is why a naive fix is wrong: SPEC and the command prose are *"a single source that every bound project resolves **live**"*, and §9 is precisely what made both callers decline to copy the schema. So the repair cannot be a third hand-synced copy of the header string.

Why now: it recurred across three rounds inside one PR and was flagged twice from the consumer side. And #707 would add a **third** producer of judged-list comments while forbidding itself any agent-file edit, so it inherits this gap as its only two implementation routes; its own activation review named that inheritance as the reason its body must reference this issue.

No existing surface absorbs it. #707 excludes `finding-judge`'s existing contract by name; #708, #709 and #710 are unrelated axes.

## Acceptance criteria

- [ ] The set of surfaces that **produce** a judged-list comment and the set that **consume** one are both derived, not listed, and each is stated as carrying the shape or not.
- [ ] Exactly one closure direction is ruled, and the one not taken is stated as rejected with its cost: **(a)** the producer is told the shape without a second hand-synced copy of it — single-sourced where both the caller instruction and the consumer test resolve it — so the composition the consumer accepts is stated where the producer reads it rather than left to the caller's discretion; or **(b)** the consumer test drops the header half and keys on the marker alone, with the measurement showing whether the marker alone still excludes the loose lookalikes the two-part test was written to exclude.
- [ ] The chosen direction is locked so a silent reversal reds — a producer instruction that stops matching the consumer test, or a consumer test that regains a half no producer is told, fails an arm rather than degrading to judgment.
- [ ] The fix does not add a hand-synced copy of the header string. If direction (a) needs the string in two files, the change states why that is not the §9 hazard it looks like.
- [ ] The change names its own negative face, or states that it has none and why that is acceptable (§6.0).

## Out of scope

- `finding-judge`'s judging contract — its per-finding schema, the axis menu, the three anti-swing rules, the evidence grammar. Only the comment's identifying shape and who is told it are at issue.
- #707's triage layer for the activation arm. That issue references this one; neither implements the other.
- Retro-fitting PR #706's round-1 comment. Its content is correct and its marker is canonical; rewriting a posted judged list to satisfy a test is the wrong direction.
- The `durable: none (<mode>)` path for a review with no PR, which is a separate arm of the same contract and is not implicated by the shape.

## Notes

- Anchors: `.claude/agents/finding-judge.md:26` (the two-part test) and `:114` (the judge's own reply shape, which the caller composes the comment from), `.claude/commands/review.md:17` and `.claude/commands/ship.md:13` (the marker-only instruction), SPEC §4.13's `Durable before the fix` paragraph and its `Contract SSOT` line (which is why the callers do not restate the schema).
- Evidence lives in PR #706's three judged-list comments; the two canonical ones state the gap from the consumer side in their own `Notes for the caller` sections.
- Source: recorded by two `finding-judge` rounds on PR #706, and named as a blocking omission by #707's round-1 `activation-reviewer` pass.

ilgyu-yi:
Closes #711

## Goal

Give the judged-list comment's shape, round derivation, and posting a **single code home** — `scripts/ghjig_judged_list.sh` — invoked by both producers and by the judge's read recipe, so `finding-judge`'s anti-swing check stops degrading whenever a prior round exists to check against.

`MISSION.md`'s enforcement dual: *"negative-only enforcement walls the agent in with no door (blocked, never guided to the alternative)"*. The comment shape had a rejecting half written and a guiding half written nowhere. This supplies the guiding half in its strongest form — the producer is not told the shape, it is handed the finished composition by shared code. It is also the first application of Directive #637's L3+L4 (contracts move to code; seam contracts live in shared code both sides call), named there as the delivered-elsewhere exemplar.

## Plan

Direction **(a)** of Issue #711's AC 2, in its strongest form: the shape is stated once, in code, and neither caller nor judge composes anything.

1. **One script, four modes.** `post <pr> <judge-output-file>` validates the judge output (a smuggled concrete marker, a non-`reviewed-head:` first line, and the no-PR `n/a` input are named rejects), derives the round as `1 + max(round=)` over canonical trusted-author comments, composes header + body + marker itself, self-validates, neutralizes mentions, and posts via temp-file `--body-file`. `rounds <pr>` / `show <pr> <round>` are the consumer side: canonicity is position-bound (header first line **and** marker last content line), duplicates surface with a distinct exit instead of being silently picked. `validate` exposes the composer's validator to the tests.
2. **Callers invoke, never compose.** `review.md` step 3.5 and `ship.md` step 1.5 reduce to the invocation; the one-comment-per-round and `durable: none (<mode>)` clauses stay.
3. **The judge consumes code output.** The read recipe's prose literals, the hand-counting instruction, the `fixture:canonical-comment` slice, and the judge-rendered envelope are deleted — the envelope existed only to remove the caller's substitution freedom, and the script removes the entire composition step.

**Ruled and rejected.** Direction (b) — dropping the header half and keying on the marker alone — is rejected with its measured cost: marker-alone still admits the marker-without-the-header class, which contains PR #706's round-1 comment and a trusted-author comment quoting a concrete marker (the class `.claude/commands/file-review.md` invented placeholder-form quoting for). The measurement is Issue #711's first comment; no header string is restated here.

**Route history.** The first three commits built this as a caller-copy instruction locked by phrase-derived smoke arms. Round 1's review measured that lock's ceiling (a mutant dropping the copy instruction stayed green), and the maintainer re-targeted the closure to the script route — the ruling and its consequences for the round-1 fix list are recorded in the [direction-decision comment](https://github.com/ilgyu-yi/ghjig-claude/pull/712#issuecomment-5406003821). The rework rides as three further commits; history is unrewritten.

## Alternatives considered

The plan contest record (SPEC §4.8), retained as the record of the original route: Plan A (`planner`) contributed the single-file frame; B1 (`correctness`) caught the unchanged derivation mis-deriving and contributed `1 + max(round=)` and the placeholder form; B2 (`robustness/risk`) measured that a presence arm was already green while the failure happened and contributed the rendered envelope; `plan-reviewer` ruled the B1 ⊕ B2 composition on A's frame (`refine`, `refine`, `ship`). The script route supersedes the composition's transport (fixture + envelope + copy instruction) while keeping what the contest established: the `max` derivation, placeholder-form marker mentions, and the suite-level negative face. A replan (`planner`, 2026-08-25) produced the rework plan under the maintainer's ruling.

## Key context

- `scripts/ghjig_judged_list.sh` — the contract home (modes, exit map `0 ok · 1 reject/failure · 2 show miss · 3 duplicate`, trusted-author filter byte-identical to `ac_closeout_gate.sh`'s).
- `scripts/test/smoke.d/71-judged-list.sh` — §183 behavior arms 183a–183o, PATH-stub `gh`, script exercised in artifact mode (#633).
- `.claude/agents/finding-judge.md` — read recipe re-based on `rounds`/`show`; keeps the trust rationale, fail-open, duplicate semantics, and one placeholder-form marker mention (the two shape-degradation `prior-round:` reasons were deleted in the round-2 fix — judged list F1).
- `.claude/commands/review.md` / `.claude/commands/ship.md` — the invocation bullets.
- SPEC §4.13 — referent moved to the script; the honest residual (a caller bullet edited to drop the invocation is review-guarded prose drift) is stated there per Directive #637 L3.

## Checklist

- [x] **Doc** (`f5fa0e7`): SPEC §4.13 names the script as the single code home; negative face = the script's behavior tests; no shape literal enters SPEC; changelog fragment rewritten.
- [x] **Test** (`3e456cb`): smoke §183 rebuilt as fifteen behavior arms over the script (fixtures for derivation, canonicity, rejects, validator, filter parity, gh-failure, round trip); the legacy phrase-derived arms and every `S183_*` symbol deleted in the same commit; intended red confirmed (script absent, fail-closed).
- [x] **Code** (`ff2f426`): the script ships; the agent-file shape section, envelope, and recipe literals deleted in the same commit; callers rewritten to the invocation; SPEC §8 script census synced (mechanical count + TOC, forced by smoke §103c — the one plan deviation).
- [x] PR body rewritten to the route; #639 carrier-handoff comment updated.
- [x] **Round-2 fix** (`e5dc83b`): the check-5 recipe-observability claim deleted per the judged list's F1 (deletion-default, Directive #637 L5); the stray blank line dropped (F3); residuals filed as #713 (F2).

## Evidence (rendered at `ff2f426`)

Producer/consumer sets, derived not listed (AC 1):

```
$ git grep -ln "ghjig_judged_list.sh post" -- .claude/commands/
.claude/commands/review.md
.claude/commands/ship.md

$ git grep -ln "ghjig_judged_list.sh" -- .claude/agents/
.claude/agents/finding-judge.md
```

Single carrier of the header literal (AC 4):

```
$ git grep -lF "Finding triage"
scripts/ghjig_judged_list.sh
scripts/test/smoke.d/71-judged-list.sh
```

Suite and lint (run by the assembler at `ff2f426`):

```
$ bash scripts/test/smoke.sh | tail -1
smoke: pass=1373 fail=0
$ bash scripts/lint.sh | tail -1
lint OK: 85 files clean (shellcheck v0.11.0)
```

## Out of scope

The round-derivation residuals (the silently-filtered anomaly classes, the zero-padded round crash, the unpaginated comment window, and their neighbors) — inherited by the script and filed as #713. Retro-fitting PR #706's comments; `finding-judge`'s judging contract; #707's activation-arm triage; any hook or merge gate. `docs/SUBAGENTS.md` is a thin pointer — `Docs: n/a`.

## Risks

- `review.md` / `ship.md` are Directive #639 item 7's **Exclusive** carriers; no #639 Execution Issue is open; this lands first and the handoff comment there is updated (the retired single-line-bullet constraint died with the legacy arm).
- The script is a new executable posting surface ingesting reviewer-authored text — `security-reviewer` is dispatched at review time alongside `code-reviewer`.
- A caller bullet silently edited to drop the invocation is prose drift no suite arm observes — review-guarded, stated in SPEC §4.13 (Directive #637 L3's honest ceiling).

## Test plan

Smoke §183 (183a–183o): behavior tests over the script behind a PATH-prepended executable `gh` stub — derivation, position-bound canonicity, the reject family, the validator driven directly, code-vs-code filter parity, fail-closed `gh` failure, and the post→rounds round trip.

## Docs touched

`SPEC.md` §4.13 + §8 census, `.claude/agents/finding-judge.md`, `.claude/commands/review.md`, `.claude/commands/ship.md`, `changelog_unreleased/fixed/711.md`.

## Ship gate

`bash scripts/test/smoke.sh`, `bash scripts/lint.sh`, `bash scripts/build_toc.sh --check`, and the CI checks.


ilgyu-yi:
## Review record — round 1, and the open decision

Archived here rather than in the PR body: review archaeology and a phase snapshot are the expiring class SPEC §1.10(c) assigns to comments.

**`code-reviewer` at `2a76f3d`: `ship after fix`.** It ran the suite in an extracted head tree (`pass=1365 fail=0`, `build_toc --check` rc=0, `lint.sh` 83 files clean), built **twelve** of its own out-of-tree mutants rather than trusting the author's table, and confirmed §183a, §183b, §183d, §183f and §183g each fire alone with no shadowing. It could not green a drifted envelope — **§183b, the arm the design turns on, holds.**

**F1 — blocking. §183c does not hold.** `s183_producer_flags` tests the producer line for two substrings: the derived heading and the fixed word `verbatim`. Measured: rewrite `.claude/commands/review.md:17` to keep the shape pointer and **drop the entire copy instruction** — restoring, in one caller, the exact gap Issue #711 exists to close — while the word `verbatim` survives incidentally in the neighbouring `@mention` clause. Full suite on that mutant: `pass=1365 fail=0`, byte-identical to clean head. A second adversarial variant — worded to invite the caller to assemble the comment itself and to disclaim any need to copy anything verbatim — also greens all seven arms. Both variants are mutant text that exists only in the reviewer's scratch tree, never in a tracked file. The arm's own `ng` text claims it reds exactly this case.

**F2 — AC 3's second limb.** `s183_rule_field` lifts two fields by two fixed introducing anchors, so a *third* requirement added to the consumer test is invisible to every arm. Measured green with a `judged-by:` line added to step 2.

**F3 — AC 2's rejected-direction clause is stated in no artifact this PR authors.** Direction (b) and its cost live in Issue #711's own first comment, which is pre-existing thread material rather than a ruling this change makes.

**F4–F6, low.** A `grep -cF` counts lines rather than occurrences against its comment's claim; the placeholder rationale still describes the retired counting semantics; and `1 + max(round=)` still re-issues a number in use in one case — zero canonical comments plus a historical non-canonical one carrying `round=1`, which is this PR's predecessor's round-2 situation.

**The open decision, and why it is not simply F1's patch.** The proposed remedy is to require the two envelope field keys on the producer line as well. That is a third substring test with the same weakness one column over: a caller can name both keys and still say "assemble it yourself". Three locks in this change's lifecycle have now been bypassed the same way — a challenger measured the base plan's pointer-presence arm green while the failure passed, the contest judge measured the composition's round trip covering only the fixture and not the envelope, and now this. Each was true, and each guarded a surface beside the one that fails.

The structural alternative is to remove the degree of freedom instead of testing that it is used correctly: have the judge emit the **whole comment body** — header, judged list, marker — as one block the caller posts unchanged. There is then no assembly step to instruct and nothing for an arm to approximate; §183c reduces to one fact rather than three. It stays inside AC 2's direction (a), since handing over the finished artifact is the strongest form of telling the producer the shape, and it preserves §183d/§183e because the rendered marker exists only in runtime output, never in a tracked file.

`finding-judge` was dispatched on the six findings with exactly that question — whether the proposed remedy closes the hole or relocates it — and its ruling had not returned when this session was flushed. **Nothing has been authored in response to any finding**, so the durable-before-the-fix ordering is intact and the next session starts from a clean decision point.

**State at flush:** three commits (`43bf616` Doc, `015c8fa` Test, `2a76f3d` Code), CI 6/6 green, checklist 7/7, working tree clean, no fix authored.

ilgyu-yi:
## Finding triage (round 1)

reviewed-head: 2a76f3dad8858e054e8baf6959aaefe71b9dda1d

dispatch: findings produced by `code-reviewer` (dispatched by the caller in the pre-flush session, verdict `ship after fix` at `2a76f3d`); judgment produced by a separately dispatched `finding-judge` instance — recorded by the caller per the limb-(a) contract.

All reproductions ran out-of-tree: head blobs and the §183 helper slice (`5622,5816`) copied into a session scratch tree, cwd and every write inside it, sourcing the section's own path-taking functions (the seam its header at `:5610` declares). No live file was mutated.

---

```
finding: F1 — the producer arm tests two substrings, so a producer that stops matching the consumer test greens
site: scripts/test/smoke.d/70-gates-contentlocks.sh#s183_producer_flags
axis-key: scripts/test/smoke.d/70-gates-contentlocks.sh#s183_producer_flags::coverage
verdict: confirmed
evidence: executed=bash <scratch>/run.sh <scratch>/m1 ; …/m2 ; …/m3 mode=script-file matches-artifact-mode=yes result=clean scratch tree 183a/b/c/g all empty; m1 (shape pointer kept, whole copy instruction dropped, "verbatim" surviving in the @-mention clause — i.e. main's pre-change bullet, the Issue's own gap) 183c:[] green; m2 ("compose it yourself … no need to copy anything verbatim") 183c:[] green; m3 (both field keys + heading + "verbatim" present, instruction inverted to "compose the comment yourself") also carries all four tokens
action: fix-now
remedy-survives: no
axis: permissive↔strict
target-position: interior
justification: the reviewer's remedy (require the two derived envelope field keys on the producer line) reds m1 and m2 — measured: neither line contains `comment-header:`/`comment-marker:` (grep -c = 0) — but greens m3, whose line carries `comment-header:`, `comment-marker:`, the heading and `verbatim` while instructing composition. So it is a strictly better vocabulary test, not a contract test, and the arm's own `ng` text at :5859 would still overclaim. Bounded at both ends, each end measured: FLOOR — the arm must red m1, m2 and m3, which requires (i) the two envelope field keys derived from the `comment-envelope` slice, (ii) each key bound to its position (`comment-header:` ↔ first line, `comment-marker:` ↔ last content line, order-sensitive, both position phrases lifted from the agent file's own "The caller posts the value of …" sentence rather than written here), and (iii) the prohibition clause ("substitutes nothing" / "copy whole lines") lifted the same way in place of the bare adverb hardcoded at :5631. CEILING — no whole-sentence equality and no hand-synced literal: adding either header literal to a caller reds 183d (census carriers-x2) and 183e, and the bullet is already load-bearingly single-line (the PR body records that residual), so a stricter pin would red on contract-preserving rewording. Independent ground: three measured mutants, not the direction of any prior failure. What the arm cannot reach at any position, and should say so in its comment: the producer is prose executed by an agent, so no file-level arm observes runtime composition — the honest ceiling is a documentation-regression detector, which is exactly the presence-arm limit B2 found one level out.
swing: not-evaluated
```

```
finding: F2 — a third requirement added to read-recipe step 2 is invisible to every arm
site: .claude/agents/finding-judge.md#read-recipe-step-2
axis-key: scripts/test/smoke.d/70-gates-contentlocks.sh#s183_roundtrip_flags::coverage
verdict: confirmed
evidence: executed=bash <scratch>/run.sh <scratch>/m4 mode=script-file matches-artifact-mode=yes result=step 2 amended with "It must additionally carry a second content line matching `^judged-by: finding-judge$`." → 183a:[] 183b:[] 183c:[] 183g:[] (183d/e/f are inert: no header literal and no `gh -q` filter byte changes)
action: fix-now
remedy-survives: no
axis: narrow↔wide
target-position: high
justification: AC 3's second limb does bind this change — the pre-change state *was* "consumer holds a half the producer is not told", so a consumer test that regains a half is the direction-(a) reversal, and `s183_rule_field` (:5654) lifts exactly two anchors, making that reversal unobservable. But the reviewer's remedy is false on arrival: step 2 carries **four** inline-code spans today, not two (measured: the marker template, the header regex, `round=`, `head=`), so "assert exactly two" reds at head and a pinned count of 4 would be an undeviced hardcoded number that also reds on innocuous editorial spans. Implementable without hardcoding, and bounded at both ends: FLOOR (wide) — every inline-code span on step 2 must be accounted for, so m4's `^judged-by: finding-judge$` reds; CEILING — a span is accounted for if it is satisfied by BOTH producer surfaces **or** is a substring of a span that is, which is what keeps head green: measured, `round=` and `head=` each occur once inside `<!-- finding-judge: round=<N> head=<sha> -->`, while `^judged-by: finding-judge$` is a substring of neither lifted literal and matches no canonical-comment fixture content line (fixture lines measured: header, `reviewed-head: <sha>`, elision, marker). Independent ground: measurement at head plus the m4 witness, not a prior failure's direction.
swing: not-evaluated
```

```
finding: F3 — AC 2's rejected-direction clause is stated in no artifact this PR authors
site: PR #712 body#Alternatives-considered
axis-key: pr-712-body#AC2-ruling::missing-statement
verdict: confirmed
evidence: executed=git log --format=%H%n%B main..2a76f3d | grep -iE 'direction \(b\)|marker.alone|header half|rejected' ; gh pr view 712 --json body ; gh pr diff 712 (SPEC hunk read in full) mode=bash -c matches-artifact-mode=yes result=zero hits across all three commit messages; the PR body's Plan names direction (a) only and `## Alternatives considered` is the planner/challenger contest record, not a direction ruling; the SPEC §4.13 hunk adds the negative face and the shape reference and rules no direction; `.claude/agents/finding-judge.md` states the shape, never the rejected alternative
action: fix-now-cheap
remedy-survives: yes
axis: implicit↔explicit
target-position: high
justification: the Issue thread does NOT satisfy the clause, on the thread's own terms — the first trusted-author comment carries direction (b)'s measurement ("Marker-alone still excludes the 'merely mentions the header text' lookalike but admits 'marker without the header', a class that contains both PR #706 round 1 and a real false positive — a trusted-author comment quoting a concrete marker") and then says verbatim: "Left as a comment because it is the reviewer's reasoning about the ruling, not a fact the body must carry." AC 2 is a criterion on the change ("Exactly one closure direction is ruled"), and pre-existing thread reasoning by a reviewer is not a ruling this change makes. Bounded both ends: FLOOR — one sentence naming (b) rejected with the measured cost, citing the Issue comment as the ground rather than re-deriving it; CEILING — it must not restate the header string or regex (the shape stays single-sourced; §183e cuts those literals at file level and §9 is the whole point of direction (a)). Independent ground: an unmet acceptance criterion, not a prior failure's direction.
swing: not-evaluated
```

```
finding: F4 — §183g's comment claims occurrence-bounding; grep -cF bounds lines
site: scripts/test/smoke.d/70-gates-contentlocks.sh#s183_derivation_flags
axis-key: scripts/test/smoke.d/70-gates-contentlocks.sh#§183g::comment-accuracy
verdict: confirmed
evidence: executed=printf 'x <retired> y <retired> z\n' > f4.txt ; grep -cF -- '<retired>' f4.txt ; grep -oF -- '<retired>' f4.txt | wc -l mode=script-file matches-artifact-mode=yes result=1 vs 2 — two occurrences on one line count as one; the comment at :5905–5906 reads "bounds how often the retired string may appear in the file at all"
action: fix-now-cheap
remedy-survives: yes
axis: none — discrete (the comment either states the guarantee the code gives or it does not; there is no quantity to position — the alternative remedy, switching to occurrence counting, would change the guarantee rather than move it along an axis, and is not needed: reinstating the retired form *as* the derivation is caught by the `count-form-is-the-derivation` flag at :5805)
target-position: n/a
justification: n/a (observe-only per SPEC §6.0 P3 — a comment correction, recorded not gated)
swing: n/a
```

```
finding: F5 — the placeholder rationale describes the retired counting mechanism
site: .claude/agents/finding-judge.md#Judged-list-comment-shape
axis-key: .claude/agents/finding-judge.md#Judged-list-comment-shape::stale-mechanism
verdict: confirmed
evidence: executed=git show 2a76f3d:.claude/agents/finding-judge.md | sed -n '26p;30p;180p' mode=bash -c matches-artifact-mode=yes result=:180 reads "read-recipe step 3 reads `round=` off **markers**, not comments, so a concrete marker … is quotable into a judged-list comment and counts a second time there", while :30 (step 3) reads "the maximum of the `round=` values the canonical comments carry, never a count of comments or of markers"
action: fix-now-cheap
remedy-survives: yes
axis: none — discrete (a mechanism description is accurate under the new derivation or it is not; the hazard's existence is not in dispute and the remedy has no tunable position)
target-position: n/a
justification: n/a (observe-only per §6.0 P3). Two defects in one sentence, both to be repaired together: "counts a second time" is count-form semantics, where the live hazard is that a quoted concrete marker inside an otherwise canonical comment **raises the max**; and "off markers, not comments" reads against step 2's canonicity requirement, which is a property of the comment. §1.10(c) recurrence is `not-evaluated` — no resolved prior round.
swing: n/a
```

```
finding: F6a — zero canonical comments plus a historical non-canonical one claiming round=1 still re-issues 1
site: .claude/agents/finding-judge.md#read-recipe-step-3
axis-key: .claude/agents/finding-judge.md#read-recipe-step-3::residual
verdict: confirmed
evidence: executed=case analysis over both derivations, read at git show 2a76f3d:.claude/agents/finding-judge.md sed -n '30p;104,115p' mode=bash -c matches-artifact-mode=yes result=with zero canonical comments, count form = 1+0 = 1 and max form = stated fallback 1 — IDENTICAL; the divergence appears only at ≥1 canonical comment. PR #706 round 2 is exactly the zero-canonical case, so the residual is real and demonstrated, and unchanged by this PR. Slot-mismatch sub-claim also confirmed: step 2 never cross-checks the header round against the marker round, `max` reads `round=` off the marker, and the :112 clause suppresses the report whenever step 4 still selects an `<N>@<sha>` — which is the common case.
action: follow-up-issue
remedy-survives: n/a
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
```

```
finding: F6b — sub-claim: the PR body's "ran short and re-issued … whenever a historical non-canonical comment existed" overstates the repair
site: PR #712 body#Plan (and .claude/agents/finding-judge.md#read-recipe-step-3, which carries the same sentence)
axis-key: pr-712-body#Plan::derivation-claim
verdict: refuted
evidence: executed=case analysis of the count form over three comment sets (non-canonical only; non-canonical + canonical(2); canonical(1,2) + non-canonical(3)) mode=bash -c matches-artifact-mode=yes result=the count form re-issues a number already claimed in ALL three, including the zero-canonical case — so the sentence, which asserts a property of the **retired** form only and makes no completeness claim about `max`, is literally true as written. Step 3 additionally states the fallback in the open ("no canonical comment → round 1"), so nothing is claimed away.
action: no-action
remedy-survives: n/a
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
```

---

## Durable fix list, ordered by action

**fix-now**

1. **F1 — `scripts/test/smoke.d/70-gates-contentlocks.sh:5756-5764` (`s183_producer_flags`), with `:5631` (`S183_COPY_KEY`) and the arm comment at `:5841-5859`.** The arm must red all three witnessed mutants. Required: derive the envelope field keys from the `comment-envelope` slice (guarded: fewer than two keys reds a named flag) and require both on the producer line; bind each key to its position with the position phrases lifted from `.claude/agents/finding-judge.md`'s own "The caller posts the value of `comment-header:` … as the comment's **first line** and the value of `comment-marker:` … as its **last content line**" sentence, order-sensitive; replace the hardcoded bare adverb at `:5631` with a lift of the same sentence's prohibition clause. Do not pin whole sentences and do not move any header literal into a caller (that reds `183d`/`183e`). Correct `:5859`'s `ng` text to what the arm then actually guarantees, and state in the `:5841` comment that the arm detects documentation regression, not runtime composition.

2. **F2 — `scripts/test/smoke.d/70-gates-contentlocks.sh:5654` (`s183_rule_field`) / `:5733-5754` (`s183_roundtrip_flags`).** Make the round trip total over read-recipe step 2 (`.claude/agents/finding-judge.md:26`): every inline-code span on that line must be either satisfied by both producer surfaces or a substring of a span that is; anything else reds a named flag. Guard the span lift non-empty. Do **not** implement the reviewer's "exactly two spans" count — step 2 carries four spans at head. Verified satisfiable at head and red on the `^judged-by: finding-judge$` mutant.

**fix-now-cheap**

3. **F3 — PR #712 body, `## Alternatives considered` (or `## Plan`).** Add one sentence ruling direction (b) rejected with its cost: marker-alone still excludes the "merely mentions the header text" lookalike but admits "marker without the header", a class containing PR #706's round-1 comment and a trusted-author comment quoting a concrete marker — the precedent `.claude/commands/file-review.md` had to invent placeholder-form quoting for. Cite Issue #711's first comment as the measurement; restate no header string or regex.

4. **F5 — `.claude/agents/finding-judge.md:180`.** Restate the placeholder rationale in `max` terms: step 3 takes the maximum of the `round=` values canonical comments carry, so a concrete marker quoted into an otherwise canonical comment **raises that maximum** — which is why the marker appears here in placeholder form. Drop "counts a second time" and "off markers, not comments".

5. **F4 — `scripts/test/smoke.d/70-gates-contentlocks.sh:5905-5906`.** Replace "bounds how often the retired string may appear in the file at all" with what `grep -cF` measures: bounds how many **lines** may carry the retired string. No implementation change needed.

**follow-up-issue**

6. **F6a — `.claude/agents/finding-judge.md:30` and `:110-112`.** File one issue covering both residuals: (i) zero canonical comments beside a non-canonical comment claiming `round=1` re-issues 1 — pre-existing, identical under the retired form, demonstrated by PR #706 round 2, so not a regression of this PR; and (ii) `unresolved — slot-mismatch` names a case the derivation does not guard and the `:112` "changes the selected prior round" clause suppresses in the common case — i.e. a field locked in place by `183a` that may be near-unreachable. Both are contract-design questions past this Issue's AC scope.

**no-action**

7. **F6b** — refuted, retained in the record with the case analysis. Not convertible to an issue; next round it reads back as `already-refuted (round 1)`.

## Notes from the judge, retained verbatim

- Not a judged record, offered as an observation: `S183_COPY_KEY='verbatim'` (`:5631`) is a hardcoded expectation about producer prose, which sits slightly against the PR body's checklist line "Nothing the arms check is hardcoded in the suite". Fix item 1 removes it as a side effect.
- Verdict token untouched: `code-reviewer`'s `ship after fix` stands as written.
- Read recipe run on this PR: one trusted-author comment ("## Review record — round 1, and the open decision"), zero canonical markers → current round 1, `prior-round: none (first round)`.

ilgyu-yi:
## Direction decision — 2026-08-25

The maintainer ruled, in-session on 2026-08-25, that this PR re-targets #711's closure from the caller-copy instruction + §183 producer arm to a **shared posting script** (envelope validation, round derivation, and posting in code both the judge's read recipe and the callers invoke), with normal script tests replacing the phrase-derived arms.

Consequences for the round-1 fix list (previous comment):
- Items 1–2 (F1/F2 arm strengthening) are **superseded** by the route change — the arms they strengthen are being replaced, not repaired.
- Items 3–5 (F3/F4/F5) resolve inside the rework of the same surfaces.
- Item 6 (F6a follow-up issue) stands and will be filed.

Ground for the deviation: Directive #637's revision of the same date (authoring doctrine; contracts move to code, no phrase-pinning arms). The Doc-phase assets at `43bf616` (shape stated in the SSOT; judge-rendered envelope) carry forward.

ilgyu-yi:
## Finding triage (round 2)

reviewed-head: ff2f4261088e886219b7080cf88f2df78210091b

Read recipe run at the head's own instructions (script extracted from the head blob into scratch, executed there): one canonical comment on PR #712, `round=1 head=2a76f3dad8858e054e8baf6959aaefe71b9dda1d`, `next=2` → current round 2, prior round resolved as `1@​2a76f3d`. All reproductions ran out-of-tree: the head's `scripts/ghjig_judged_list.sh` copied into session scratch, cwd and every write inside scratch, `gh` stubbed via a scratch PATH prepend for the synthetic histories. No live file was mutated; no raw unfiltered comment read was performed.

```
finding: F1 — the two shape-degradation reasons in finding-judge.md claim an observability this PR's own rework removed
site: .claude/agents/finding-judge.md#check-5-prior-round-tri-state
axis-key: .claude/agents/finding-judge.md#check-5-prior-round-tri-state::observability-claim
verdict: confirmed
evidence: executed=bash <scratch>/head-scripts/ghjig_judged_list.sh rounds 706 (live history) ; same script with PATH-stubbed gh over two synthetic histories: (A) a single trusted comment shaped like PR #706's round 1 — marker-form last line claiming round 1, non-canonical header; (B) a single trusted comment whose header says round 2 while its marker-form last line says round 1 (slot-mismatch) mode=script-file matches-artifact-mode=yes result=live: only rounds 2 and 3 print, the marker-without-header comment is invisible with no anomaly signal; stub A: zero facts, next=1, rc 0 — the recipe-following judge reports `prior-round: none (first round)`, the exact silent degradation Issue #711 opens with; stub B: zero facts, next=1, rc 0 — the slot-mismatch comment is silently non-canonical (canon_fact returns 1 at its hn!=mn check, printing nothing). Secondary defects verified by reading git show ff2f426:.claude/agents/finding-judge.md — step 2 (:21) states no "header half" (the halves moved into the script's file header), so :104's "fails the header half of step 2" dangles; and since neither anomaly ever surfaces through `rounds`/`show`, :107's "read that off step 4's own output" can never observe a changed selection
action: fix-now
remedy-survives: yes
axis: fewer↔more
target-position: low
justification: independent ground is three measurements, not a prior failure's direction: through the sanctioned recipe neither `unresolved — non-canonical-marker` nor `unresolved — slot-mismatch` can ever fire, so the contract text is false at head, and honoring it would require the raw unfiltered read step 1 names as the injection channel. Deletion-default per Directive #637 L5 is a policy ground. Bounded both ends: FLOOR — remove or rebind only the false block (:103–:105, the :107 fire-condition sentence, and the dependent no-PR sentence at :109 that says "both reasons"), never the tri-state itself — `unresolved — duplicate` is measured-real (the script's distinct exit 3) and stays; CEILING — if the author takes the rebind branch (:107 already legalizes named overrides), the two reasons must be explicitly scoped to an override that did its own trusted-filtered read, never presented as recipe-observable. Sync note for the author: the PR body's Key context bullet "keeps … the two `prior-round:` reasons" must move with whichever branch is taken.
prior-round: 1@​2a76f3dad8858e054e8baf6959aaefe71b9dda1d
swing: none
```

Boundary against round 1's item 6: this is not a re-litigation of F6a. F6a deferred the *design* residuals (zero-canonical re-issue; near-unreachable slot-mismatch under the pre-rework prose recipe); F1 is contract text this PR ships asserting an observability the same diff deleted — a defect born in the rework, inside this PR's scope. The reviewer drew that boundary correctly.

```
finding: F2 — the PR body's "filed separately per the round-1 judged list's item 6" has no referent
site: PR #712 body#Out-of-scope
axis-key: pr-712-body#Out-of-scope::filed-separately-claim
verdict: confirmed
evidence: executed=gh issue list --state all --search "judged"|"canonical"|"re-issues" (three sweeps, open+closed) ; gh pr view 712 --json comments piped to grep for the filing claim mode=bash -c matches-artifact-mode=yes result=no issue in either state matches the round-derivation residual; the direction-decision comment itself says "Item 6 (F6a follow-up issue) stands and will be filed" — future tense. "filed separately" is false at head, a §1.10(d)-class self-claim shipped without its evidence
action: fix-now
remedy-survives: yes
axis: none — discrete (a referent either exists and is cited, or the sentence is reworded to intent; there is no continuum to occupy)
target-position: n/a
justification: n/a (discrete; observe-only per §6.0 P3). Both branches of the reviewer's remedy survive; the filing branch is preferable because it also discharges round-1 item 6, which is otherwise still outstanding — but rewording to intent is equally truth-restoring. Check-6 recurrence: claim-class, but its axis key was not confirmed in the resolved prior round (round 1's PR-body items were a different anchor, and F6b was refuted, which per check 2 may not be read back as recurrence) — instance correction survives.
prior-round: 1@​2a76f3dad8858e054e8baf6959aaefe71b9dda1d
swing: n/a
```

```
finding: F3 — one stray trailing blank line is the whole net diff in 70-gates-contentlocks.sh
site: scripts/test/smoke.d/70-gates-contentlocks.sh#EOF
axis-key: scripts/test/smoke.d/70-gates-contentlocks.sh#EOF::whitespace-leftover
verdict: confirmed
evidence: executed=git diff 45d497b..ff2f426 -- scripts/test/smoke.d/70-gates-contentlocks.sh mode=bash -c matches-artifact-mode=yes result=exactly one hunk, one added empty line after the 182d arm — rework leftover; the file otherwise reverted to base when §183 moved to 71-judged-list.sh
action: fix-now
remedy-survives: yes
axis: none — discrete (a whitespace-only hunk is present or absent)
target-position: n/a
justification: n/a (discrete). Drop it: it makes the PR's changed-file census list a file with no change, and the removal folds into the F1 fix commit at zero cost (#637 subtraction-first).
prior-round: 1@​2a76f3dad8858e054e8baf6959aaefe71b9dda1d
swing: n/a
```

```
finding: F4 — a CRLF hand-authored comment defeats the $-anchored shape regexes and reads as non-canonical
site: scripts/ghjig_judged_list.sh#canon_fact
axis-key: scripts/ghjig_judged_list.sh#canon_fact::crlf-edge
verdict: confirmed
evidence: executed=printf '## Finding triage (round 2)\r\n' | /usr/bin/grep -cE "$HDR_RE" (and the marker-regex equivalent) mode=bash -c matches-artifact-mode=yes result=0 matches under BSD grep (GNU grep behaves the same) — a trailing \r defeats both anchors, exactly as observed. Nuance found while measuring: the first run matched (count 1) because this machine's PATH grep is ugrep 7.8.4, which is CRLF-tolerant — so the edge's bite is PATH-dependent (bites under BSD/GNU grep, i.e. CI and stock shells; not under ugrep). The wrong-mode first measurement was discarded and re-run against /usr/bin/grep per the mode-aware grammar
action: drop
remedy-survives: n/a
axis: n/a
target-position: n/a
justification: n/a — the reviewer demands no action and the design absorbs the edge: the script is the only sanctioned producer and posts LF via --body-file; a CRLF hand-authored comment degrades to non-canonical, the same safe direction as any lookalike. Worth one line in the item-6 residual issue when it is filed (F2's filing branch), since it is another member of the silently-filtered class — offered as a note for that issue's author, not as a new judged record.
prior-round: 1@​2a76f3dad8858e054e8baf6959aaefe71b9dda1d
swing: n/a
```

## Durable fix list, ordered

**fix-now**

1. **F1 — `.claude/agents/finding-judge.md`, check-5 tri-state block.** Delete the false block — the two-reasons paragraph, its fire-condition paragraph, and the dependent no-PR "both reasons" sentence — or rebind both reasons explicitly to the named-override path (an override that did its own trusted-filtered read is the only party that can ever record them). Deletion is the default (#637 L5). Keep the tri-state and `unresolved — duplicate` (measured-real, script exit 3). Sync the PR body's Key context bullet that claims the two reasons are kept.
2. **F2 — PR #712 body, `## Out of scope`.** File the round-1 item-6 residual issue and cite its number (preferred — it also discharges the outstanding item 6), or reword "filed separately" to intent. If filed, consider carrying F4's CRLF edge as one line in that issue.
3. **F3 — `scripts/test/smoke.d/70-gates-contentlocks.sh`.** Remove the stray trailing blank line; fold into the same fix commit.

**drop**

4. **F4** — confirmed as a fact, design-consistent, no action; retained here with the corrected-mode measurement (BSD/GNU grep reds, ambient ugrep tolerates — PATH-dependent bite).

## Notes for the caller

- Verdict token untouched: `code-reviewer`'s `ship after fix` stands as written.
- Anti-swing: prior round resolved (`1@​2a76f3d`); no fix-now item shares an axis key with a round-1 item, so no opposite-end flag. Round 1's smoke-arm keys (`s183_*`) died with the file move to `71-judged-list.sh` and do not match F3's key.
- Observation, not a judged record: this judgment's own evidence fixtures necessarily contained concrete marker lines; they live only in scratch and are described, never quoted, above — the `post` forgery guard is why.

dispatch: findings produced by the ship's code-reviewer invocation; judgment by a separately dispatched finding-judge — recorded by the caller per the limb-(a) contract.

ilgyu-yi (commented):
Verdict: approve — code-reviewer `ship` at this head; the round-2 fix commit deletes exactly the false recipe-observability block with zero dangling references, every PR-body claim re-measured true at head (suite 1373/0, lint clean, censuses hold), and the composed security review found no High/Medium/Low findings on the posting script. Advisory idiom notes and the deferred residuals are carried by Issue #713.

Issue #715, opened 2026-08-25
post's INT/TERM traps orphan the backgrounded gh, and two carrier comments still name MAINTAINER


ilgyu-yi:
## What

Two residuals from PR #714's round-1 judged list (both ruled `defer-to-issue` there; the evidence — an executed TERM-orphan repro and the two head-blob prose reads — is pinned in that comment):

1. **Orphaned gh child on INT/TERM.** `scripts/ghjig_judged_list.sh`'s `cmd_post` traps INT/TERM to exit through the EXIT cleanup, but the backgrounded `gh pr comment` survives the script's death; the judge's repro landed the post AFTER the script exited 143. A caller that trusts the death code and re-runs `post` mints the next round — never a duplicate canonical round (the exit-3 refusal holds), but "script failed" no longer implies "no post landed". Remedy shape, judge-verified: `trap 'kill $! 2>/dev/null; exit 130' INT` plus the TERM twin (single quotes defer `$!` to signal time). This closes the orphan's survival, not the in-flight network race.
2. **Prose-code drift beside the changed literal.** The comment lines at `scripts/ac_closeout.sh:10` and `.claude/hooks/helpers/ac_closeout_gate.sh:354` still enumerate the comment-trust set with `MAINTAINER` while the adjacent `-q` literal reads `OWNER`/`MEMBER`/`COLLABORATOR` — PR #714 dropped the inert token from the literal only, honoring its activated byte-identity scope. Deletion-default repair (Directive #637 L5): drop the token, or the whole prose enumeration — the code literal beside it is the SSOT.

## Why

Item 1 leaves the one residual wedge on the post path: ambiguity about whether a post landed after an interrupt, which costs a hand-inspection every time it fires. Item 2 is living-set prose one line from the code it contradicts — all three of PR #714's security votes independently named the restore-the-token editing trap it invites. Both were measured on that PR's review round; each remedy is a few lines with existing §183 idioms to lock them. No open issue owns either (the judged list's `defer-to-issue` action is discharged by this filing), and #713 closes with that PR, so a comment there reaches no flow.

## MISSION fit

MISSION "The mechanism": the judged-list layer is the review phase's durable manifest — ambiguity about whether a post landed degrades it, and a living-set comment contradicting adjacent code is the drift class the subtraction doctrine exists to delete.

## Acceptance criteria

[...]

PR #677, opened 2026-08-18
feat(#676): resolve a body's quotations at the file they are attributed to

ilgyu-yi:
Closes #676

## Goal

The shell binds its reviewers and judges to evidence-with-command — `.claude/agents/finding-judge.md:60`: *"A verdict with no command is not a verdict — read the source at the head, run the reproduction, and record what you executed."* Nothing binds the **author** of a durable-artifact body on whether a quotation is real. Verification is entirely downstream: the author writes prose claims about artifacts, the reviewer re-derives each one with commands, and every re-derivation costs a round.

## How this serves the mission

`MISSION.md:26` — *"a Claude Code session running in `unattended` mode against a real engineering issue produces a merged PR whose quality matches what a careful senior engineer would have produced in the same hours."* That closing clause is a **cost** bound. This PR moves a decidable class of defect from reviewer-round cost to pre-dispatch cost. In §6.0 terms it supplies the **positive face** of a pair whose negative face (the reviewer gate) already exists.

## Plan — merged from a two-challenger contest

`planner` produced Plan A; two mutually-blind `plan-challenger`s ran on **correctness** and **cost/effort**; `plan-reviewer` judged, issued one `refine`, and shipped the merged Plan M.

**`A-stands` was unavailable — A's two load-bearing numbers were falsified by independent re-measurement.**

| A claimed | Re-measured |
|---|---|
| 18 plain-quote spans in #676, zero of them citations | **3**, and one (`"expensive direction"`) **is** a citation resolving at `SPEC.md:1231` |
| attribution 17/17, zero mis-attributions | **15/17** — the ≤3-lines-back fallback attributes two schematic placeholders to `CHANGELOG.md`, defeating A's own bound |

**#676 is an outlier, not a calibration set.** Star spans (`*"…"*`) vs plain, per body: #676 = 17/3; every other body = 0–4 star, 5–18 plain (15 vs 83 across ten). **#639 has zero star spans** — A's rule extracts nothing from #639 and #664.

**What merged:**

- **B1** — extraction widened to `*"…"*` **and** plain `"…"`; attribution **same-line only**; a bare `` `:NN` `` inherits the last path on its line; SPEC part (a) keys on **attribution**, not on a delimiter.
- **B1 + B2 convergent cut** — delete the ≤3-lines-back fallback. Measured: it fires **twice in 32 spans and is wrong both times**. Two challengers, different axes, same mechanism.
[...]

PR #722, opened 2026-08-26
feat(#721): ship the rounds instrument

### LINKED ISSUE #721: the rounds-to-merge trend is asserted from memory, never printed by an instrument
Parent Directive: #637

## What

Directive #637's motivating metric — review rounds per PR, and its trend toward merge — has no instrument: every rounds figure in the repo's record so far was hand-asserted from memory (the exact born-false channel the parent's L2 closes). The canonical substrate exists — `scripts/ghjig_judged_list.sh rounds <pr>` derives one PR's canonical rounds from the trusted judged-list markers — but nothing aggregates it across PRs, so the trend the parent's completion review must read ("printed at runtime, never asserted", item 7) cannot be printed.

Delivery, per the parent's item 7: **one rounds instrument** — a script that, for a given set of merged PRs (an explicit PR list, or a recent-merged window), prints the canonical round count per PR and the sequence as a trend line. It reads rounds ONLY through `scripts/ghjig_judged_list.sh rounds` (the single code home — L4: shared code both sides call, no second marker parser), gates nothing, and is an observed-outcome reporter: the parent's completion review reads its output; nothing else consumes it.

## Why

The parent names this instrument as the way its own success is measured — "The rounds instrument discriminates" is a listed success signal, and `/complete-directive 637` must read the rounds-to-merge trend from it rather than from an asserted figure. Every week it is absent, each new merged PR's round count exists only as marker comments no one aggregates, and any trend claim in a completion body would have to be hand-carried — the defect class this Directive exists to end, in its own completion evidence. No existing surface covers it: `ghjig_judged_list.sh` is single-PR by contract, and no other script reads the markers.

## MISSION fit

MISSION "The mechanism": the judged-list markers are durable memory; this instrument is the selective-injection reader that turns them into a runtime-printed fact, so the Directive's own record can obey L2.

## Acceptance criteria

- [ ] AC1 — the instrument prints, for each PR in its input set, `pr=<N> rounds=<K>` derived solely via `scripts/ghjig_judged_list.sh rounds` (no second parser of the marker grammar anywhere in the script), plus one trend line over the sequence; a fixture set carrying canonical markers for rounds 1..N yields `rounds=N`, and a set containing only non-canonical lookalikes yields `rounds=0` (the parent's two-sided signal).
- [ ] AC2 — the instrument is advisory and fail-open as a reporter (a PR whose rounds derivation fails is reported as such and does not abort the sweep), and it gates nothing — no exit-code consumer, no hook, no CI wiring.
- [ ] AC3 — input is explicit: a PR-number list argument, or a `--recent <M>` window resolved via `gh pr list --state merged`; the resolution command is stated in the script header (single code home).
- [ ] AC4 — normal-test locks only: smoke arms execute the shipped script against fixture comment sets (reusing the §183/§184 stub idioms); no prose-grepping arm.
- [ ] AC5 — full smoke suite green.

## Priority

P1

## Out of scope

- Any gating or escalation on the count — that axis is Directive #639's (its caller-owned ladder and per-change attribution substrate); this instrument prints a trend and neither selects nor preempts #639's substrate choice (the parent's own non-goal, restated here as the boundary).
- The parent's other items (1–6, 8) — siblings; item 1 (the doctrine section) merged via #720.
- Changing `scripts/ghjig_judged_list.sh` — this issue adds a consumer; residuals on that script are #715/#718's.
- SessionStart or advisory wiring — the consumer is `/complete-directive`'s completion review, by hand; wiring a reader into the friction advisory is a separate decision this issue does not open.

## Notes

- Anchors: `scripts/ghjig_judged_list.sh` (the `rounds` mode and its exit grammar — 0 with `round=`/`next=` facts; 3 on a duplicate canonical round), `scripts/test/smoke.d/71-judged-list.sh` (the executable `gh` stub idiom to reuse).
- Source: Directive #637 artifact-level boundary item 7; the parent's Non-goals bound the instrument against #639 explicitly ("it gates nothing, is not a reject count").

ilgyu-yi:
Closes #721

## Goal
Ship `scripts/ghjig_rounds.sh` — Directive #637 item 7's rounds instrument: it reports the judged-round count per PR and the sequence as a trend, so the Directive's motivating metric is read at runtime instead of asserted from memory.

MISSION "The mechanism": the judged-list markers are durable memory; this instrument is the selective-injection reader that turns them into a runtime-printed fact, so the Directive's own completion record can obey §1.11 L2 (the author never carries facts).

## Plan (A′ — the composition the plan-reviewer ruled)

Ship **`scripts/ghjig_rounds.sh`** — a fail-open, gate-nothing reporter aggregating per-PR canonical round counts across a PR set, reading rounds **only** through a child invocation of `scripts/ghjig_judged_list.sh rounds` (sibling-relative `BASH_SOURCE` self-location, the repo idiom; no PATH lookup, no marker-grammar literal anywhere in the new script).

**Derivation — `rounds = next − 1`.** The child prints `next=$((max + 1))` over its `max_round()`, so `next − 1` is the ordinal of the last judged round — the rounds-to-merge metric itself. Counting `round=` fact lines is only a proxy: because `post` always derives `next = max+1`, live numbering is contiguous, so a gap can arise only from a **deleted** marker, under which the count *undercounts* ({1,3} → 2) and the instrument shipped to end false figures would print one. `next − 1` also yields `0` on a lookalike-only set, satisfying the parent's two-sided signal, and reuses the code home's derivation (§1.11 L4).

**Output grammar.** Per PR one line `pr=<N> rounds=<K>`; child exit 3 → `rounds=ambiguous(duplicate)`; any other non-zero → `rounds=error(rc=<rc>)`; the sweep continues and the instrument exits 0. The two error forms stay distinct: exit 3 is an operator-repairable ambiguous history, not generic failure. One trailing `trend: <tok> <tok> …` line, one token per PR in output order, `?` for an ambiguous/error entry so no entry is silently dropped. Optional stderr `anomaly:` note when the fact-line count disagrees with `next − 1` (a deleted marker) — stderr only, never touching the exit code.

**Input — merge-ordered.** An explicit PR-number list prints in the given order. `--recent <M>` resolves the window in **merge order**, which `gh pr list` does not return: measured on this repo's last 100 merged PRs, the listing is createdAt-descending with five mergedAt inversions (#677/#675, #641/#635, #603/#601, #588/#582, #562/#561), and window *membership* is biased the same way — #675 (created 08-17, merged 08-21) sits below four PRs that merged earlier, so at a window boundary near that position a plain `--limit M` omits a more-recently-merged PR while keeping earlier ones, flattering exactly the metric this instrument reports. So: **over-fetch, sort by `mergedAt` descending, take M, emit oldest→newest**. The resolution command is stated verbatim in the header (single code home). On the explicit-list route the caller owns the order, and the header says so: that trend is caller-ordered, merge-ordered only if the caller made it so.

**Contract home** is the script header (the SPEC §4.13 pattern `ghjig_evidence.sh` follows). SPEC §8 gains the tree line and the count bump `27 → 28` — **test-forced**, not merely SSOT discipline: smoke §103c enumerates top-level `*.sh` leaf-by-leaf and pins the header integer. Zero edits to SPEC §1.10/§1.11.

**`complete-directive.md` untouched.** It is named nowhere in #637, so it is not one of its carriers; the ACs require no consumer wiring, and the parent body — which the completion reviewer receives verbatim — already names the instrument obligation.

**Phases** (`feat`, strict): Doc = executable header-contract stub (body exits 1) encoding the mergedAt-ascending resolution and the `next − 1` derivation, plus the SPEC §8 surfaces; Test = `scripts/test/smoke.d/73-rounds.sh` (§185) wired after §184, red against the stub, each arm's comment stating the invariant asserted rather than the transient stub state; Code = implementation to green.

**Fixtures** reuse the §183 executable-PATH `gh` stub idiom, extended to dispatch `pr view <N>` to per-PR fixture JSON and to serve a `pr list` route.

## Target base

`main`

## Alternatives considered
- **Plan A** (base): the skeleton above, but with `rounds` = the count of the child's `round=` fact lines and `--recent` as a plain reversal of `gh pr list`.
- **B1** (axis: correctness): dominates-A — measured live that `gh pr list --state merged` returns createdAt-descending (five mergedAt inversions in this repo's last 100 merges), so A's reversal can print an inverted trend and A's stub would lock that assumption green; and showed the count derivation is a proxy that undercounts on a deleted marker where the code home's own `next − 1` is the metric.
- **B2** (axis: simplicity): reasoned concession — four subtraction attacks failed (a fourth mode on the judged-list script would mix a fail-open sweep into a fail-closed, parity-locked seam home; `--recent` and the trend line are AC text; a separate smoke file is the lower-coupling choice). Its grammar-unification advisory was rejected: collapsing exit 3 into a generic failure hides an operator-repairable ambiguous history.
- **Verdict**: refine → ship. Round 1 ruled the composition (A's skeleton and two-form error grammar + B1's `next − 1` and mergedAt-ascending sort + B1's fixture arms, plus a merge-ordered window: over-fetch → sort → take M) and refuted B1's own claim that §103c does not check the §8 count — the bump is test-forced. Round 2 confirmed the revised plan carries it, with two advisory prose corrections applied here (a born-false carrier parenthetical deleted; the explicit-list ordering caveat and window-boundary precision added).

## Key context
- `scripts/ghjig_rounds.sh` — the deliverable; header = contract home
- `scripts/ghjig_judged_list.sh` — the code home consumed as a child (`rounds` mode; `next=` is max+1; exit 3 = duplicate canonical round); unmodified
- `scripts/test/smoke.d/71-judged-list.sh` — the executable-PATH `gh` stub idiom §185 extends
- `scripts/test/smoke.d/50-perproject-recall.sh` §103c — enumerates top-level scripts and pins the count integer
- `SPEC.md` §8 — tree line + count `27 → 28`, TOC regenerated

## Checklist
- [x] **Doc**: `scripts/ghjig_rounds.sh` as an executable header-contract stub carrying the full contract; SPEC §8 tree line + count `27 → 28` + TOC regen; `changelog_unreleased/added/721.md` (AC3's header-stated resolution command, AC2's gates-nothing/fail-open statement)
- [x] **Test**: `scripts/test/smoke.d/73-rounds.sh` (§185) wired after §184, red against the stub — the 1..N canonical set → `rounds=N`; the lookalike-only set → `rounds=0`; the gap set `{1,3}` → `rounds=3`; exit-3 → `rounds=ambiguous(duplicate)` with the sweep continuing; a failing PR → `rounds=error(rc=…)` with later PRs still reported and exit 0; an inverted-`mergedAt` `pr list` fixture → the trend in merge order; a `#675`-shaped late-merged/early-created PR entering the window; the structural no-marker-literal arm (AC1, AC2, AC3, AC4)
- [x] **Code**: implement the two input routes, the child invocation with rc capture, the `next − 1` derivation, the anomaly note, and the trend emission; §185 green
- [x] **Code**: full smoke suite green, with the run rendered into this body via `scripts/ghjig_evidence.sh` (AC5)

## Out of scope
Any gating, escalation, exit-code consumer, hook or CI wiring (#639's axis); any change to `scripts/ghjig_judged_list.sh` (its residuals are #715/#718's); `complete-directive.md` and any consumer or SessionStart wiring; SPEC §1.10/§1.11 text; #637's sibling items 1–6, 8.

## Risks
- **Fail-open masking**: a wholly-down `gh` yields all-`error` lines and exit 0 by design; the mitigation is that every failure is *reported* rather than blank, and the consumer reads output, not an exit code.
- **Parser-duplication drift**: a future edit could inline marker parsing; the structural no-literal arm plus the two-sided lookalike fixture make that regress red.
- **Window residual**: membership is corrected by sorting an over-fetched window, so a PR merged long after a very large batch could still fall outside the over-fetch; the header states the over-fetch factor.
- No security surface: read-only `gh`, no state writes, no network beyond `gh`.
- **`grep -c` under `set -euo pipefail`** returns 1 on zero matches, which would abort the sweep on the legitimate zero-rounds case; handled with `|| facts=0`. Noted for any future arm or edit on this route.

## Open questions
None blocking.

## Decisions
- 2026-08-26: contest ruled the composition (round 1 `refine` naming it, round 2 `ship` confirming it); rejected B2's grammar unification, A's plain reversal, A's count-of-lines derivation, and B1's refuted §103c rationale.
- 2026-08-26: the SPEC §8 count bump is test-forced (§103c pins the integer and enumerates the leaves), so it rides the Doc commit with the TOC regen the tree insertion requires.
- 2026-08-26 (Code phase, header kept accurate): the Doc header left the over-fetch unquantified, so the Code commit named the concrete bound — 4×M, floor 20, cap 200, never below M — and added the one grammar case the Doc phase had not covered: a child that exits 0 without a parsable `next=` reports `error(rc=0)`. No stated guarantee was weakened.
- 2026-08-26 (Code phase, deviation): the child's stderr is forwarded verbatim rather than suppressed — suppressing it would discard the child's own anomaly classes and its `die` reason, which is what makes an `error(rc=1)` line diagnosable. Verified against §185d: the gapped fixture's history is two canonical comments, so the child raises no anomaly and the arm's "exactly one `anomaly:`" assertion still measures only this instrument's line.
- 2026-08-26 (recorded gap): the exit-1 branch for a wholly-unresolvable `--recent` window is contract-only — no §185 arm runs `--recent` against an absent/empty listing, so that path is stated in the header and unlocked by a test. Named here rather than left silent; a candidate follow-up if #721's exit-code table warrants full coverage.

- 2026-08-26 (judged rounds 1 and 2, both durable on this PR): round 1 ruled six fix-now items and round 2 four more. Landed: the two unboundedness rationale clauses and the M-exactness claim deleted at every site, the exit-1 line generalized to anything that prevented the sweep from starting rather than enumerating preflights, the header stating that the child's stderr is forwarded verbatim without `pr=` attribution, `M` lifted with `10#` before any arithmetic, every explicit-list argument validated before the sweep emits, the not-a-listing-prefix clause deleted at both its sites once the band was measured, the header rewrapped, and two §185 arms so the argv-order and base-10 behaviors no longer survive mutation. Four dispositions were deferred to #724 and eight findings dropped with grounds; every remedy that rewrote rather than removed was rejected in favour of a deletion.

## Test plan
- [x] §185 arms red against the stub for their stated invariants (7 born-red, 3 born-green; `pass=1400 fail=7` at the Test commit), green after Code
- [x] Full smoke suite green at the final head — rendered via `scripts/ghjig_evidence.sh`:

```
$ bash scripts/test/smoke.sh 2>&1 | tail -1
smoke: pass=1409 fail=0
pin: 5c9306af7c35a72fe890c5c8fcbad0875473a061 2026-08-26Z
```

## Docs touched
- [x] Script header (contract home), SPEC §8 tree line + count + TOC, changelog fragment — no `docs/*.md` pointer (nothing there indexes scripts, and an add with no live function is what §1.11 rule (a) forbids)

## Ship gate
- [x] All tests pass
- [x] code-reviewer passed
- [~] N/A for the round-2 delta — round 1's security pass returned approve at the reviewed head, and the delta since is one file with no security surface added
- [x] Docs in sync
- [x] PR body curated
- [x] CI green


ilgyu-yi:
## Finding triage (round 1)

reviewed-head: e7dbbe39a12772dea0a9ab5e28f1825812f2c922

**Artifact provenance.** Head resolved independently via `gh pr view 722 --json headRefOid`. Both scripts read as blobs at that sha (`git show <sha>:scripts/…`) and extracted into a scratch directory; every reproduction ran with cwd and all writes inside scratch, behind a PATH `gh` stub, invoked as a **script file** (`bash <file>`) — the artifact's own mode. No PR base was checked out; no ambient-tree write.

**prior-round resolution.** `scripts/ghjig_judged_list.sh rounds 722` → `next=1`, no `round=<N> head=<sha>` fact lines, no `anomaly:` lines. No canonical judged comment exists, so this is round 1: `prior-round: none (first round)` on every record, and every `swing:` is `not-evaluated` (a `swing: none` here would be vacuous). Check 6 (recurrence) is therefore `not-evaluated` for every item — no axis key can have been confirmed in a resolved prior round.

**Dedupe.** 4 code-reviewer findings + 4 security Low/Info + 1 advisory-note block → 11 records. Overlaps folded: F1≡Info-1 (J1), F2≡Low-2 (J2), F4≡Low-1 (J4/J5). Attribution is on each `finding:` line.

---

```
finding: J1 — [code-reviewer F1 + security Info-1] the contract home never states that the child's stderr is forwarded verbatim, so the `anomaly:` channel carries child-class lines that name no PR
site: scripts/ghjig_rounds.sh#header-anomaly-channel
axis-key: scripts/ghjig_rounds.sh#header-anomaly-channel::contract-home-omission
verdict: confirmed
evidence: executed=bash scripts/ghjig_rounds.sh --recent 4 (fixtures: pr 60 headerless marker, pr 44 gapped canonical {1,3}, pr 61 header/marker mismatch, pr 42 contiguous 1..3) mode=script-file matches-artifact-mode=yes result=3 lines matched ^anomaly:, of which 2 name no pr= (the child's two classes); only the parent's line carries `pr=44`. Header lines 39-41 define the parent's gap anomaly and assert "anomaly output is never a fact" but say nothing about forwarding; the forwarding decision lives only in the PR body's Decisions entry (a record, not the contract home — §1.11 L4). Confirmed separately: the pr-44 gapped fixture's two comments are both canonical, so §185d's `=1` anomaly assertion measures only the parent's line — the reviewer's caveat is exact.
action: fix-now
remedy-survives: yes
axis: implicit↔explicit
target-position: high
justification: the reviewer offered a disjunction; branch (a) — state the forwarding in the header — survives, branch (b) — prefix forwarded child stderr with `pr=<N>` — does not survive as the required fix. Independent grounds, not directional: (i) §1.11 subtraction rule (a) reaches "modify in place" before "add", and the header already owns an anomaly-channel paragraph (39-41) whose properties are stated for only half the channel's content, so this is a modify, not an add; (ii) branch (a) touches no behavior, so §185d's "exactly one anomaly" arm stays valid by construction, which the finding requires of any fix; (iii) branch (b) changes behavior (the child's stderr is deliberately uncaptured, and capturing it also re-prefixes the child's `die` reason), needing a new AC4 test arm — that is scope the author owns deciding, deferred with J5. Endpoint note: `explicit` here is bounded at documenting the two forwarded classes and their lack of PR attribution, not at documenting the child's whole stderr surface.
swing: not-evaluated
```

```
finding: J2 — [code-reviewer F2 + security Low-2] the over-fetch "cap" does not bound a large M, so the header's two unboundedness rationale clauses are false
site: scripts/ghjig_rounds.sh#resolve_recent
axis-key: scripts/ghjig_rounds.sh#resolve_recent::header-claim-false
verdict: confirmed
evidence: executed=bash scripts/ghjig_rounds.sh --recent <M> for M in 1 5 6 50 51 100 200 201 1000 100000, reading the limit the gh stub received mode=script-file matches-artifact-mode=yes result=limit= 20,20,24,200,200,200,200,201,1000,100000 — `--recent 100000` sent `pr list --state merged --limit 100000`. Precedence at lines 148-151 is factor → floor → cap → never-below-M, so the cap is the operative bound only for 50 < M ≤ 200 (measured: M=51..200 → 200; M=201 → 201; M=50 → 200 by the factor, not the cap; M≤5 → the floor 20). Both reviewers' band arithmetic reproduces exactly. Two sites carry the false rationale, not one: header lines 50-52 ("bounded so one report cannot fetch unboundedly; a merge older than the fetched page is out of reach by that cap, not by a knob") and the inline comment lines 75-76 ("capped so a large M cannot turn one report into an unbounded fetch").
action: fix-now
remedy-survives: yes
axis: fewer↔more
target-position: low
justification: deletion, per §1.11 L5 — and Low-2's alternative branch (clamp `m` to the cap and say the window is capped) does not survive: it would silently truncate an operator-requested window, minting a new claim to guard, where L5 makes removing the claim the default. Independent grounds: no acceptance criterion depends on these clauses (#721 AC3 requires only that the resolution command be stated in the header, and it stays verbatim at line 55); a live contract does not depend on them either (no smoke arm greps them — checked: §185h asserts only `limit > M`, §185i greps shape literals, §185j greps the usage line). Both endpoints stated: delete exactly the two rationale clauses at both sites (low end) while retaining the quantified bound "4×M, floor 20, cap 200, never below M", which the band probe above measures as accurate (high end — the retained text is the floor below which the deletion must not go).
swing: not-evaluated
```

```
finding: J3 — [code-reviewer F3] the exit-code table enumerates two exit-1 causes; three preflight paths also exit 1
site: scripts/ghjig_rounds.sh#header-exit-codes
axis-key: scripts/ghjig_rounds.sh#header-exit-codes::header-claim-incomplete
verdict: confirmed
evidence: executed=bash scripts/ghjig_rounds.sh 42 under three deprived PATHs mode=script-file matches-artifact-mode=yes result=gh absent → "ghjig_rounds: gh CLI not found", exit 1 (line 96); jq absent → "ghjig_rounds: jq not found", exit 1 (line 97); child present but not executable → "the judged-list code home is not executable: …", exit 1 (line 98). None is a usage error or an unresolvable `--recent` window, so the table's disjunction at lines 65-67 is false as a complete enumeration.
action: fix-now
remedy-survives: yes
axis: narrow↔wide
target-position: high
justification: the reviewer offered two branches; the generalization ("1 = anything that prevented the sweep from starting") survives and the "name the preflight class" enumeration branch does not. Independent ground, not directional: an enumeration is add-shaped and re-arms this exact finding the next time a preflight is added — the round generator §1.11 L5 closes — whereas the class is closed under future preflights. Both endpoints stated: the wide end is bounded at "prevented the sweep from starting", which deliberately excludes the per-PR derivation failures that exit 0 (the fail-open property AC2 requires); the narrow end is the present 2-item list, already measured false. §185j stays green under this edit — it greps only `^usage: ghjig_rounds.sh <pr>` and the literal `--recent` in usage() stderr, not the header.
swing: not-evaluated
```

```
finding: J4 — [code-reviewer F4 + security Low-1] the explicit-list route hands every argv element to the child unvalidated, so a crafted argument forges report lines and breaks the documented field grammar
site: scripts/ghjig_rounds.sh#explicit-list-route
axis-key: scripts/ghjig_rounds.sh#explicit-list-route::input-validation
verdict: confirmed
evidence: executed=bash scripts/ghjig_rounds.sh with three argv shapes mode=script-file matches-artifact-mode=yes result=(a) `42 --recent` → `pr=42 rounds=error(rc=1)` / `pr=--recent rounds=error(rc=1)` / `trend: ? ?`, exit 0 — a typo'd flag becomes a reported PR on the success path (`case "${1:-}"` only guards position 1, so `-*) usage` never fires); (b) `'1 rounds=99' 42` → `pr=1 rounds=99 rounds=error(rc=1)` — a consumer reading the first `rounds=` sees 99; (c) `$'1\npr=9 rounds=9'` → stdout exactly `pr=1` / `pr=9 rounds=9 rounds=error(rc=1)` / `trend: ?`, byte-identical to the security reviewer's transcript — a whole fabricated report line. No shell or format-string escape reproduced. `is_num` (line 102) is referenced only at line 165, so it is dead on this route.
action: fix-now
remedy-survives: yes
axis: permissive↔strict
target-position: high
justification: an extreme position justified independently of any prior failure — (i) definitional: the child validates `<pr>` with the byte-identical `is_num` predicate and rejects, so strict is the seam's existing contract, not a new position, and the accepted set is a closed definitional set (non-empty digits), not a chosen point on a continuum, so the two-endpoint rule has no continuous quantity to bind; (ii) measured: three argv shapes above produce forged or grammar-breaking stdout on the exit-0 success path; (iii) the rejection disposition is already contract-named — a non-numeric PR argument is a usage error, which the exit table names and §185j locks, so `is_num "$apr" || usage` needs no new exit semantics. Scope: F4's form (argv on the explicit-list route) survives; Low-1's extension "at `report_pr` entry, applied on both routes" does not survive as stated — see J5.
swing: not-evaluated
```

```
finding: J5 — [security Low-1, second half] the --recent route is equally unvalidated: report_pr trusts jq's `.number`, bounded only by the GitHub schema
site: scripts/ghjig_rounds.sh#report_pr
axis-key: scripts/ghjig_rounds.sh#report_pr::input-validation
verdict: confirmed
evidence: executed=git show e7dbbe39a12772dea0a9ab5e28f1825812f2c922:scripts/ghjig_rounds.sh (lines 113-115, 154-159, 168-173) plus bash scripts/ghjig_rounds.sh --recent 4 against a stub-served listing mode=script-file matches-artifact-mode=yes result=`report_pr "$rpr"` prints `pr=%s` before any numeric test; the only `is_num` on this path is the child's, after the value is already destined for stdout. jq's `select(.number != null)` filters null but not a non-integer. Reaching it requires a compromised or stubbed `gh` — no privilege crossing.
action: defer-to-issue
remedy-survives: n/a
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
```

Deferral target for J5 — **a new carrier issue for `scripts/ghjig_rounds.sh` residuals**, not #715 (residuals on `scripts/ghjig_judged_list.sh`) and not #718 (residuals on `scripts/ghjig_evidence.sh`); the new script has no carrier yet, and the #715/#718 pattern is one carrier per script. Reason for deferral rather than fix-now: a `usage`-shaped rejection on this route (Low-1's literal "both routes at `report_pr` entry") would abort the sweep with exit 1 on an API-supplied value, contradicting AC2's fail-open reporter posture, and neither reviewer states a non-fatal disposition — that disposition is a design decision, not a one-line fix, and it belongs beside J1's branch (b) and the L4 question in the note below.

```
finding: J6 — [security Low-3] a leading-zero M reaches bash arithmetic as an octal literal — the exact hazard the child code home guards, re-introduced in the consumer
site: scripts/ghjig_rounds.sh#resolve_recent
axis-key: scripts/ghjig_rounds.sh#resolve_recent::numeric-lift
verdict: confirmed
evidence: executed=bash scripts/ghjig_rounds.sh --recent 08 and --recent 010, plus bash-mode arithmetic probes mode=script-file matches-artifact-mode=yes result=`--recent 08` → "ghjig_rounds.sh: line 148: 08: value too great for base (error token is \"08\")", exit 1, with no `ghjig_rounds:`-prefixed reason — `is_num "08"` passes and `[ 08 -ge 1 ]` passes because `test` reads base-10. Mode note: a first probe of `$((010*4))` returned 40, measured in the session's zsh (matches-artifact-mode=no) and DISCARDED; re-measured inside a bash script it returns 32, and `0100*4` returns 256. That re-measurement additionally REFUTES the finding's own severity rationale: `--recent 010` does not fail at `--argjson m 010` (jq 1.7.1 accepts it as 10, verified), it succeeds and silently sends `--limit 32` where the header promises 4×M = 40 (contrast: `--recent 10` sends 40). Kinship verified at scripts/ghjig_judged_list.sh:87-90 — `norm_round() { printf '%s\n' "$((10#$1))"; }`, commented "Applied at every lift seam so shell arithmetic never sees an octal literal"; the quotation resolves as attributed.
action: fix-now
remedy-survives: yes
axis: implicit↔explicit
target-position: high
justification: the remedy `m=$((10#$m))` after `is_num` survives and the kinship RAISES the action rather than changing it. Independent grounds: (i) definitional absolute — `is_num` admits `0[0-9]*`, so without an explicit base the accepted input set and the evaluated input set differ by construction; there is no continuum to bound, so the two-endpoint rule is inapplicable, and the position is fixed by the predicate, not by a prior failure's direction; (ii) measured — one input crashes past the `die()` reason contract and another silently narrows the over-fetch to 3.2×M against a header that says 4×M, which is a wrong figure in the instrument built to end wrong figures; (iii) the code home states this rule at its own lift seam, so the consumer is one hop from the guard. Check 6 does not arm (behavior, not claim-class; and no resolved prior round), so this is not a recurrence ruling — the same-class-across-a-seam argument stands on its own. The deeper §1.11 L4 fix (one shared normalizer both scripts invoke) is OUT OF THIS PR by #721's Out of scope, which forbids changing `scripts/ghjig_judged_list.sh`; it goes to the J5 carrier.
swing: not-evaluated
```

```
finding: J7 — [security Low-4] a null mergedAt shortens the --recent window with no notice, against a header that promises the M most recently merged PRs
site: scripts/ghjig_rounds.sh#resolve_recent
axis-key: scripts/ghjig_rounds.sh#resolve_recent::window-shortfall
verdict: confirmed
evidence: executed=bash scripts/ghjig_rounds.sh --recent 3 against a 3-entry listing with one "mergedAt":null mode=script-file matches-artifact-mode=yes result=two PRs reported (`pr=51`, `pr=53`, `trend: ? ?`), exit 0, and `grep -c '^anomaly:\|^notice:'` on stderr = 0. The exactness claim is stated in four places: header line 16, line 48 ("cut to M"), usage line 86, and the `resolve_recent` comment at 143-144.
action: fix-now
remedy-survives: no
axis: fewer↔more
target-position: low
justification: the divergence is real but the reviewer's remedy — add a `notice:`/`anomaly:` stderr class on `resolved count < M` — does not survive, for two independent reasons, neither directional. (i) The finding's "silently" is refuted in part by measurement: the shortfall IS observable on stdout, because the header's own trend contract prints exactly one token per reported PR, so `--recent 3` answering with `trend: ? ?` states the sample size to any reader who counts it. A new stderr class would carry no information the output lacks. (ii) §1.11 L5 orders the prose repair ahead of new behavior, and the added class would also need an AC4 test arm; further, a listing legitimately holding fewer than M merged PRs (a young repo) is not an anomaly, so the remedy as stated fires on a normal state. Both endpoints stated: the low end is bounded at deleting the exactness claim at the four sites above — "up to M" — and must NOT go further: the M cut, the mergedAt ordering, the over-fetch, and the verbatim resolution command all stay (AC3 depends on the last). The high end — a notice channel plus its arm — is what is rejected on ground (i). Author note: §185j stays green under a usage-line rewording, since it greps only `^usage: ghjig_rounds.sh <pr>` and the literal `--recent`.
swing: not-evaluated
```

```
finding: J8 — [security Info-2] gh and jq resolve through PATH via `command -v`, not pinned paths
site: scripts/ghjig_rounds.sh#preflight
axis-key: scripts/ghjig_rounds.sh#preflight::dependency-resolution
verdict: confirmed
evidence: executed=git show <head>:scripts/ghjig_rounds.sh lines 96-97, and every §185-idiom reproduction in this judgment (a PATH-prepended executable `gh` stub) mode=script-file matches-artifact-mode=yes result=`command -v gh` / `command -v jq`; a poisoned PATH does substitute both — and that same property is what makes the stub-based smoke arms and these reproductions possible at all.
action: drop
remedy-survives: n/a
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
```

Drop ground for J8: an accepted repo-wide posture (the child and the harness share it) that the test seam depends on; pinning paths would break §185 and the arms this judgment ran. No header claim is falsified — the reviewer states it as a note, not a defect.

```
finding: J9 — [security Info-3] sibling resolution is symlink-relative, so a symlinked install dies looking for the child beside the symlink
site: scripts/ghjig_rounds.sh#HERE
axis-key: scripts/ghjig_rounds.sh#HERE::self-location
verdict: confirmed
evidence: executed=bash <dirlink>/ghjig_rounds.sh 42 and bash <instdir>/ghjig_rounds.sh 42 mode=script-file matches-artifact-mode=yes result=directory symlink to scripts/ (the repo's `.claude/ghjig-root` install shape) → `pr=42 rounds=3` / `trend: 3`, exit 0 — `pwd -P` resolves THROUGH the directory symlink and the sibling is found; a symlink to the script file itself in a sibling-less directory → "ghjig_rounds: the judged-list code home is not executable: …", exit 1. The reviewer's description is exact, including that the stated "never through PATH" property holds.
action: drop
remedy-survives: n/a
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
```

Drop ground for J9: the failure is loud, named, and non-zero (line 98's preflight), the repo's actual install shape is measured working, and abusing it requires write access to the install directory. No claim is falsified.

```
finding: J10 — [code-reviewer advisory notes] report_pr mixes three concerns in 28 lines; is_num is dead on the explicit-list route; report_pr shares stdin with the --recent heredoc
site: scripts/ghjig_rounds.sh#report_pr
axis-key: scripts/ghjig_rounds.sh#report_pr::idiom
verdict: confirmed
evidence: executed=bash scripts/ghjig_rounds.sh --recent 4 against a 4-PR listing, plus a read of lines 113-141 / 162-181 at the head blob mode=script-file matches-artifact-mode=yes result=all four PRs reported (`trend: 0 3 0 3`), so the child does not consume the heredoc — "safe today" is confirmed by measurement, not by inspection. `report_pr` does span 113-141 and does carry rc classification, `next=` extraction, and grammar emission. `is_num` is referenced only at line 165.
action: drop
remedy-survives: n/a
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
```

Drop ground for J10: optional by the reviewer's own framing, with no measured defect behind any of the three; the `child_next` extraction and the `</dev/null` hardening are adds, which §1.11 subtraction rule (a) places last, and the stdin note's own premise measured safe. The `is_num`-is-dead half is **subsumed by J4** — that fix revives it, so it needs no separate action.

```
finding: J11 — [code-reviewer record-shape nit] the PR body's "last 100 merged PRs / five inversions" corpus figure is typed narration rather than a rendered block or pointer (§1.11 L2)
site: PR #722 body#Plan — Input — merge-ordered
axis-key: pr-722-body#plan-input-merge-ordered::record-shape
verdict: confirmed
evidence: executed=gh pr view 675/677 --json number,createdAt,mergedAt mode=CI matches-artifact-mode=yes (the artifact is the PR body; its mode of verification is the live API it cites) result=#675 created 2026-08-17 merged 2026-08-21; #677 created 2026-08-18 merged 2026-08-19 — a genuine createdAt/mergedAt inversion, so the cited pointer resolves and the substance is true. That REFUTES the finding's L2 characterization in part: the sentence names five concrete PR pairs, which is L2 (ii)'s pointer form, not bare narration. What survives is narrower: the corpus scope "measured on this repo's last 100 merged PRs" carries no pinned command+output, which §1.10(b) requires of a corpus claim.
action: drop
remedy-survives: n/a
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
```

Drop ground for J11: §1.11 L1 makes the PR body a **record** whose truth-obligation is authoring-time, and the claim is measured true here; the reviewer's own disposition is conditional ("if it is ever touched"); and L5's closing rule is that a fix round may remove claims and fix behavior but cannot be obligated into minting new prose — re-rendering a true record claim is exactly that. Recorded so a later toucher applies deletion-or-pointer, never a refresh.

---

## Summary for the author

**fix-now (6), all in one commit:** J1 (header states the child-stderr forwarding — modify the anomaly paragraph, touch no behavior), J2 (delete the two unboundedness clauses at **both** sites: header 50-52 and inline comment 75-76; keep the quantified bound), J3 (generalize the exit-1 line; do **not** enumerate the three preflights), J4 (`is_num "$apr" || usage` on the explicit-list route), J6 (`m=$((10#$m))` after `is_num`), J7 (**remedy replaced** — delete the exactness claim at four sites rather than add a notice channel; the author owns the wording).

Five of the six are header/prose subtractions on one file plus two one-liners. No smoke arm is put at risk: nothing greps the deleted phrases, §185j pins only the usage line's prefix and the literal `--recent`, and §185d's anomaly count is untouched because J1 changes no behavior.

**defer-to-issue (1):** J5 → a **new** carrier for `scripts/ghjig_rounds.sh` residuals (not #715, not #718 — both are per-script carriers for other scripts). Three items belong together in it: J5 itself, J1's rejected branch (b) (prefix forwarded child stderr with `pr=<N>`), and J6's L4 question (a shared base-10 normalizer both scripts invoke, which needs `scripts/ghjig_judged_list.sh` edited and is therefore barred by #721's Out of scope).

**drop (4):** J8, J9, J10, J11 — each confirmed as described and each ruled not a defect on a stated ground.

**Honest note on the ledger.** No finding was refuted **in whole**; both reviewers ran reproductions and all of them reproduced at the head. Three sub-claims were refuted, and two of them changed a ruling: (a) J6's severity rationale — `--recent 010` does *not* fail at jq, it silently sends `--limit 32` where the header promises 40, which strengthens the finding; (b) J7's "silently" — the shortfall is readable from the trend token count, which is why `remedy-survives: no`; (c) J11's L2 characterization — the sentence carries five resolvable pointers. Also recorded: one of my own measurements was taken in the wrong mode (zsh arithmetic gave 40 for `010*4`), and it was discarded and re-taken in bash (32) rather than allowed to resolve J6.

**Not a vote.** No `ship`/`block`/`approve` token here; `code-reviewer`'s `ship after fix` and `security-reviewer`'s `approve` stand unaltered.

**Files:** the judged artifact is `/Users/nhn/ilgyu/ghjig-claude/scripts/ghjig_rounds.sh` (at e7dbbe3), its test face `/Users/nhn/ilgyu/ghjig-claude/scripts/test/smoke.d/73-rounds.sh`, the child code home `/Users/nhn/ilgyu/ghjig-claude/scripts/ghjig_judged_list.sh` (lines 87-90 for the octal kinship), and the doctrine at `/Users/nhn/ilgyu/ghjig-claude/SPEC.md` §1.11 (lines 612-630). Reproduction harness: `/private/tmp/claude-501/-Users-nhn-ilgyu-ghjig-claude/78745e01-c216-413d-8387-171f67de7666/scratchpad/judge722/` (`run.sh`, `run2.sh`, `run5.sh`–`run8.sh`, `bin/gh`, `fx2/`).

**Posting.** There is a PR, so `durable:` is not declared here — hand this reply verbatim to `scripts/ghjig_judged_list.sh post 722 <file>` before the first fix is written; the script derives the round and renders the header and marker itself. This reply intentionally contains no concrete marker.

ilgyu-yi:
## Finding triage (round 2)

reviewed-head: 1b2929e5f4a420e65c86036cdb19188614ce3683

**Artifact provenance.** Head resolved independently (`gh pr view 722 --json headRefOid`). Both blobs read at that sha (`git show <sha>:scripts/…`) and extracted into scratch; the prior head e7dbbe3 extracted alongside for F3's byte-identity check. Every reproduction ran as a **script file** behind a PATH `gh` stub, with cwd and all writes inside scratch, against a scratch `SHELL_ROOT` holding the head blobs — so the two mutations never touched the ambient tree. The §185 section file is the head blob, sourced by a scratch runner supplying only `ok`/`ng`/`TMP`/`SHELL_ROOT`.

**prior-round resolution.** `scripts/ghjig_judged_list.sh rounds 722` → `round=1 head=e7dbbe39a12772dea0a9ab5e28f1825812f2c922`, `next=2`, no `anomaly:` lines, no duplicate refusal. Body read via `show 722 1`. So `prior-round: 1@​e7dbbe3` on every record; check 5 and check 6 are **live**, and `swing: none` is legal where earned.

---

```
finding: F1 — the surviving rationale clause at :53 ("wide enough that the window is not a listing prefix") is false for M ≥ 200
site: scripts/ghjig_rounds.sh#resolve_recent
axis-key: scripts/ghjig_rounds.sh#resolve_recent::header-claim-false
verdict: confirmed
evidence: executed=bash <scratch>/inst/ghjig_rounds.sh --recent <M> for M in 1 5 6 49 50 51 100 150 190 198 199 200 201 500 1000 100000, reading the --limit the gh stub received mode=script-file matches-artifact-mode=yes result=limit= 20,20,24,196,200,200,200,200,200,200,200,200,201,500,1000,100000. At M ≥ 200 limit == M EXACTLY, so `.[0:$m]` is a no-op and the fetched page IS the top-M createdAt-descending prefix — membership is exactly the listing prefix, only ordering is still corrected. The reviewer's band arithmetic reproduces, and the measurement is SHARPER than the finding: the margin decays continuously (150 at M=50, 100 at M=100, 10 at M=190, 2 at M=198, 1 at M=199, 0 from M=200), so the claim is already near-vacuous well inside the band. Second site carries the same claim as a design REQUIREMENT the code does not meet above the cap: the inline comment at :76-79 ("the fetch must reach PAST the M topmost createdAt-descending entries or the window could only ever be a listing prefix").
action: fix-now
remedy-survives: yes
axis: fewer↔more
target-position: low
justification: the reviewer's disjunction — branch (a) delete the clause survives; branch (b) "bound it (for M below the cap)" does NOT survive, killed by the measurement above and not by any rule about direction: there is no band boundary that makes the claim substantively true, since the over-fetch margin is 1 entry at M=199, so a "below the cap" qualifier would restate a near-vacuous claim and mint a new surface to guard (§1.11 L5 puts deletion first). Site set is BOTH occurrences, per round 1's J2 precedent in this same file (one claim, two sites): :53 and the false limb of :76-79. Both endpoints stated — low end: delete exactly those two clauses. High end (the floor the deletion must not cross): retain the quantified bound "4×M, floor 20, cap 200 (never below M)", which this probe re-measures as accurate; retain the verbatim resolution command at :56 (#721 AC3 depends on it); retain the merge-ORDER rationale at :46-51, which is true for all M and is what §185g/§185h lock. Recurrence (check 6) EVALUATED and does NOT arm, on the scope test: §1.10(c)'s claim class is a claim that EXPIRES (live count, census, self-referential measurement, review archaeology). This clause does not expire — its truth is fixed by code in the same file, and it was false at authoring, born-false rather than rotted. Not a re-litigation of round 1's retained floor either: J2's floor was the quantified bound, which this deletion leaves intact. Fix-now rather than deferred to #724 on two grounds: the header is the living set and the declared contract home, where round 1 ruled four header-claim falsehoods fix-now on this same file; and #724's Out of scope bars "re-opening the --recent window's merge-ordering or over-fetch design", so a defer of a prose-only deletion would land there mis-framed.
prior-round: 1@​e7dbbe3
swing: none
```

```
finding: F2 — both round-1 behavioral fixes ship with no negative face, and the header's negative-face claim is therefore unbacked for them
site: scripts/test/smoke.d/73-rounds.sh#§185
axis-key: scripts/test/smoke.d/73-rounds.sh#§185::negative-face-unbacked
verdict: confirmed
evidence: executed=bash <scratch>/run185.sh (the head's §185 section sourced against a scratch SHELL_ROOT) three times mode=script-file matches-artifact-mode=yes result=baseline 10 pass / 0 fail; mutant A (`sed '189,191d'` — argv guard deleted) 10 pass / 0 fail; mutant B (`m=$((10#$1))` → `m=$(($1))`) 10 pass / 0 fail. Both mutations survive, as the reviewer measured. Mutant A's live output is worse than stated: `bash mutA 42 notanum` → `pr=42 rounds=0` / `pr=notanum rounds=0` / `trend: 0 0`, rc 0 — a FABRICATED PR reported with a rounds figure on the exit-0 success path, i.e. round 1's J4 forged-report class re-armed exactly. Mutant B: `--recent 08` → rc 1 with the raw `line 156: 08: value too great for base` diagnostic, i.e. round 1's J6 re-armed. The reviewer's two proposed arms both DISCRIMINATE, verified: head `42 notanum` → rc 1, stdout empty, usage on stderr (vs mutant A rc 0 + two report lines); head `--recent 08` → limit=32, rc 0 (vs mutant B rc 1, no --limit ever sent).
action: fix-now
remedy-survives: yes
axis: fewer↔more
target-position: interior
justification: the obligation is NOT #721 AC4 — read at the issue, AC4 constrains the KIND of arm ("normal-test locks only… no prose-grepping arm") and neither requires nor forbids an arm per behavior, so the reviewer's own caveat is exact. It is §1.10(d): the header at :6-7 is a machine-pair claim ("the smoke suite's §185 behavior arms are its negative face"), authored in the living set, and §1.10(d) requires such a claim to ship one attempted violation and its red — two attempted violations drew GREEN. The alternative remedy of narrowing the header sentence instead (subtraction rule (a) would normally place modify before add) is rejected on doctrine, not preference: naming which behaviors this round left unlocked is review archaeology in a durable artifact, which §1.10(c) forbids — so there is nothing here to remove or modify, which is precisely when rule (a) admits an add. The recorded-gap route the body's Decisions already used once (the unresolvable-window path) does not extend here on a §6.0 P1 cost-asymmetry: these two behaviors are FIXES for defects a review round already caught with executed evidence, and a caught defect that can silently re-arm costs a whole further review cycle, against one test-author dispatch on the existing `s185r_run` harness. Both endpoints stated — floor 2 arms: fewer leaves one of the two mutations surviving, measured above. Ceiling 2 arms: more would lock paths this round did not add (the unresolvable-window exit-1 branch the body's Decisions names as a standing recorded gap), which is #724's territory, not this round's. Scoping note on the surviving remedy, measured: arm 1's argv should be `42 <non-numeric>`, not the bad argument alone — head's stdout is then EMPTY, which also pins the validate-before-act ordering the code comment at :186-188 claims ("it must not follow a partially emitted report"); the bad argument alone would not exercise it.
prior-round: 1@​e7dbbe3
swing: none
```

```
finding: F3 — a raw bash diagnostic leaks past the die() reason contract on an out-of-range M
site: scripts/ghjig_rounds.sh#--recent-case-arm
axis-key: scripts/ghjig_rounds.sh#--recent-case-arm::unsanitized-numeric-operand
verdict: confirmed
evidence: executed=bash <scratch>/inst/ghjig_rounds.sh --recent 99999999999999999999 and the same argv against the e7dbbe3 blob mode=script-file matches-artifact-mode=yes result=head → stderr `…/ghjig_rounds.sh: line 175: [: 99999999999999999999: integer expression expected` then `ghjig_rounds: reject: <M> must be at least 1 (got: …)`, rc 1, stdout empty — posture correct, reason contract leaked. Prior head identical at its line 166, and the statement is byte-identical at both: `    [ "$2" -ge 1 ] || die "reject: <M> must be at least 1 (got: $2)"` (verified by sed at both blobs). Pre-existing, not a delta regression; the reviewer's routing is exact.
action: defer-to-issue
remedy-survives: n/a
axis: n/a
target-position: n/a
justification: n/a
prior-round: 1@​e7dbbe3
swing: n/a
```

Deferral target for F3 — **#724**, the per-script carrier for `scripts/ghjig_rounds.sh` (read at the issue: open, three items, and its Notes already anchor this exact seam). Grounds for deferral rather than fix-now: it is pre-existing and byte-identical at the prior head, so it is not this delta's defect and #721's AC set does not reach it; the posture is already correct (rc 1, empty stdout); and it is the same disposition class as #724's item 3 — what a rejection looks like when an operand reaches a shell numeric context before its predicate. It needs a fourth item and its own AC there; how that lands on an already-reviewed issue body is the caller's call.

```
finding: F4 — header line 70 is 87 columns, breaking the header block's wrap
site: scripts/ghjig_rounds.sh#header-exit-codes
axis-key: scripts/ghjig_rounds.sh#header-exit-codes::line-width
verdict: confirmed
evidence: executed=python3 codepoint+byte scan over header lines 1-71 of the head blob mode=script-file matches-artifact-mode=yes result=only two lines exceed 80 CODEPOINTS — L16 at 81 and L70 at 87; max over the whole header is (87, line 70). The four 81-BYTE neighbours L39/L43/L52/L62 are 79 codepoints (they carry `−`/`×`/`→`), exactly as the finding describes, so the naive byte scan that flags them is the wrong measurement and L70 is the unique real outlier. Introduced by the delta (the diff's `+ A per-PR derivation failure is never one of these. Nothing here fails closed: this is`).
action: fix-now
remedy-survives: yes
axis: none — discrete (the block's wrap width is already fixed at 80 by every surrounding line; the item is conformance of one line to an existing position, not the choice of a position)
target-position: n/a
justification: n/a — born observe-only per §6.0 P3, recorded not gated. One rewrap, inside the same header paragraph F1's and the exit-code edits already touch, so it costs nothing to carry in this commit and nothing is lost by dropping it.
prior-round: 1@​e7dbbe3
swing: n/a
```

```
finding: F5 — the AC5 evidence pins a superseded head, and one Plan sentence contradicts the shipped contract home
site: pr-722-body#test-plan-ac5-evidence
axis-key: pr-722-body#test-plan-ac5-evidence::pin-superseded
verdict: confirmed
evidence: executed=gh pr view 722 --json body, then grep for the pin and the Plan sentence mode=CI matches-artifact-mode=yes (the artifact is the PR body; its verification mode is the API it is fetched from) result=(i) the `## Test plan` rendered block reads `pin: e7dbbe39a12772dea0a9ab5e28f1825812f2c922 2026-08-26Z` with `smoke: pass=1407 fail=0` — a correctly-pinned §1.10(b) record of a run at the PREVIOUS head, so #721's AC5 is unattested for the shipped artifact 1b2929e. (ii) `## Plan` → "Output grammar" reads "the sweep continues and the instrument exits 0 (only a usage error exits non-zero)", contradicting head :66-69, which names four exit-1 causes — the correction this very delta made. (iii) `## Decisions` holds five entries, the newest dated 2026-08-26 for the round-1 Code phase; none covers the round-2 fix commit. (iv) REFUTED IN PART: the `## Ship gate` boxes are indeed all unchecked, but that is the correct state for a review round still in flight — those boxes ("code-reviewer passed", "CI green", "PR body curated") are the /ship step's own checklist, and checking them now would be the attestation-before-the-fact this repo treats as forged evidence. That sub-claim is excluded from the fix list.
action: fix-now
remedy-survives: yes
axis: none — discrete (each surviving item is a presence/absence question about a record surface — a pin either resolves at the artifact or it does not, a false parenthetical is present or deleted, a Decisions entry exists or does not; there is no continuum to occupy)
target-position: n/a
justification: n/a — born observe-only per §6.0 P3. Three surviving items, each with its remedy and its ground. (i) Re-render the AC5 block via `scripts/ghjig_evidence.sh` (confirmed executable at the head) — this is a NEW PIN, which §1.10(c) sanctions ("removal or pinning, never a refresh"), not a refresh of a live figure. Ordering constraint, or the author walks straight into the refresh loop: render ONCE, after this round's fix commit, against the FINAL pre-merge head — a re-render at 1b2929e would be superseded by the very commit that performs it. (ii) Delete the parenthetical "(only a usage error exits non-zero)" per §1.11 L5, do not restate the class. The §1.11 L1 objection that records age rather than being repaired does NOT shield it: this sentence was false at AUTHORING (round 1's J3 confirmed the same falsehood by executing the three preflight paths at e7dbbe3), and the body's birth review is still running, so it is admissible on both of L1's limbs. Second axis key, recorded so a future round can match on it: `pr-722-body#plan-output-grammar::claim-false`. (iii) One Decisions entry covering the round-2 corrections, per the repo's commit → PR-body-update order. Recurrence (check 6) evaluated and does not arm on any of the three: round 1's only PR-body record was J11 at `pr-722-body#plan-input-merge-ordered::record-shape` (confirmed, dropped), a different site and kind, and the reviewer did not re-raise it — correctly, and it stays not re-litigated.
prior-round: 1@​e7dbbe3
swing: n/a
```

---

## Summary for the author

**fix-now (4).** F1 (delete the false clause at **both** sites — header :53 and the false limb of the inline comment :76-79; keep the quantified bound, the verbatim resolution command, and the merge-order rationale); F2 (**two §185 arms**, the only item needing the test author: `42 <non-numeric>` → rc 1 with empty stdout, and `--recent 08` → the resolution's `--limit` is 32); F4 (rewrap :70 to ≤80); F5 (re-render the AC5 evidence block **after** the fix commit at the final head, delete the "only a usage error exits non-zero" parenthetical, add one Decisions entry — the Ship-gate boxes are **not** part of this).

Three of the four are edits inside one header paragraph plus the PR body. F2 is the only new work, and it is two invocations on the existing `s185r_run` harness with fixtures that already exist. No arm is put at risk: nothing greps the deleted clauses, §185g/h assert numeric behavior rather than the prose, and §185h's `limit > M` assertion at M=2 stays true.

**defer-to-issue (1).** F3 → **#724** as a fourth item with its own AC.

**Refuted in part (retained, excluded from the fix list).** F5's Ship-gate sub-claim (unchecked boxes are correct mid-round). F1's alternative remedy branch (b) and F2's header-narrowing alternative are both killed — by measurement and by §1.10(c) respectively — and are recorded so neither is re-proposed next round.

**Direct answer to the F1 question posed.** F1 is **not** a recurrence under check 6 and **not** a re-litigation of round 1's retained floor. Check 6's scope is claims that *expire*; this clause's truth is fixed by code in the same file and it was false at authoring. J2's retained floor was the quantified bound, which F1's deletion leaves standing. What F1 *is*: a genuinely new band, and the anti-swing check passes on the same end of the same axis (deletion again, `low` again) rather than reversing — which is the shape a converging round should have.

**Two observations, not judged records** (I add no findings): header **line 16 is 81 codepoints**, also introduced by the delta, same class as F4 but outside the finding set; and **§185h's own comment at :394-395** carries the same over-broad generalization F1 deletes ("so the window cannot be a prefix of the listing") — its *assertion* (`limit > M` at M=2) is sound, only the comment over-reaches, and it is in a different file.

**Verdict untouched.** `code-reviewer`'s `ship after fix` stands exactly as written; I refuted nothing in whole and cast no vote.

**Honest ledger.** All five findings reproduced. Two measurements came out *stronger* than the reviewer stated — F1's margin decays continuously to 1 entry at M=199 (which is what kills the "bound it" branch), and F2's mutant A prints a fabricated `pr=notanum rounds=0` on the exit-0 success path. One sub-claim was refuted (F5's Ship-gate boxes). One measurement discipline note: my first width scan used `awk`, which counted **bytes** and falsely flagged four 79-codepoint neighbours; it was discarded and re-taken codepoint-aware before F4 was ruled.

**Posting.** There is a PR, so no `durable:` declaration: hand this reply verbatim to `scripts/ghjig_judged_list.sh post 722 <file>` **before the first fix is written**; the script derives the round and renders the header and marker itself. This reply intentionally contains no concrete marker.

**Files.** Judged artifact `/Users/nhn/ilgyu/ghjig-claude/scripts/ghjig_rounds.sh` and its test face `/Users/nhn/ilgyu/ghjig-claude/scripts/test/smoke.d/73-rounds.sh` (both read at 1b2929e, not from the worktree); doctrine at `/Users/nhn/ilgyu/ghjig-claude/SPEC.md` §1.10(c)/(d) and §1.11; findings input `/private/tmp/claude-501/-Users-nhn-ilgyu-ghjig-claude/78745e01-c216-413d-8387-171f67de7666/scratchpad/722-findings-r2.md`. Reproduction harness (all writes confined here): `/private/tmp/claude-501/-Users-nhn-ilgyu-ghjig-claude/78745e01-c216-413d-8387-171f67de7666/scratchpad/judge722r2/` — `band.sh`, `run185.sh`, `arms.sh`, `bin/gh`, `inst/`, `root/scripts/`.

ilgyu-yi (commented):
## Review verdict: approve

`code-reviewer` returned `ship` and `security-reviewer` returned `ship`; both resolved the head independently and both reported `5c9306af7c35a72fe890c5c8fcbad0875473a061`, matching the caller's privately-held head.

**Code review.** No finding at fix-now or block severity. Every real item it found was already confirmed and dispositioned in round 1 or round 2, and the two whose grounds rested on a measurement — round 1's J10 stdin premise and the over-fetch arithmetic behind round 1's J2 / round 2's F1 — were re-measured at this head and hold. It verified by execution that the `next − 1` derivation reproduces on live data (`--recent 5` printed `trend: 3 2 1 2 1`, with #706's non-contiguous history raising the anomaly note the header documents), that the merge-ordering correction is load-bearing against today's real listing, that the fail-open route reports a failing PR and continues, that stdout is byte-empty on a usage rejection, and that the SPEC §8 count bump is test-forced rather than asserted. All five round-2 fix-now items are present at this head, including the evidence pin now naming this exact sha.

**Security review.** No High, no Medium; no exploitable defect at this head. It confirmed by execution that there is no command construction (hostile `number` strings carrying `;` and `$( )` reached argv intact and executed nothing), that untrusted PR-comment bytes cannot reach the parent's anchored parsers, that the two `set -euo pipefail` traps on this path are handled, that window resolution fails closed while per-PR failures fail open as designed, and that the artifact gates nothing — no hook, no CI, no allow-list, no state write. Its three Low items and two Info notes are defense-in-depth or latent-not-live by its own measurement, and two of them are already carried by issue #724; the remaining ones are routed there rather than re-opening this head.

Approve, pinned to the head above.

PR #727, opened 2026-08-26
fix(#694): bound each 174h leg by its own measurement, not one guess

### LINKED ISSUE #694: smoke §174h's 20s wall-clock bound reds on a developer machine while CI is green
## What

Smoke `§174h` (in `scripts/test/smoke.d/70-gates-contentlocks.sh`, added by PR #677) times `scripts/lint_citations.sh` along five cost axes against a strict under-20s bound per leg. On a developer machine the 250-span leg measures exactly 20 seconds, failing the `-lt 20` comparison — so the full local suite reports `fail=1` on a healthy tree. Measured 2026-08-22 (arm64 macOS, during PR #693's ship): three separate full-suite runs, each `pass=1332 fail=1` with `§174h` the only red (`spancount=20s/200`); a detached worktree at `origin/main` on the same machine reds identically (`pass=1331 fail=1`), and `lint_citations.sh` is the same blob at both ends (`git rev-parse origin/main:scripts/lint_citations.sh` = `b51091807a…`). CI (`ubuntu-latest`, `macos-latest`) is green on the same commits.

The bound's own comment says it is "deliberately loose (20 s) — they separate linear from quadratic by more than an order of magnitude without pinning machine speed"; the 250-span leg no longer has that margin on real hardware.

## Why

A permanently-red arm on developer machines makes `fail=1` the expected local outcome, which trains authors to stop reading the failure line — the alarm-fatigue path by which a genuinely new red gets ignored. It also breaks clean-window claims structurally: PR #693's AC4 ("full suite passes with no live agent worktree") could only be discharged via CI, with a per-PR justification comment explaining the local red. MISSION "The mechanism": a signal that fires on healthy state is noise the shell then has to argue around instead of evidence it can lean on.

## Acceptance criteria

- [ ] AC1 — **the green half**: a full local suite run on a machine where the 250-span leg costs ~20s is green at `§174h`, with that run's `ok` line recorded in a comment on this Issue.
- [ ] AC2 — **the red half, per leg, recorded**. `§174h` asserts five legs; for **each** leg a mutation that defeats **that leg's** guard is applied and the arm's verdict under it is recorded in a comment on this Issue, naming the mutation applied. A leg whose verdict is that the arm **stays green** is a result to record, not a gap in the survey — that leg's guard is then either repaired or retired with the reason stated. The mutation is applied **to the guard or to the code path the leg measures — never to that leg's fixture**: shrinking a probe until an assertion trips shows only that the assertion reads its input, not that it catches the regression the leg exists for. Beyond that the choice is the implementer's; what is fixed is that every leg is covered and every verdict written down. The arm's own probes do not discharge this: they are green by construction.
- [ ] AC3 — **the separation claim is made true or deleted, per leg and two-sided.** For each of the five legs, state the healthy cost and that leg's own pre-fix cost, and show the chosen bound sits above the first by a stated margin and below the second by a stated margin. **The pre-fix costs are measured by this change, not quoted from the arm's comment**: of the three figures the comment records, only two are that leg's — the one-line 4 MB leg's 66.9 s and the 1.6 MB body leg's 118 s. The 250-span leg's "two minutes" is the comment's figure for a 64 KB body, roughly three times this leg's probe, and the many-line and 30-span legs have no recorded pre-fix cost at all. A margin stated against a figure belonging to a different fixture is the defect this criterion exists to prevent, one order down. The above-side margin is stated as a **ratio**, not a difference, so a bound one second above the healthy cost does not satisfy it.
- [ ] AC4 — **the arm's own separation claim is corrected or removed.** The comment at `scripts/test/smoke.d/70-gates-contentlocks.sh` asserts the bounds *"separate linear from quadratic by more than an order of magnitude without pinning machine speed"*. Where AC3's two-sided window exists, the claim is restated to the separation actually measured; where it does not, the clock guard for that leg is **retired** in favour of a structural assertion and the claim is deleted rather than reworded — deletion is the default repair (§1.11 L5). Two constraints on retirement. It states **which regression class the retired clock stopped catching** (the cap assertion `s174h_clsC -eq 200` bounds the number of forks, so it survives a deleted cap but not per-span work becoming uniformly slower). And it is **unavailable at the 1.6 MB body leg**, whose only assertion is the clock — `probeB`'s output is discarded and there is no `s174h_clsB` companion — so retiring it would delete the leg rather than substitute for it; that leg's window is wide and is expected to need no retirement.
- [ ] AC5 — full smoke suite green in a window with **no live agent worktree** (#642 explains why that qualifier is load-bearing).
- [ ] AC6 — the change names its own negative face, or states that it has none and why that is acceptable (§6.0).

## Out of scope

- **Any edit to `scripts/lint_citations.sh`.** Stated by **file**, not by behaviour, so a pure performance change (fewer forks per span) that alters no cap and no output is excluded too. That script's caps and bounds are #677's shipped contract; this issue changes the smoke arm only — which binds the shipped diff, not a transient mutation harness AC2 requires. A new environment seam in that file — alongside the existing ones at `scripts/lint_citations.sh:166-169` — is excluded on the same ground.
- **The other four legs' thresholds are in scope only through AC3 and AC4**, which bind every leg.
- `§174g`'s timing clause, which **#678** owns.

## Notes

- Arm: `scripts/test/smoke.d/70-gates-contentlocks.sh`, the `§174h` block (search anchor: `every factor of the cost must be bounded`).
- Measurements pinned in PR #693's body (Test plan section) and issue #656's AC4-qualification comment.
- **Nearest adjacency is #678**, opened three days earlier — same file, same clock-based class, inverse failure direction: `§174g` reds on a host fast enough that two `$SECONDS` samples land in the same second, where this arm reds on a host slow enough to reach the bound. They are disjoint (different arm, different remedy) but must be sequenced consciously, because **#678's judged remedy for `§174g` is "a non-strict relation **plus** an absolute ceiling, **or a proxy that counts work done rather than time elapsed**"**. Only the first branch collides: an absolute wall-clock ceiling is the construct AC3 and AC4 constrain, so if #678 takes it, whichever lands second inherits the other's constraint; if #678 takes the work-counting proxy, no ceiling exists and the coupling does not arise.
- Related: #677 (authored the arm), #642 (the other known local-window hazard, `§148i-cli` under a live worktree), #695 (the same file, an arm whose `ok` message overclaims its assertions).

ilgyu-yi:
Closes #694

## Goal

Smoke `§174h` redded the full local suite on a healthy tree. The cause was not a slow machine: the arm applied **one 20s bound to five legs whose healthy costs span two orders of magnitude**, so a single number could not be both loose enough to pass a healthy host and tight enough to catch the regression each leg exists for. It was too loose for four legs and too tight for one.

`MISSION.md` "The mechanism": an arm that fires on a healthy state is noise the shell then argues around instead of evidence it can lean on. It also broke clean-window claims structurally — a PR wanting to assert "full suite green" had to explain a red first.

## What changed

**Each leg's bound now comes from that leg's own measured pair** — at least 4× above its healthy cost and at least 4× below its pre-fix cost. The pre-fix half is **measured**, against the pre-bounding reader blob or with that leg's own cap lifted; it is not quoted from the arm's prose. The measured pairs and the resulting windows are recorded on the issue.

**The 250-span leg gets no clock bound at all.** Its healthy and pre-fix costs sit 1.25× apart — inside the ±1s quantisation of the `$SECONDS` sampling the arm uses — so the two-sided window is arithmetically empty. No bound separates healthy from broken there, which is why 20s redded healthy hosts. Its real guard was never the clock but `s174h_clsC -eq 200`, the assertion that the cap truncated 250 spans to 200; that is machine-independent and is verified still failing when the cap is lifted.

**What the retirement gives up is stated, not hidden**: the class where per-span work becomes uniformly slower while the cap still truncates — the cap bounds the *number* of forks, not the cost of each. Written at the arm per §1.11 L3 rather than papered over.

**The arm's own false claim is deleted, not reworded.** It asserted its bounds "separate linear from quadratic by more than an order of magnitude"; at the 250-span leg the separation is 1.25×. Deletion is the default repair (§1.11 L5).

## Target base

`main`

## Key context

- `scripts/test/smoke.d/70-gates-contentlocks.sh` — the `§174h` block
- `scripts/lint_citations.sh` — its caps are documented "Overridable for measurement", which is how the pre-fix half was measured without editing the file (out of scope by file)

## Checklist

- [x] **Fix**: per-leg bounds from measured pairs; the 250-span clock guard retired in favour of its cap assertion; the false separation claim deleted; ok/ng messages restated to what the assertions actually check
- [x] **Fix**: `changelog_unreleased/fixed/694.md`
- [~] **Doc**: N/A — the arm's own comment block is the contract home for its bounds and is updated in place; no SPEC or `docs/*.md` surface states them (`grep` for the arm's bound returns only this file).
- [~] **Test**: N/A — the arm *is* the test. AC2's obligation is a mutation survey rather than a new test, and it is recorded on the issue.

## Test plan

- [x] Full suite `pass=1407 fail=0` locally — the first clean local run since this arm regressed.
- [x] **Per-leg mutation survey, all five red**, each mutation on the leg's guard or measured code path and never on its fixture. Recorded on #694 with the applied mutation and the observed result per leg. The 250-span row is the one that matters: retiring its clock left its cap assertion still failing when the cap is lifted.

## Notes for the reviewer

- The `66.9 s` figure near the top of the arm's comment is **left in place deliberately**. It is #677's record of the doubling curve that motivated the one-line leg, not a figure any bound here rests on. My own measurement of that leg's pre-fix cost is 26.7s and is what its bound is derived from.
- A first draft of this change took that leg's pre-fix cost from the prose instead of measuring it, and the bound that produced (15s) sits outside the window the measured figure gives. It is 5s. Recorded because the acceptance criterion requiring measurement is what caught it.
## Review round 1 — `ship after fix`, applied

Every measured figure in this change was reproduced independently by the reviewer on its own host and matched. Four fixes, three of which were typed claims that were wrong.

1. **The retirement's stated reason was false.** The comment attributed the empty window to `$SECONDS`' ±1 s quantisation; the gap is four to five seconds — four to five times the quantisation — and a ratio is not comparable to a second. The absolute that followed ("no bound separates healthy from pre-fix") is false too: a bound of 22 separates them on this host. Replaced with the argument that holds — a bound can be placed inside a 1.25× window but not *safely*, since no value clears the 4× floor on both sides and any value inside it pins machine speed to within 25%. **The retirement itself was upheld**: lifting the cap makes the classified count read 250 instead of 200, so the remaining guard is live.
2. **Two pre-fix figures for the same fixture collided.** `66.9 s` and `26.7 s` sat twelve lines apart on the same axis with nothing reconciling them. `66.9 s` is now marked as #677's figure on its own host, with this change's own measurement named as what the 5 s bound derives from. Left in place rather than deleted, since it is that PR's record and no bound rests on it.
3. **"Three orders of magnitude" was wrong** for the healthy costs (0.20–19.91 s ≈ two orders). True only of the healthy-to-pre-fix figures, and the comment now says which.
4. **AC1 and AC6 were undischarged** — the green `ok` line and the §6.0 negative-face statement are now recorded on #694.

The reviewer also confirmed the four surviving bounds are safe on CI by a stronger argument than headroom: CI is green today at the old `-lt 20` while this host measures 20 s on that leg, so CI is not slower than this host on the fork-heavy path.

**Worth stating plainly, because it is this PR's own subject.** The defect being repaired here is a false claim in a test's comment. The first repair deleted it and wrote a different false claim in its place. That is the same defect one turn later, and it is why the fix list above is mostly about figures rather than about code.

### DIFF
--- a/changelog_unreleased/fixed/694.md
+++ b/changelog_unreleased/fixed/694.md
@@ -1,0 +1,1 @@
+- Smoke §174h no longer reds on a healthy developer machine: each of its five cost legs now carries a bound derived from that leg's own measured healthy and pre-fix costs, and the 250-span leg — whose two costs are too close for any bound to separate them — is guarded by its cap assertion instead of by the clock (#694).
--- a/scripts/test/smoke.d/70-gates-contentlocks.sh
+++ b/scripts/test/smoke.d/70-gates-contentlocks.sh
@@ -4330,7 +4330,10 @@ if [ -n "$S174G_DIR" ] && [ -d "$S174G_DIR" ]; then rm -rf "$S174G_DIR"; fi
 #   - 4 MB as 60 000 short lines, one span — the file-size axis.
 #   - the SAME 4 MB as ONE line: the per-record normalisation is quadratic in line
 #     length, so a one-line file never entered the trim loop (0.32 s at 256 KB
-#     doubling cleanly to 66.9 s at 4 MB). Reachable with nothing planted — a minified
+#     doubling cleanly to 66.9 s at 4 MB — #677's figure on its own host; this
+#     arm's own measurement of the same axis at the pre-fix blob fca2e4b is
+#     26.7 s, and that is the figure the 5 s bound below derives from).
+#     Reachable with nothing planted — a minified
 #     bundle, a one-line JSON dump, an inlined SVG.
 #   - 30 spans against the same file — the multiplier a one-span arm cannot see.
 #   - a 1.6 MB BODY on one line: the extractor's own scan is quadratic in one BODY
@@ -4343,14 +4346,16 @@ if [ -n "$S174G_DIR" ] && [ -d "$S174G_DIR" ]; then rm -rf "$S174G_DIR"; fi
 # d376190 on arm64 macOS, the pre-fix half against the pre-bounding reader blob
 # (fca2e4b) or with that leg's own cap lifted. many-line 0.56 s / >120 s -> 15;
 # one-line 0.29 s / 26.7 s -> 5; 30-span 5.63 s / >400 s -> 60; body-line
-# 0.20 s / 28.3 s -> 5. A single
-# uniform bound cannot be stated: the figures differ by three orders of magnitude
-# across the legs, so one number is either loose enough to miss a regression or
-# tight enough to red a healthy host.
+# 0.20 s / 28.3 s -> 5. A single uniform bound cannot be stated: the healthy-to-
+# pre-fix figures differ by three orders of magnitude across the legs, so one
+# number is either loose enough to miss a regression or tight enough to red a
+# healthy host.
 # The 250-span leg carries NO clock bound. Its measured pair is 19.9 s / 24.9 s -
-# a 1.25x window, inside $SECONDS' own +/-1 s quantisation - so no bound separates
-# healthy from pre-fix there, and the 20 s bound it used to carry redded healthy
-# hosts (#694). Its guard is `s174h_clsC -eq 200`, which asserts the cap actually
+# a 1.25x window. A bound CAN be placed inside a gap that narrow on one host, but
+# it cannot be placed safely: no value clears the 4x floor both sides, and any
+# value inside a 1.25x window pins machine speed to within 25%. The 20 s bound it
+# used to carry is the proof - this host measures 19.9 s and redded on a healthy
+# tree (#694). Its guard is `s174h_clsC -eq 200`, which asserts the cap actually
 # truncated 250 spans to 200: that is the mechanism bounding the cost, and it is
 # machine-independent. What retiring the clock gives up is the class where per-span
 # work becomes uniformly slower while the cap still truncates - the cap bounds the
--- a/scripts/test/smoke.d/70-gates-contentlocks.sh
+++ b/scripts/test/smoke.d/70-gates-contentlocks.sh
@@ -4338,8 +4338,23 @@ if [ -n "$S174G_DIR" ] && [ -d "$S174G_DIR" ]; then rm -rf "$S174G_DIR"; fi
 #     the cited artifact's lines were capped (measured: 118 s in extraction alone).
 #   - 250 spans against the span-count cap: rungs 1 and 3 fork a git process per span,
 #     so a GitHub-legal 64 KB body of minimal spans ran for two minutes.
-# The bounds are deliberately loose (20 s) — they separate linear from quadratic by
-# more than an order of magnitude without pinning machine speed.
+# Each leg's bound is set from ITS OWN measured pair (#694): at least 4x above that
+# leg's healthy cost and at least 4x below its pre-fix cost, both measured at
+# d376190 on arm64 macOS, the pre-fix half against the pre-bounding reader blob
+# (fca2e4b) or with that leg's own cap lifted. many-line 0.56 s / >120 s -> 15;
+# one-line 0.29 s / 26.7 s -> 5; 30-span 5.63 s / >400 s -> 60; body-line
+# 0.20 s / 28.3 s -> 5. A single
+# uniform bound cannot be stated: the figures differ by three orders of magnitude
+# across the legs, so one number is either loose enough to miss a regression or
+# tight enough to red a healthy host.
+# The 250-span leg carries NO clock bound. Its measured pair is 19.9 s / 24.9 s -
+# a 1.25x window, inside $SECONDS' own +/-1 s quantisation - so no bound separates
+# healthy from pre-fix there, and the 20 s bound it used to carry redded healthy
+# hosts (#694). Its guard is `s174h_clsC -eq 200`, which asserts the cap actually
+# truncated 250 spans to 200: that is the mechanism bounding the cost, and it is
+# machine-independent. What retiring the clock gives up is the class where per-span
+# work becomes uniformly slower while the cap still truncates - the cap bounds the
+# NUMBER of forks, not the cost of each. Stated rather than covered (SPEC 1.11 L3).
 # §174j shares this repo, so it is nested here — but its ng path is duplicated into
 # BOTH guard arms below. Nested inside the success branch alone it emitted neither ok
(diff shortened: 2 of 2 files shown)


ilgyu-yi (commented):
## Review verdict: approve

`code-reviewer` returned `ship after fix` on the prior head, and all four fix-list items are applied at the head this review pins. No security surface is touched; the diff is one smoke section and one changelog fragment.

**What the review verified by execution.** It rebuilt the arm's five fixtures independently from the head's reader blob and reproduced every figure the change rests on — the healthy costs, the pre-fix costs from the pre-bounding blob, and the cap-lift measurement. It confirmed the reader blob at this head is byte-identical to `main`'s, so the by-file out-of-scope constraint holds. It reproduced the premise too: the 250-span leg measures exactly at the old bound on its host, so `-lt 20` failed on a healthy tree.

**The retirement was upheld and is not a weakened arm.** Lifting the span-count cap makes the retained assertion read 250 instead of 200, so that guard is live rather than tautological, and no safe clock bound exists for that leg — a window that narrow pins machine speed to within a quarter. The four surviving bounds were judged safe on CI by a stronger argument than headroom: CI is green at the old bound while the author's host measures at it, so CI is not the slower runner on the fork-heavy path.

**Three of the four fixes were false claims, and that is this change's own subject.** The defect being repaired here is a false claim in a test's comment; the first repair deleted it and wrote a different false claim in its place, attributing the empty window to `$SECONDS` quantisation when the gap is several times that. The absolute that followed it was false too — a bound does separate the two costs on one host; what it cannot do is separate them safely. Corrected to the policy argument. Two further typed figures were corrected: two different pre-fix costs for the same fixture sat twelve lines apart with nothing reconciling them, and the healthy costs span two orders of magnitude rather than three, which is true only of the healthy-to-pre-fix figures.

**Acceptance criteria discharged.** The green run's `ok` line and the §6.0 negative-face statement are recorded on the issue, alongside the per-leg mutation survey in which every leg reds under a mutation applied to its own guard or measured code path and never to its fixture.

Approve, pinned to the head above.

Issue #704, opened 2026-08-22
the audit consumer scripts harden their parsing: two-sided class= boundary, torn-line skip, diagnosable jq errors


ilgyu-yi:
## What

The reviewer-reject/audit consumer scripts share three parsing weaknesses, each measured during PR #703's contest and review:

1. **Unanchored `class=` capture** (`scripts/axis_recurrence.sh`, `scripts/promotion_candidates.sh`): `capture("class=(?<c>[a-z-]+)")` truncates out-of-vocabulary tokens — measured, `class=scope_bleed` and `class=scopeX` on one issue both capture as `scope` and emit a false pair — and matches inside `subclass=`. The verified remedy is the **two-sided token boundary** `(^| )class=(?<c>[a-z-]+)( |$)` with the `test()` prefilter aligned; PR #703's finding-judge measured the one-sided `(^| )class=` form insufficient (the truncation merge survives), so the two-sided form is the contract.
2. **jq stderr silencing** — in **all three** consumers, not one: the redirect closes the jq program at `scripts/promotion_candidates.sh:52`, `scripts/narrowing_candidates.sh:53` and in `scripts/axis_recurrence.sh`. It converts a program compile error into a healthy-looking placeholder line at rc 0; the smoke pin prevents shipping a broken program, but a live-machine breakage is undiagnosable. The first two are the files item 1's fix rewrites the jq program in, which is where the risk is highest. Scope or drop the redirect.
3. **Whole-parse death on a torn line** (`scripts/promotion_candidates.sh`, `scripts/narrowing_candidates.sh`): `jq -rs` dies on the first malformed line with stderr suppressed — measured in PR #703's plan contest, a fixture with a real above-threshold pair plus one truncated line printed the no-candidates placeholder over a REAL result. `scripts/axis_recurrence.sh` already carries the per-line `fromjson? // empty` posture; port it to both siblings.

## Why

MISSION "The mechanism" — these reporters are §6.0 P3's deferred positive faces; a reporter that silently converts real signal into healthy-looking quiet (item 3) or fabricates signal from malformed input (item 1) corrupts exactly the evidence surface the mechanism leans on. Why now: all three are measured (transcripts on PR #703's judged list and plan-contest record), the false-negative class is the worst failure a visibility tool can have, and the fix set is a few lines per file with existing fixture-smoke idioms (§94/§181) to lock them.

[...]

Issue #695, opened 2026-08-22
smoke §148i-cli's ok-message states a caller census its assertions do not check


ilgyu-yi:
## What

Smoke `§148i-cli`'s green-path message states a repo-wide census its assertions do not check. The `ok` line reads:

> `148i-cli: ghjig_state_dir_cli defined exactly once (in hookrt.sh); the wrapper is its only caller and never re-defines it …`

The arm's actual checks (same block, `scripts/test/smoke.d/70-gates-contentlocks.sh`, search anchor `s148i_defcount`) are: exactly one definition; that definition in `hookrt.sh`; the wrapper file does not re-define it; the wrapper file calls it. **No assertion enumerates callers**, and the census claim is false at head: `§148i-clibranch`, in the same smoke file, sources `hookrt.sh` and calls `ghjig_state_dir_cli` directly. So every suite run prints, as a passing fact, the exact claim PR #693's round-1 F2 corrected in `hookrt.sh`'s own comment (now "only script caller on the RUNTIME path", naming the smoke test caller).

## Why

MISSION "The mechanism" — the shell's ok/ng lines are read as facts by humans and by later rounds' reviewers; a green line asserting an unchecked, false census re-seeds precisely the restatement-drift that issue #656 was filed to end, from inside the test suite that guards against it. Why now: found during PR #693's ship, where the same class was corrected in `hookrt.sh`'s own comment; each week it stands, every local and CI run re-prints the false claim. The caller set, rendered rather than counted from memory — an earlier draft of this body called the smoke file the *fourth* path, which resolves at neither the filing commit nor head:

```
$ git grep -l 'ghjig_state_dir_cli'
.claude/commands/file-review.md
.claude/hooks/hookrt.sh
.githooks/_lib.sh
SPEC.md
scripts/ghjig_file_review_post.sh
scripts/test/smoke.d/70-gates-contentlocks.sh
pin: 38c6c74b396c81fa1088a1c0f768dea306e066db 2026-08-26Z
```

## Acceptance criteria

- [ ] AC1 — every place in the `§148i-cli` block that states a caller census agrees with what the arm asserts. That is **two sites, not one**: the `ok`/`ng` messages, and the arm's own header comment, which carries the identical claim. Narrowing only the messages leaves the false census in the file and does not discharge this.
[...]

PR #731, opened 2026-08-27
fix(#725): unify the audit trail on the per-project tier

### LINKED ISSUE #725: the audit trail splits by writer context, so the consumer that aggregates reject records cannot open the file they are in
## What

`audit_log` picks its destination from the context the **writer** ran in, so one trail lands in two files and each consumer reads a different half. The split is documented and intentional; what is new is that the consumer which aggregates the reject records can no longer open the file they are in.

The mechanism is two lines. `audit_log` resolves through `ghjig_state_dir` (`.claude/hooks/hookrt.sh:161`), which yields the per-project directory only when `CLAUDE_PROJECT_DIR` is set and otherwise **empty** (`:88-92`), falling back to the legacy shared `${GHJIG_ROOT:-}/.claude/audit/audit.jsonl` (`:162`). Claude Code sets `CLAUDE_PROJECT_DIR` for hook commands and a Bash-tool subprocess often runs without it — the resolver's own docblock says both. A sibling resolver exists for exactly the second case, `ghjig_state_dir_cli` (`:108-117`), and `audit_log` does not call it.

**The posture is deliberate, and that is why this is a contract change rather than a conformance fix.** `SPEC.md:894` states the resolution order and its `otherwise **empty**, in which case callers fall back to the legacy shared` branch; `SPEC.md:904` states the intent — *"in plain-Bash context (env unset, no arg) behavior is unchanged (back-compat — the legacy shared paths remain the floor)"*; `.claude/hooks/hookrt.sh:85-87` says the same in code. Nothing here is out of conformance with §3.2.2. What has changed since §3.2.2 was written is **who writes**: the audit log's writers were hooks, and the command-emitted categories — `reviewer-reject`, `activation`, `file-review`, `issue-file` — arrived later and land on the floor by that design. The consequence below is what makes the posture wrong now, and the fix is an edit to §3.2.2 itself, Doc-first.

**The registry is the precedent argued for, not a rule already binding.** `SPEC.md:902` says of the other two-tier store: *"Writers (`inject`, `self_register`) always write per-project"*, with the legacy path a read-only floor consulted only when the per-project file is absent. That asymmetry is what this Issue proposes to extend to the audit log — and the reason it does not already transfer is mechanical: `ghjig_registry_file` takes an explicit project-dir argument (`SPEC.md:896-899`), which is what lets its writers be unconditional, while `ghjig_state_dir` has no argument form. `ghjig_state_dir_cli` (`.claude/hooks/hookrt.sh:108-117`) is the later function that fills that gap, scoped by its own docblock to one caller.

Both files are live and their populations are disjoint by category:

```
$ for f in .claude/ghjig-state/audit/audit.jsonl .claude/audit/audit.jsonl; do printf "%s: lines=%s reviewer-reject=%s first=%s last=%s\n" "$f" "$(wc -l < "$f" | tr -d " ")" "$(grep -c reviewer-reject "$f" || true)" "$(head -1 "$f" | cut -c8-17)" "$(tail -1 "$f" | cut -c8-17)"; done
.claude/ghjig-state/audit/audit.jsonl: lines=54 reviewer-reject=0 first=2026-07-02 last=2026-08-17
.claude/audit/audit.jsonl: lines=75 reviewer-reject=12 first=2026-08-15 last=2026-08-26
pin: d376190dc59d8999a07f20faa498f9db503022e8 2026-08-26Z
```

Two claims behind those figures, stated at the extent they hold. Every **command**-emitted record — `reviewer_reject_audit`, `activation`, `file-review`, `issue-file` — is on the legacy floor, because the reviewer-gated commands emit from a Bash-tool step (`.claude/commands/activate.md` step 3, `.claude/commands/file-issue.md` step 4); that direction is exceptionless in the corpus above. The converse is **not** exceptionless: the floor also carries three hook-matcher records post-divergence (`merge-review` ×2, `ac-closeout` ×1, all `helper-missing` warns), so "hook writes per-project" is a tendency, not a law, and nothing here rests on it. What the two claims are needed for is narrower and does hold: the per-project file carries **no** `reviewer-reject` record, which is what blinds the aggregating consumer, and the legacy file carries **no** `block` event, which is what AC4 rests on.

**The consumer that aggregates reject records cannot open the file they are in.** `scripts/promotion_candidates.sh` resolves its input through `resolve_audit_log` (`scripts/lib/audit_log_path.sh`), which mirrors the writer's branch including the fallback, and `.claude/hooks/session_start.sh:191`/`:199` invoke it as a hook child. Same script, same commit, same moment:

```
$ printf "as a hook (CLAUDE_PROJECT_DIR set — the SessionStart shape):\n"; CLAUDE_PROJECT_DIR="$(git rev-parse --show-toplevel)" bash scripts/promotion_candidates.sh; printf "\nas a Bash-tool step (CLAUDE_PROJECT_DIR unset — the shape every command step runs in):\n"; bash scripts/promotion_candidates.sh
as a hook (CLAUDE_PROJECT_DIR set — the SessionStart shape):
promotion candidates (reviewer-reject reason-class aggregation; LIVE-only; min_rejects=3):
  (none above threshold)

as a Bash-tool step (CLAUDE_PROJECT_DIR unset — the shape every command step runs in):
promotion candidates (reviewer-reject reason-class aggregation; LIVE-only; min_rejects=3):
  activation | evidence-insufficient | rejects=3
  activation | unverifiable-ac | rejects=3
pin: d376190dc59d8999a07f20faa498f9db503022e8 2026-08-26Z
```

Classes above the threshold are invisible to the shape the advisory runs in. The counts themselves move — this Issue's own activation rounds write reject records — so the standing claim is the disjointness above, not a census.

**The advisory's remediation line contradicts itself today.** `.claude/hooks/session_start.sh:224` tells the operator to *"review via /audit or scripts/narrowing_candidates.sh + scripts/promotion_candidates.sh + scripts/ceremony_candidates.sh + scripts/spec_drift_candidates.sh"*. Run by hand from a terminal those scripts resolve the **other** tier than the advisory that recommended them, and `.claude/commands/audit.md:6` hardcodes the legacy floor outright (*"Query `.claude/ghjig-root/.claude/audit/audit.jsonl`"*), so `/audit` cannot surface a hook-fired `block` or `escape` at all. The instrument reports a signal and hands over a procedure that cannot reproduce it.

**Onboarded targets are the worse case**, and it follows from the same fallback expression rather than from an observation this repo can make: `GHJIG_ROOT` is the shell repo while the per-project state is the target's, so a target's command-emitted records land in the **shared shell repo's** log across all targets. Not measured here — this repo is its own shell root, so both paths sit inside it.

## Why

Two MISSION items, one of them a stated success signal.

`MISSION.md:31` — *"**The escape hatch stays narrow.** Audit-log queries (`/audit`) show that bypasses (`SKIP_HOOKS=...`) cluster in the small set of legitimate cases the SPEC names"*. `SKIP_HOOKS` escapes are written by `should_skip` on the hook path, so they are in the per-project tier, and `/audit` — the instrument MISSION names for that measurement — is hardcoded to the other one. Not a degraded signal for that category; a zero one.

`MISSION.md:20` — the isolation model makes the audit log per-project **by name**: *"ephemeral state**: the audit log, caches, and scope registry under that project's own `ghjig-state/`"*. That is the ground on which the registry's write-isolation rationale legitimately generalizes here, and the item the onboarded-target case above violates directly, since a target's command-emitted records land in the shared shell repo rather than that target's `ghjig-state/`.

`MISSION.md:18` — the audit line is a block's deferred positive face, *"latent until a consumer is wired to read the aggregate, at which point the log becomes the channel through which the shell sees its own friction and improves under use."* A consumer wired to the wrong tier is the same latent record wearing a different disguise.

**Why now.** Directive #638's item 4 adds an `info`-kind record to `reviewer_audit.sh` and constrains that it *"must be inert to every existing audit consumer"*, verified *"against each reader that keys the audit log"* — a verification that cannot mean anything while a reader and the writer resolve different files. Pinned rather than asserted, because the round-1 leg of this argument was falsified by a merged ADR the Directive body did not know about: at `d376190dc59d8999a07f20faa498f9db503022e8`, `gh issue view 638 --json state,labels` returns OPEN with the `directive` label, and `gh issue list --state all --search "in:body 638"` returns no Execution Issue for the advisory-record channel. If either has changed by the time this is read, re-derive the leg rather than trusting this sentence.

**Why this is not urgent, stated so the priority is not argued from a false premise.** It is *not* blocking Directive #639: ADR 0007 (`docs/ADRs/0007-file-review-marker-reject-count-substrate.md`, Status Accepted, merged via #690) already selected substrate (i), the `/file-review` marker, and rejected the audit-trail candidates (ii)/(iii)/(iv) on the ground that `.claude/ghjig-state/` and `.claude/audit/` are git-excluded and therefore absent from a fresh clone. This fix does not change that ground, so #639's selection stands either way.

## MISSION fit

`MISSION.md:31` ("Success looks like" → the escape hatch stays narrow) and `MISSION.md:18` (the enforcement dual's deferred positive face), both quoted above.

## Acceptance criteria

- [ ] AC1 — **the writer becomes unconditional**, matching the rule `SPEC.md:902` already states for the registry: `audit_log` resolves the per-project tier whether or not `CLAUDE_PROJECT_DIR` is set. `GHJIG_STATE_DIR_OVERRIDE` remains the test seam and still wins — "every context" does not include defeating that. `ghjig_state_dir_cli` exists for that case; whether it is the mechanism is the implementer's call, but a context-conditional destination is not an acceptable outcome.
- [ ] AC2 — **readers get the read-only floor, and only that**: the legacy path is consulted when the per-project file is absent, never preferred, and `.claude/commands/audit.md`'s hardcode is replaced by the same resolution. This is the registry's posture at `SPEC.md:902` applied to the audit log.
- [ ] AC3 — **two-sided and mutation-verified**: a record written with `CLAUDE_PROJECT_DIR` unset is visible to a reader running with it set, and a reader finds the per-project tier in preference to a populated legacy file. Reverting AC1 must red the arm; the arm must not pass by comparing two branches of the same conditional to each other.
- [ ] AC4 — **`/audit` surfaces a hook-fired record** — a `block` or an `escape` — which is the class it cannot see today, and which is the class `MISSION.md:31` names. A fix that satisfies AC1 and AC2 but not this is wrong.
- [ ] AC5 — **the records already on the floor are addressed**, migrated or left with the reason stated; silent abandonment is not acceptable. `scripts/promotion_candidates.sh` then reports the same candidate set in both shapes at a pinned commit.
- [ ] AC6 — **§3.2.2 is edited, not exempted.** `SPEC.md:894`'s fallback branch and `:904`'s "the legacy shared paths remain the floor" are what AC1 obsoletes for the audit log, so the Doc phase changes them and records why the registry's writer asymmetry now extends here; the registry's own posture is unchanged. The rewrite must also say what replaces the floor's current job: under an unconditional writer it is **AC2's read-only fallback** that continues to honour #311's "existing setups keep working" constraint, which `:904` presently discharges by writing there. Every other SPEC site naming the audit-log path is **derived by a command at the head under review** — never carried from a prior enumeration — and each names the canonical tier or points at the revised §3.2.2. §16's tree annotations legitimately name the legacy path and are not swept up.
- [ ] AC7 — **the surfaces the chosen mechanism moves are settled in the Doc phase, not discovered in the Code phase.** If AC1 is met by routing `audit_log` through `ghjig_state_dir_cli`, that resolver's caller census changes, which reaches `.claude/hooks/hookrt.sh:102-104`'s own statement of its callers, SPEC §5.7.1, and smoke `§148i-cli` — whose message wording is #695's open subject. Each is updated or explicitly ruled unaffected.
- [ ] AC8 — full smoke suite green in a window with no live agent worktree, and the change names its own negative face or states why it has none (§6.0). Note `§174h` reds independently on a developer machine (#694).

## Out of scope

- **The audit record schema** — no new categories, fields, or call sites. That vocabulary is §6.1's and originates with #361; #639 item 2 may extend it and this Issue does not.
- **The promotion threshold and `scripts/promotion_candidates.sh`'s advisory-to-hook-promotion semantics.** Whether the classes this makes visible should be promoted is a decision this Issue enables and does not make.
- **#639's substrate selection**, settled by ADR 0007, and **#638 item 4's new record**.
- **Reconciling how the two files came to hold what they hold.** The defect stands on the resolution mechanism; AC5 requires a decision about the existing records, not an archaeology of their origin.
- **The `§6.5(d)` sibling readers** (`narrowing_candidates.sh`, `ceremony_candidates.sh`, `spec_drift_candidates.sh`) except through the shared `resolve_audit_log`.

## Notes

- Anchors: `.claude/hooks/hookrt.sh:161-162`, `:88-92`, `:108-117`; `scripts/lib/audit_log_path.sh`; `.claude/hooks/session_start.sh:191`, `:199`, `:224`; `.claude/commands/audit.md:6`; `SPEC.md:902`.
- **Supersedes the unfiled draft at `.claude/state/blocked-issue-drafts/audit-reader-tier.md`**, whose reader-side analysis this body carries and whose boundary it corrects. Its reading of §3.2.2 as a deliberate floor rather than a contradiction is correct and is carried above; an earlier version of this body discarded it and claimed a conformance violation instead. That draft scoped itself to readers and put *"Changing `audit_log`'s writer resolution or the two-tier model"* out of scope — but a reader-only fix cannot reach the reject trail, because making `promotion_candidates.sh` read the per-project tier is what it already does as a hook child, and the records are on the floor. The draft is removed with this filing rather than left as a second, contradictory boundary.
- Found while checking whether this session's own `audit_log` calls had landed. They had — on the floor — while the per-project file showed no record for the day.
- Related: #361 (the reason-class trail), #314 / #316 (the per-project move that created the second tier), #602 (which added `ghjig_state_dir_cli` for the Bash-tool context), #638 item 4, #695 (the `§148i-cli` arm pinning that resolver's caller census, which AC1 may widen), #704 (parsing robustness in the same consumer family, explicitly not the writer).

ilgyu-yi:
Closes #725

## Goal

Unify the audit trail: `audit_log` writes the per-project `ghjig-state/audit/audit.jsonl` unconditionally (a dedicated resolver — override seam → `CLAUDE_PROJECT_DIR` → main-worktree top level → self-location), the legacy shared file is demoted to a read-only floor, and the worktree/env-unset silent-drop shapes are closed — so the consumers that aggregate reject records can open the file the records are in.

MISSION "The mechanism" (enforcement's deferred positive face): the audit line is latent until a consumer reads the aggregate; today the consumer and the writer resolve different files, and a third shape drops records entirely. MISSION's isolation model names the audit log per-project by name.

## Plan (composed — the contest ruled a composition)

- **Writer**: dedicated audit-destination resolver in `hookrt.sh`, called only by `audit_log`; `ghjig_state_dir_cli` (and `/file-review` staging) untouched. Rungs: `GHJIG_STATE_DIR_OVERRIDE` → `$CLAUDE_PROJECT_DIR/.claude/ghjig-state` → main-worktree top via `git rev-parse --path-format=absolute --git-common-dir` (submodule guard, `--show-toplevel` fallback) → `BASH_SOURCE` self-location. Legacy write branch deleted.
- **Reader**: `resolve_audit_log` mirrors the rungs; legacy floor read-only, only when the per-project file is absent; `/audit` resolves via the shared reader.
- **Wrapper**: the deny() compensating `CLAUDE_PROJECT_DIR` export in `scripts/ghjig_file_review_post.sh` is deleted as obsolete; staging path/TTL/one-shot semantics untouched.
- **Smoke census**: §20a–d and §53a–c re-seed via `GHJIG_STATE_DIR_OVERRIDE`; §135d moves to the new reader order; §357 tripwire extends to snapshot the live per-project sink; §148i-audit rewritten (asserts resolver-routed audit, not the deleted export); new arms: two-sided visibility, reader preference, worktree write.
- **Migration**: one-time dedup-guarded plain append of this repo's floor into the per-project file (byte-identical-line dedup; measured overlap zero — `comm -12` in the contest record), operational and evidenced here, not a commit.

## Alternatives considered

Contest record (SPEC §4.8):
- **Plan A** (base): route `audit_log` through a worktree-aware `ghjig_state_dir_cli` (shared resolver).
- **B1** (axis: correctness): **dominates-A** — A's smoke-arm census missed the legacy-write-pinning arms (§20a–d, §53a–c), the §357 tripwire guards only the legacy sink, §135d reds, and A's tier-4 kept the `/`-rooted drop; rival = A's mechanism + those closures.
- **B2** (axis: robustness/risk): **dominates-A** — A's staging-collision mitigation is false at head (staging is a fixed un-keyed leaf pinned by §148h), A's shared-resolver change creates a new cross-session staging-destroy channel, and the measured drop was audit-only; rival = decouple (dedicated audit resolver, staging untouched).
- **Verdict**: `refine — carry the composed winner forward` (B2's decoupled resolver + B1's census closures, tier-4 self-location, overlap-zero migration); the judge also found the shared blindspot §148i-audit and verified the deny-export deletion improves per-project isolation without touching the hook tier.

## Key context

- `SPEC.md` §3.2.2 (the rewritten contract; third failure shape recorded there), §6.1 tail, §5.9, §7, §8 tree
- `.claude/hooks/hookrt.sh` — `audit_log`, the new resolver's Code-phase home; `scripts/lib/audit_log_path.sh` — the reader mirror
- `scripts/ghjig_file_review_post.sh` — deny() export deletion (Code phase)
- `scripts/test/smoke.d/` — §20/§53/§135d/§357/§148i-audit rewrites + new arms (Test phase)
- Issue #725 activation round-1 comment — the third failure shape and its re-derivation commands

## Checklist

- [x] **Doc**: SPEC §3.2.2 rewrite + every falsified SPEC site swept (incl. the three non-SPEC surfaces the manifest missed: TROUBLESHOOTING, ESCAPE_HATCH, README); audit.md via `resolve_audit_log`; `changelog_unreleased/fixed/725.md`
- [x] **Test**: census rewrites + §357 live-sink extension + §148i-audit recut + new §187a–d; red at head on exactly the six unimplemented-contract arms (135d, 148i-audit, 187a, 187b, 187c-floor, 187d)
- [x] **Code**: `ghjig_audit_dir` resolver + `audit_log` routing (legacy branch deleted, structurally locked); reader mirror + floor; deny-export deleted; §83c recut (deviation, see Decisions)
- [x] **Code**: migration append ran (222 lines, byte-line dedup, re-run appends 0 — idempotent; `audit-migration` record emitted); both-shapes evidence below

## Out of scope

Audit record schema; promotion thresholds; #639's substrate (ADR 0007 ground unchanged — both files stay git-excluded); archaeology of the divergence; recovering records already lost to torn-down worktrees (accepted loss, recorded in §3.2.2); `ghjig_state_dir`/cache/registry resolution; §5.7.1 and §148i-cli wording (ruled unaffected; the latter is #695's subject).

## Risks

- **Sequencing vs #704**: both the fix and the status quo append with `>>`; unifying the destination widens concurrent writers on one file while #704 (consumer torn-line skip) is open — the consumers' torn-line hardening is that issue's subject, noted here so the two land compatibly.
- **Smoke self-pollution**: under an unconditional writer, an env-unset `audit_log` with cwd in the shell repo writes the live per-project file — exactly why the census closures re-seed via the override seam and §357 now snapshots the live sink.
- **git dependency**: rung 3 adds one `git rev-parse` only on the non-hook path; older git or non-repo cwd degrades down the rungs, never aborts.

## Open questions

None blocking (the contest resolved A's three: migration = plain append with byte-line dedup; §148i-cli wording stays with #695; SPEC:947 kept with annotation).

## Decisions

- 2026-08-27 — composition over either dominator: decoupled audit resolver (staging untouched) + full smoke-census closure; deny-export deleted as the compensation for the exact conditional this PR removes.
- 2026-08-27 — §83c recut rode the Code commit (Phase-B miss, reported by the implementer): the arm pinned the retired legacy-write behavior verbatim and its fixture was the sole §357 live-sink pollution source; recut to the §3.2.2 contract (non-repo cwd + override seam).
- 2026-08-27 — migration evidence (post-append, both shapes byte-identical):

```
$ CLAUDE_PROJECT_DIR=$PWD bash scripts/promotion_candidates.sh   # and: env -u CLAUDE_PROJECT_DIR …
promotion candidates (reviewer-reject reason-class aggregation; LIVE-only; min_rejects=3):
  plan-review | schema-incomplete | rejects=16
  activation | evidence-insufficient | rejects=12
  issue-review | evidence-insufficient | rejects=5
  plan-review | evidence-insufficient | rejects=4
  plan-review | unverifiable-ac | rejects=3
pin: 4399375a50a093704e677677854413a3dab68c9e 2026-08-27Z
```

  AC4's shape: `resolve_audit_log ""` returns the per-project file and a hook-fired `block` record is readable through it (verified at the same pin).

## Test plan

- [x] Rewritten census arms green under the override seam; §357 caught §83c's live-sink write during Phase C and the recut cleared it
- [x] Two-sided visibility: env-unset write from a fixture repo is visible to the env-set reader (§187a)
- [x] Worktree arm: a write from a linked worktree lands in the main tree's ghjig-state (§187b)
- [x] Reader preference: per-project beats populated floor; absent per-project falls to floor (§187c/-floor)
- [x] Both-shapes `promotion_candidates.sh` identical output post-migration, pinned in Decisions
- [x] Full suite green — `pass=1418 fail=0` at the Code commit

## Docs touched

- [x] SPEC §3.2.2 + swept sites; audit.md; TROUBLESHOOTING; ESCAPE_HATCH; README

## Ship gate

- [x] All tests pass — `pass=1418 fail=0` at `d775893` (local + reviewer scratch-clone reproduction)
- [x] code-reviewer passed — round-1 `ship after fix` (judged list on this PR); both fix-now items applied; round-2 re-certification `ship` at `d775893`; §4.11 security tier 3/3 approve re-tallied at the same head
- [x] Docs in sync — doc-writer sweep + reviewer doc-sync finding (F1) fixed and re-verified
- [x] PR body curated
- [x] CI green — 6/6 checks at `d775893`


ilgyu-yi:
## Finding triage (round 1)

reviewed-head: 4399375a50a093704e677677854413a3dab68c9e
prior-round: none (first round) — `scripts/ghjig_judged_list.sh rounds 731` printed no fact lines, no anomalies, `next=1`
durable: PR path — caller posts this reply verbatim via `scripts/ghjig_judged_list.sh post 731 <file>` before any fix is written

```
finding: F1 — SPEC §6.7 Evidence-routing asserts _lib.sh's seed equals the audit resolver's rung 3, and _lib.sh's comment describes the deleted ghjig_state_dir routing
site: SPEC.md#§6.7-evidence-routing (+ .githooks/_lib.sh#cpd-seed)
axis-key: SPEC.md#§6.7-evidence-routing::false-equivalence
verdict: confirmed
evidence: executed=scratch repro (git init A + worktree W; source head hookrt.sh; ghjig_audit_dir with CPD unset vs CPD=$(show-toplevel); then audit_log) mode=function-sourced matches-artifact-mode=yes result=rung 3 resolved .../A/.claude/ghjig-state (main tree, common-dir derive) while the _lib.sh-style seed resolved .../W/... (worktree top) and audit_log WROTE worktree-local — the two rules diverge in a linked worktree and the record dies at teardown, so "same rule" and "regardless" are both false there. Sentence is PR-authored (git diff base..head -- SPEC.md shows the rewrite). _lib.sh:25-29 comment falsified: head audit_log routes via ghjig_audit_dir, never ghjig_state_dir/legacy-shared (hookrt.sh diff).
action: fix-now
remedy-survives: yes — for the two prose recuts (state the divergence or delete the equivalence clause; deletion-default per repo prose doctrine). The optional third limb (re-seed _lib.sh from common-dir) is NOT part of this fix — see C-R1 below. Caution for the author: do not re-import §3.2.2:896's "survives teardown" phrasing unqualified — it holds only on the CPD-unset rung.
axis: implicit↔explicit
target-position: high
justification: §1.10 measurably-false-claim rule (definitional): a durable artifact may not assert an equivalence execution refutes; the divergence is measured above (A vs W). Independent of any failure direction.
swing: not-evaluated
note (check 6, stated honestly): prose-about-code class, but not a §1.10(c) volatile claim (no live count/census/self-measurement), and there is no resolved prior round — recurrence rule not armed; the instance remedy (recut/delete) survives.
```

```
finding: F2 — §50a smoke header says it "Scans .claude/audit/audit.jsonl" while the arm scans $SMOKE_AUDIT
site: scripts/test/smoke.d/30-dirmode-clusters.sh#§50a-header
axis-key: scripts/test/smoke.d/30-dirmode-clusters.sh#§50a-header::stale-comment
verdict: confirmed
evidence: executed=git show <head>:...30-dirmode-clusters.sh (sed 1100-1150) + git show <base>:... mode=script-file matches-artifact-mode=yes result=header names the legacy path; arm sets AUDIT_FILE="$SMOKE_AUDIT"; byte-identical at base — pre-existing, not introduced
action: drop
remedy-survives: n/a
axis: n/a  target-position: n/a  justification: n/a
swing: n/a
```

```
finding: F3 — bare audit_log calls under set -euo pipefail face a new non-zero return path
site: scripts/onboard_target.sh#audit_log-callers (+ scripts/migrate_v3.sh)
axis-key: scripts/onboard_target.sh#audit_log-callers::fail-path
verdict: confirmed (as observation; reviewer's own ruling "not a regression" stands)
evidence: executed=git show <head>:.claude/hooks/hookrt.sh (ghjig_audit_dir + audit_log bodies) mode=script-file matches-artifact-mode=yes result=return-1 requires total rung failure; rung 4 self-locates from BASH_SOURCE, which resolves whenever hookrt.sh exists on disk — unreachable in the stated contexts
action: drop
remedy-survives: n/a
axis: n/a  target-position: n/a  justification: n/a
swing: n/a
```

```
finding: B-L1 — inherited GIT_DIR/GIT_WORK_TREE redirects the audit record into an arbitrary other repo (sec B, Low; subsumes sec A-L1's env limb)
site: .claude/hooks/hookrt.sh#ghjig_audit_dir (+ scripts/lib/audit_log_path.sh#resolve_audit_log)
axis-key: .claude/hooks/hookrt.sh#ghjig_audit_dir::env-trust
verdict: confirmed
evidence: executed=scratch repro (cd nonrepo; GIT_DIR=$R/B/.git; source head hookrt.sh; ghjig_audit_dir) mode=function-sourced matches-artifact-mode=yes result=resolved .../B/.claude/ghjig-state — cwd not in any repo, destination steered entirely by inherited env; with GIT_DIR unset it degraded cleanly to rung-4 self-location, rc=0
action: fix-now
remedy-survives: yes — `env -u GIT_DIR -u GIT_WORK_TREE -u GIT_COMMON_DIR git rev-parse …` in both resolvers. Measured/traced impact: (1) `git grep GIT_DIR <head> -- scripts/ .claude/hooks/ .githooks/ scripts/test/` → zero hits, no fixture sets it; (2) the .githooks tier never reaches these git calls — _lib.sh seeds CLAUDE_PROJECT_DIR before sourcing hookrt.sh, so ghjig_audit_dir returns at rung 2 (traced at head); (3) neutralized fallback behavior measured safe (R2b above).
axis: permissive↔strict
target-position: high
justification: extreme position independently grounded (rule 3): audit_log_path.sh's own header states the definitional contract "The inherited ambient env is never an input (#539)" — GIT_DIR steering the same resolver violates it; plus the zero-impact measurements above. Not derived from any prior failure's direction. The surface is PR-introduced (rung 3 / the consumer's git rung are new), so the fix belongs in this PR.
swing: not-evaluated
```

```
finding: A-L1 — git-view steering incl. hostile core.worktree relocates the derived destination (sec A, Low)
site: .claude/hooks/hookrt.sh#ghjig_audit_dir
axis-key: .claude/hooks/hookrt.sh#ghjig_audit_dir::env-trust
verdict: confirmed
evidence: executed=same scratch repro as B-L1 (env limb); core.worktree limb per reviewer's own probe (did not steer from the repo dir; requires standing in the attacker-designated worktree) mode=function-sourced matches-artifact-mode=yes result=env limb reproduced; payload is one advisory append as the invoking user, no enforcement consumer
action: drop — env limb subsumed by B-L1's fix-now; the core.worktree limb is the accepted-risk class the reviewer named (standing inside a hostile repo already concedes more than an advisory append), and env -u does not and need not close it
remedy-survives: n/a
axis: n/a  target-position: n/a  justification: n/a
swing: n/a
```

```
finding: A-L2 — the mkdir+append follows a symlinked ghjig-state (sec A, Low)
site: .claude/hooks/hookrt.sh#audit_log
axis-key: .claude/hooks/hookrt.sh#audit_log::symlink-follow
verdict: confirmed
evidence: executed=scratch repro (repo C, .claude/ghjig-state → $R/target symlink, CPD unset, audit_log via sourced head hookrt.sh) mode=function-sourced matches-artifact-mode=yes result=rc=0, audit.jsonl created under $R/target — write followed the symlink
action: drop — identical posture pre-existing on the CPD rung (reviewer verified same at head~); payload is an advisory append as the invoking user with no enforcement consumer; a lone `[ ! -L ]` on the final component is incomplete hardening (any parent component can be the symlink) and would buy false assurance; reviewer demanded no remediation
remedy-survives: n/a
axis: n/a  target-position: n/a  justification: n/a
swing: n/a
```

```
finding: A-I1 — rung 3 seeds .claude/ghjig-state/audit in an unonboarded repo with no info/exclude entry; git add -A can commit audit reasons (sec A Info; = sec B-I1)
site: .claude/hooks/hookrt.sh#ghjig_audit_dir
axis-key: .claude/hooks/hookrt.sh#ghjig_audit_dir::unonboarded-seed
verdict: confirmed
evidence: executed=scratch repro (fresh repo D, no onboarding; audit_log via sourced head hookrt.sh; git -C D status --porcelain + check-ignore) mode=function-sourced matches-artifact-mode=yes result=`?? .claude/` untracked, check-ignore says NOT ignored — a later git add -A stages the audit line
action: defer-to-issue — both security lenses flagged it; the remedy space (write the exclude entry at first per-repo write, vs. gate rung 3 on an onboarded marker) is a design decision touching onboarding semantics, not a recut this PR owes
remedy-survives: n/a (no in-PR remedy proposed)
axis: n/a  target-position: n/a  justification: n/a
swing: n/a
```

```
finding: B-I1 — same unonboarded-repo seed surface (sec B Info)
site: .claude/hooks/hookrt.sh#ghjig_audit_dir
axis-key: .claude/hooks/hookrt.sh#ghjig_audit_dir::unonboarded-seed
verdict: confirmed
evidence: same executed repro as A-I1
action: defer-to-issue — same single issue as A-I1 (one issue, not two)
remedy-survives: n/a  axis: n/a  target-position: n/a  justification: n/a
swing: n/a
```

```
finding: A-I2 — relative override value yields a cwd-relative dir (sec A Info)
site: .claude/hooks/hookrt.sh#ghjig_audit_dir
axis-key: .claude/hooks/hookrt.sh#ghjig_audit_dir::override-shape
verdict: confirmed
evidence: executed=git show <head>:.claude/hooks/hookrt.sh (rung 1 prints GHJIG_STATE_DIR_OVERRIDE verbatim) mode=script-file matches-artifact-mode=yes result=pre-existing shape shared with ghjig_state_dir; the override is a harness-owned test seam, not reachable by runtime input
action: drop
remedy-survives: n/a  axis: n/a  target-position: n/a  justification: n/a
swing: n/a
```

```
finding: B-I2 — deny() now routes the reject record by caller cwd, not the wrapper-resolved esd; cross-repo attribution drift (sec B Info)
site: scripts/ghjig_file_review_post.sh#deny
axis-key: scripts/ghjig_file_review_post.sh#deny::attribution
verdict: confirmed
evidence: executed=git diff base..head -- scripts/ghjig_file_review_post.sh mode=script-file matches-artifact-mode=yes result=the `export CLAUDE_PROJECT_DIR="${esd%…}"` prefix is removed; audit_log resolves via its own resolver from process context. Drift arises only when the wrapper is invoked with cwd parked in a foreign repo — not a supported invocation shape; record still lands per-project
action: drop
remedy-survives: n/a  axis: n/a  target-position: n/a  justification: n/a
swing: n/a
```

```
finding: B-R1 — worktree of a bare main (and git <2.31) falls back to --show-toplevel, record dies at teardown (sec B Residual)
site: .claude/hooks/hookrt.sh#ghjig_audit_dir
axis-key: .claude/hooks/hookrt.sh#ghjig_audit_dir::teardown-loss
verdict: confirmed
evidence: executed=git show <head>:.claude/hooks/hookrt.sh (rung 3: `case "$gcd" in */.git)` fails for a bare common dir → --show-toplevel fallback) + git show <head>:SPEC.md §3.2.2:896 mode=script-file matches-artifact-mode=yes result=fallback confirmed by code; SPEC §3.2.2 names the family "Records already lost to those shapes are unrecoverable — accepted loss"
action: drop — inside the documented accepted-loss family; availability-only, advisory records
remedy-survives: n/a  axis: n/a  target-position: n/a  justification: n/a
swing: n/a
```

```
finding: C-R1 — the teardown-loss closure holds only on the env-unset rung: an agent-worktree CPD, and _lib.sh's --show-toplevel pre-seed, defeat rung 3's main-tree pin (sec C Residual; behavioral half of F1)
site: .githooks/_lib.sh#cpd-seed
axis-key: .githooks/_lib.sh#cpd-seed::teardown-loss
verdict: confirmed
evidence: executed=scratch repro R1b/R3 (seed → W top; audit_log wrote worktree-local) mode=function-sourced matches-artifact-mode=yes result=loss shape reproduced; identical at head~ (pre-#725 the same worktree-local write happened via ghjig_state_dir under the same seed) — pre-existing, not introduced
action: defer-to-issue — the re-seed of _lib.sh from the common-dir derive is a behavior change to the git-hook tier that also retargets its OTHER ghjig_state_dir consumers (caches, registry) to the main tree; that scope question deserves its own issue, not a rider on a prose-fix commit. Anti-swing check, run explicitly: re-seeding does not reverse any decision this PR took — the PR's rung order (CPD before git-derive) and resolver split are untouched; the seed is a different site. No opposite-end move.
remedy-survives: n/a (in-PR); the prose halves of this subject are discharged by F1's fix-now
axis: n/a  target-position: n/a  justification: n/a
swing: n/a
```

```
finding: C-I1 — once the per-project file exists, floor-resident historical records stop counting toward aggregates (sec C Info)
site: scripts/lib/audit_log_path.sh#resolve_audit_log
axis-key: scripts/lib/audit_log_path.sh#resolve_audit_log::floor-shadowing
verdict: confirmed
evidence: executed=git show <head>:scripts/lib/audit_log_path.sh (floor consulted only when per-project file absent) + SPEC §3.2.2:898 mode=script-file matches-artifact-mode=yes result=behavior as described and documented as a deliberate completeness trade ("only when the per-project file is absent, never in preference to it")
action: drop
remedy-survives: n/a  axis: n/a  target-position: n/a  justification: n/a
swing: n/a
```

**Summary for the author** — fix-now set (2): F1 (recut the SPEC §6.7 sentence + the `_lib.sh:25-29` comment; state the divergence or delete the false clause; no behavior change) and B-L1 (env-neutralize the git calls in `ghjig_audit_dir` and `resolve_audit_log`). Defer-to-issue set (2 issues): unonboarded-repo exclude-entry (A-I1/B-I1, one issue) and the `_lib.sh` re-seed decision (C-R1, one issue). Everything else drops with grounds recorded above. No verdict token is altered: `code-reviewer`'s `ship after fix` and the 3/3 security approves stand as issued.

Closing observation (not a judged record): in the R2b measurement, rung 4's self-location resolved relative to wherever `hookrt.sh` sits — expected, but the author recutting F1 should not describe rung 4 as "the shell root" in target-repo prose without the §3.2.1 override caveat already present in §3.2.2.

dispatch: findings produced by one code-reviewer + three security-reviewer invocations (all worktree-isolated, artifact-only, reviewed-head blind-compared MATCH by the caller; security tally 3/3 valid approve at the same head); judgment produced by a separate worktree-isolated finding-judge invocation that authored none of the findings; the caller filtered/reordered/paraphrased nothing.

ilgyu-yi (commented):
## Review verdict: approve

`code-reviewer` returned `ship` on the round-2 re-certification at the head this review pins. Round 1 at `4399375` returned `ship after fix` with two judged fix-now items (round-1 judged list on this PR); both are verified applied by execution at `d775893`: the SPEC §6.7 and `.githooks/_lib.sh` recuts state the measured worktree divergence (fixture-reproduced: the seed derives the current-worktree top, rung 3 the main-worktree top) without importing any unqualified survives-teardown claim, and the inherited-git-env steering is closed (`env -u` on all three git calls in both resolvers, pre-fix controls at the prior head steering to the attacker repo; `ghjig_state_dir_cli` byte-identical). The fix delta is exactly four files with the resolver changes being pure env-prefix additions. The §4.11 merge-security-surface tier re-tallied fresh at this head: 3/3 valid approve (gate arms byte-identical except the deny() export deletion; block-integrity sabotage probes held; cross-project isolation and degradation rungs measured clean). Suite `pass=1418 fail=0` at the head, reproduced in a reviewer scratch clone; CI 6/6.

Issue #738, opened 2026-08-27
the activation and completion label flips succeed with the reviewer step silently omitted


ilgyu-yi:
Parent Directive: #692

## What

The two near-free Layer-1 gates (parent items 1 and 2): today `gh issue edit <N> --remove-label status:proposed` and `gh issue close <N> --reason completed` (on a `directive`-labelled issue) succeed with the reviewer step silently omitted — the approval is command-prose only, stated verbatim in the carriers ("no PreToolUse hook reads `directive-review`"). Both actions already have machine-readable evidence artifacts nothing consumes:

1. **activation gate** — refuse `--remove-label status:proposed` unless a `<!-- activation-verdict: pass -->` comment exists post-dating the issue's last body/label change (the freshness predicate `/activate`'s loop-safety skip already computes).
2. **completion gate** — refuse `gh issue close --reason completed` on a `directive`-labelled issue unless the `## Directive Completion` / completion-pass comment `/complete-directive` produces exists. (Today `trusted-filer-mutate` waves `--reason completed` through with no evidence check.)

Both matchers mirror `merge-review` (#586): PreToolUse, fail-closed on any `gh`/lookup failure, §7-escapable with an audit record. Per the parent's evidence-before-action constraint, `/activate`'s pass path reorders to post the pass comment **before** removing labels (at head it removes first), so the gate finds the artifact when the action fires — this session's own `/activate` runs already practice that order.

## Why

- **MISSION fit** — via the parent (#692, "The mechanism"): hooks the agent cannot drift past without leaving evidence; these two actions are the lifecycle's irreversible-adjacent flips (Proposed→Active; Directive→Completed) currently guarded by prose alone.
- **Why now** — the parent activated today with these as its near-free items: the evidence artifacts exist, are machine-readable, and the freshness predicate is already specified in the carrier; the gate is consumption, not invention.
- **Existing coverage** — none: `trusted-filer-mutate` checks `--reason completed` presence, not evidence; no matcher reads activation-verdict comments (the parent's own census).

## Acceptance criteria

- [ ] AC1 — parent signal 1's two-sided pair holds: `--remove-label status:proposed` with no fresh pass comment is blocked naming the missing evidence and the `/activate` remedy; with a fresh one, allowed. Freshness = the pass-marker comment post-dates the last body/label modification (the loop-safety predicate).
[...]

Issue #741, opened 2026-08-27
gh-command matchers over-match a gated form quoted as data, and two issue-target parsers drift


ilgyu-yi:
## What

Every `gh`-command PreToolUse matcher greps the raw command string, so a command that merely **mentions** a gated form in a data position (a quoted string, a heredoc, a `-m` message) fires the matcher. For the fail-open matchers this was cheap (a spurious pass-through); for the new fail-closed evidence gates (#738) it turns a mention into a real block plus a live `gh` lookup, and — because the shared issue-number resolver shlex-tokenizes the whole compound command — the remedy message can name the wrong issue.

Two coupled sites, one command-parsing-tolerance defect (PR #739 round-1 judged list, findings F1 + F8, both `defer-to-issue`, executed evidence pinned there):

1. **Entry-grep over-match.** The matcher entry/refine greps match text anywhere in the command; the fix shape is segment-level / command-position matching — the `push_segments` precedent (#446) that splits a compound command into executed segments before matching.
2. **Resolver drift.** `resolve_gh_issue_target` (helpers/issue_type.sh) anchors `gh issue <verb>` adjacency without the #499 leading-global-flag tolerance, so `gh --repo o/r issue close …` resolves an empty issue; #738's `gh_issue_target` compensates with a **second** widened selector parser beside the shared resolver — the #276 drift class (two parsers for one job). Widening the shared resolver's anchor and deleting the duplicate fallback is the durable fix.
3. **Folded in (F8):** `_evidence_issue_gql` makes two `gh repo view` calls (owner, name) where one `--json owner,name` would do — same helper, benign latency, one-line fix.

## Why

- **MISSION fit** — the enforcement layer's precision: a gate that fires on a data-position mention imposes over-block friction (every session writing fixtures/docs about these commands hits it) and, via the resolver drift, can misname its own remedy. The fail direction is safe (over-block, never a false allow), which is why this is a precision/hygiene issue, not a security one — but the two parsers are a maintenance liability the #276 centralization exists to prevent.
- **Why now** — #738 just made the over-match consequential (fail-closed) and introduced the duplicate parser as a disclosed temporary compensation; its PR Decisions section names this filing as the durable home so the fallback stays temporary.
- **Existing coverage** — none: the matchers share the raw-string grep by construction; no open issue addresses segment-level matching or the resolver anchor.

## Acceptance criteria

[...]

Issue #718, opened 2026-08-25
four judged residuals on the evidence helper: path relativity, has_nul disposition, timeout zero, pin scanning


ilgyu-yi:
## What

Four residuals on `scripts/ghjig_evidence.sh`, each ruled `defer-to-issue` by PR #717's round-1 judged list (the evidence — executed probes and head-blob reads — is pinned in that comment):

1. **Header says "repo-relative path"; resolution is cwd-relative.** The quote face resolves via `git ls-files -- "$path"` and `git show "HEAD:./$path"`, both cwd-relative; from a repo subdirectory a repo-relative path fails closed rather than resolving. Either resolve against the toplevel or reword the header — the header is the contract's single code home, so the imprecise term is a Directive #637 L5 accuracy matter.
2. **`has_nul`'s internal failure is fail-open.** It returns 2 on a `wc -c` failure and is consumed as `if has_nul …; then die`, so an internal failure proceeds as "no NUL" — a fail-open branch inside an otherwise fail-closed emitter. Practically unreachable (a local readable temp file), ruled by construction.
3. **`GHJIG_EVIDENCE_TIMEOUT=0` means instant kill.** Zero passes the digits-only gate; every command times out immediately (measured: rc=2, "timed out after 0s"). Decide: refuse 0, or state it as a sanctioned instant-timeout knob.
4. **First-match pin scanning reads a forged interior pin.** The canonical recovery is position-anchored (the pin is the last content line before the closing fence), but a naive `grep '^pin: ' | head -1` consumer — including the suite's own extraction helper — reads a payload-forged interior pin. Anchor every consumer to position, starting with the suite helper.

5. **NUL-tail no-eol misclassification** (PR #717 round-2 judged list): `[ -n "$(tail -c 1 …)" ]` drops a trailing NUL in command substitution, so evidence output ending in NUL is classified newline-terminated and emits a malformed block at rc=0 (pin concatenated onto the NUL tail, no marker). Remediation direction: refuse NUL in evidence output (mirror `has_nul`) or a byte-safe final-byte check.
6. **C0 control bytes reach the attribution and command lines unrefused** (round-2 judged list): a CR-bearing path or command renders a display-layer forged attribution/`$`-line in a terminal; bytes and fence stay intact. Remediation direction: extend the newline refusals to CR (ideally all C0 controls) in `run_quote` and `run_evidence` — the marker-lookalike rendering caveat (`\ no-eol\r`, trailing space) folds in for free.

[...]

PR #746, opened 2026-09-22
fix(#745): pin the evidence gates' GraphQL round-trip to the repo host

### LINKED ISSUE #745: evidence 게이트의 GraphQL 호출이 host-pin되지 않아 GHES 타깃에서 상시 차단된다
## What

`_evidence_issue_gql` (`.claude/hooks/helpers/ac_closeout_gate.sh:597`) calls `gh api graphql` **without `--hostname`**, while every other `gh api` call in the same file is host-pinned via `_ac_repo_host` — `merge_is_self` (`:132-134`), `review_gate_accepts` (`:445-450`, `:491`). This one site was missed.

`gh repo view` / `gh issue view` infer the host from the remote; `gh api` does not. So a host-less call resolves gh's **default host**, and on a GHES target the repository does not resolve — the response carries `errors[].type == "NOT_FOUND"` and `repository == null`.

`activation_evidence_fresh` and `completion_evidence_present` both map a non-empty `.errors` or `issue == null` to **2 (lookup failure)**, and their callers fail closed. That is the designed behaviour — the defect is not the gate, it is that the gate reads its value from the wrong host.

**Consequence**: on a GHES target, the Proposed→Active label flip guarded by `activation-evidence` and the Directive close-as-completed guarded by `completion-evidence` are **both permanently fail-closed**. The only route through is the §7 escape, which disables the gate rather than satisfying it.

## Why

MISSION **"Multiple repos run on it"** — at least two unrelated upstream repos adopting the shell as their canonical workflow is a success condition, and one of them lives on GHES. When host-inferring `gh` subcommands and non-inferring `gh api` are mixed inside one file, that adoption splits by host.

MISSION **"The flow holds in unattended runs"** is directly implicated too: with both dir-mode lifecycle gates fail-closed, an unattended run cannot pass those two checkpoints without an escape.

**Why now** — a regression introduced by `ec6fed3` (PR #739, closes #738). Not latent: the whole dir-mode lifecycle on a GHES target stops at those two checkpoints **today**. Waiting a quarter means every activation and completion in that period passes only by escape, which is the normalisation MISSION **"The escape hatch stays narrow"** exists to exclude.

## Acceptance criteria

- [ ] `_evidence_issue_gql`'s `gh api graphql` call is pinned to the host `_ac_repo_host` resolves. A resolution failure (empty / malformed) returns non-zero **fail-closed** and never falls back silently to the default host — the same posture `review_gate_accepts` already uses.
- [ ] Measured on the dual-host `gh` shim: with the repo on the GHES host, `activation_evidence_fresh` **discriminates fresh / absent / stale** rather than collapsing to 2 (lookup failure). Under Doc→Test→Code the Test commit records that this case is RED without the pin.
- [ ] github.com no-regression — the same case stays GREEN in the variant where the default host already is the repo host.
- [ ] `scripts/test/smoke.sh` is green.

## Out of scope

- **#741 AC3 already reserves the same function** — folding `_evidence_issue_gql`'s two `gh repo view` calls into one `--json owner,name`. A different defect, but **the same few lines**, so the two branches overlap; whichever lands first, the other rebases onto it.
- Changing `_ac_repo_host`'s own resolution rules — used exactly as #610 defined them.
- The evidence gates' predicates (trusted author, line-1 marker, freshness comparison) — owned by #738.
- The four residuals on `scripts/ghjig_evidence.sh` — owned by #718 (a different file).

## Notes

- Prior art: #610 (the host-pin discipline and smoke §153), #738/#739 (the evidence gates).
- `scripts/test/smoke.d/90-ghes-host.sh` already carries §153 — it covers `ghjig_file_review_post.sh`, `review_gate_accepts` and `merge_is_self`, with a dual-host shim and two static content-locks. The evidence gates landed after #610 and are absent from that list.

ilgyu-yi:
Closes #745

## Goal

Pin the dir-mode evidence gates' single `gh api graphql` round-trip to the repo host. Unpinned, it resolves gh's default host, so on a GitHub Enterprise Server target the repository never resolves and both `activation-evidence` and `completion-evidence` fail closed on *lookup-failure* rather than on evidence. The Proposed→Active flip and a Directive completed-close are reachable there only through the §7 escape.

MISSION *"Multiple repos run on it"* — at least two unrelated upstream repos adopting the shell is a success condition, and one of them lives on GHES. When host-inferring `gh` subcommands and non-inferring `gh api` are mixed inside one file, adoption splits by host. *"The flow holds in unattended runs"* is implicated directly (an unattended run cannot pass those two checkpoints), and *"The escape hatch stays narrow"* names the cost of the status quo: today the only route through disables the gate rather than satisfying it.

## Plan

`_evidence_issue_gql` issues the gates' one round-trip with no `--hostname`. The three sibling `gh api` sites in the same file are already pinned via `_ac_repo_host` under #610; this site was missed when #738/#739 added the gates.

**The pin is branch-scoped, and that is the whole design.** `_ac_repo_host` resolves the **cwd** repo's host, not the target's. An unconditional pin would send an explicit `--repo ilgyu-yi/ghjig-claude` target to a GHES host and mint a **new** wrong block on a path that works today. So the pin applies on the cwd-derived branch — where owner, name and host all come from one repo and are coherent by construction — while an explicit `-R`/`--repo owner/name` selector is left unpinned and resolves gh's default host. That is also what the sibling REST read in `completion_evidence_present` resolves for the same selector, so both halves of that gate agree on one host per target.

A consequence for the test: driving the case through `--repo` exercises the branch the fix deliberately leaves alone and is green before and after. The RED case is **GHES cwd, no `--repo`**.

## Alternatives considered

- **Plan A** (base): branch-scoped pin reusing `_ac_repo_host`, new smoke section with its own dual-host shim.
- **B1** (axis: correctness): built a rival that generalises `_ac_repo_host` to take an `owner/name` and pins uniformly. Withheld rather than submitted, as a zero-gain diff. Reasoned concession.
- **B2** (axis: robustness/risk): a genuine non-dominating alternative — fold owner+name+host into one `gh repo view --json owner,name,url`, halving the gh calls on the gated path and turning a forgotten test-shim patch from silently-green into loudly-RED. **Rejected on scope**: it annexes #741 AC3, which this issue puts explicitly out of scope. A plan may not delete another issue's AC — that is an SSOT edit, not a plan decision.
- **Verdict**: build A with three carries — initialise `host=""`; correct the risk accounting (below); and name no follow-up issue in the SPEC rows. B2's call-count win is routed by widening #741 AC3 to `--json owner,name,url` so the fold lands once, in its own issue.

## Checklist

- [x] **Doc**: both §6.1 rows record the pin, the no-default-host-fallback abort, the explicit-selector semantics, and `an unresolvable repo host` in each lookup-failure cause list; `changelog_unreleased/fixed/745.md` (`f4c314b`)
- [x] **Test**: new smoke section driving the cwd-derived path on a dual-host shim (`8fbce23`, `903aac4`) — fresh/absent/stale discriminated on a GHES cwd instead of collapsing to lookup-failure, a degenerate-host variant that must not accept the default host's answer, github.com no-regression, and the explicit-`--repo` path pinned as-is
- [x] **Code**: branch-scoped pin in `_evidence_issue_gql` (`8a40e52`, `53ccfa8`) with `host=""` initialised and a `set -u`-safe optional-argv expansion; name the new cause inside the two existing lookup-failure messages, no third message
- [x] dev gates green — `scripts/lint.sh` 92 files clean, `scripts/test/smoke.sh` 1448/0

## Risks

- **The `timeout 5` bound is not a guarantee.** `_ac_run_gh` falls back to **unbounded** `gh` when neither `timeout` nor `gtimeout` is on PATH — both are homebrew coreutils, absent on stock macOS. The pin adds one host-resolution call per gated fire on the cwd-derived branch, so that unbounded-stall surface widens by one call until #741 AC3's fold retires it.
- **A new fail-closed path**: a remote-less clone or an unauthenticated `gh` now fails the gate at host resolution. The same branch already calls `gh repo view` twice and already returns non-zero in those states, so this adds a third flake window rather than a new failure class — and the direction is the safe one under §6.0 P1.
- **Cross-host explicit selectors stay unfixed**: a GHES target named by `-R`/`--repo` from any cwd still resolves at the default host, on both the GraphQL and the REST halves. Pre-existing and untouched here; #747 owns threading the selector's host through.
- **Merge-order coupling with #741**: the change appends beside the two `gh repo view` lines #741 AC3 replaces and does not touch them, so either landing order is a small adjacency.

## Out of scope

- #741 AC3 — folding the two `gh repo view` calls into one.
- `_ac_repo_host`'s own resolution rules (#610).
- The gates' predicates — trusted-author set, line-1 marker, freshness comparison (#738).
- `scripts/ghjig_evidence.sh` residuals (#718).


ilgyu-yi:
## Finding triage (round 1)

reviewed-head: 8a40e5209fc8b87b1b8de1de3270e9982a68a73a

prior-round: none (first round) — `bash scripts/ghjig_judged_list.sh rounds 746` → `next=1`, no `round=` fact lines, no anomaly lines. `swing:` is therefore `not-evaluated` on every `fix-now` item (a `swing: none` would be vacuous), and check 6's recurrence test is `not-evaluated` throughout.

durable: PR #746. My scratch write was refused by the out-of-scope Edit/Write matcher (`edit outside registry blocked: …/scratchpad/746r1/n1_repro.sh`), so per the dispatch this reply **is** the artifact — hand it verbatim to `scripts/ghjig_judged_list.sh post 746 <file>` before the author writes any fix. It carries no marker of its own.

Head re-resolved twice (start and end of judgment): unchanged. All source read via `git show 8a40e520…:<path>`; every run's cwd and writes inside the session scratchpad, never the ambient tree. Git read-only throughout.

---

## Judged list

```
finding: J1 — SPEC.md's activation-evidence row carries a duplicated verb introduced by the Doc commit ("The round-trip carries carrying …")
  [code-reviewer F1 · sec-A L1 · sec-B L3 · sec-C L1]
site: SPEC.md#§6.1-activation-evidence-row
axis-key: SPEC.md#§6.1-activation-evidence-row::claim-text
verdict: confirmed
evidence: executed=git show 8a40e52:SPEC.md | grep -n 'carries carrying' mode=script-file matches-artifact-mode=yes result=one hit, SPEC.md:2002 (activation row only; the completion row at :2003 is clean). Base row read the same way says "…derive from one `gh api graphql` round-trip carrying `lastEditedAt`…" — the Doc commit spliced "The round-trip carries" in front and left "carrying". Mechanism exactly as the reviewer stated.
action: fix-now
remedy-survives: yes (one-word deletion; no prose added)
axis: none — discrete (a duplicated verb is a present/absent token; there is no continuum to position on)
target-position: n/a
justification: the SSOT row this PR rewrote is ungrammatical at head; the remedy is a deletion inside the row the author is already editing.
swing: n/a
```

```
finding: J2 — the same row still claims the lookup-failure message "carries `merge-review`'s re-run remedy verbatim", and this PR's own Code commit made that false
  [code-reviewer F2 · sec-A L2]
site: SPEC.md#§6.1-activation-evidence-row
axis-key: SPEC.md#§6.1-activation-evidence-row::claim-accuracy
verdict: confirmed
evidence: executed=grep -n "then re-run" on `git show 8a40e52:.claude/hooks/pre_tool_use.sh` and on `git show 7027f90:…` mode=script-file matches-artifact-mode=yes result=
  merge-review (:616, unchanged) — "Confirm the PR exists and gh is reachable, then re-run."
  activation @ base (:985) — "Confirm the Issue exists and gh is reachable, then re-run."   ← same shape, PR↔Issue only
  activation @ head (:985) — "Confirm the Issue exists, **that this directory is a git checkout with a gh-authenticated remote**, and that gh is reachable, then re-run."
  So the word was approximately true at base and is structurally false at head. `grep -o "re-run remedy[^)]*"` on base and head SPEC: "verbatim" is present in both — the Doc commit did not add the word, but it rewrote the whole row and kept it, and the Code commit then invalidated it. code-reviewer's causal account is the accurate one; sec-A's "drift the PR inherits rather than creates" is half right (inexact pre-PR) but understates it — the *material* falsification is this PR's.
action: fix-now
remedy-survives: yes — **the deletion form only**
axis: none — discrete (a claim about another file's text is true or false at head; §1.10(c) licenses exactly two remedy forms for it, remove or pin — a closed pair, not a continuum)
target-position: n/a
justification: §1.10(c) remedy bias, read at SPEC.md:597 ("The remedy for a stale volatile claim is removal or pinning, never a refresh"). Deleting `verbatim` leaves the weaker true claim, which is **already the completion row's own wording** at :2003 ("blocks with `merge-review`'s re-run remedy", no "verbatim") — so the deletion converges the two rows rather than inventing a form. **Explicitly does NOT survive: restating the new message text in SPEC.** That is the refresh form; it re-arms the identical finding the next time either message changes, which is the loop §1.10(c) exists to terminate.
swing: not-evaluated
```

```
finding: J3a — the inline comment's leading clause ("gh's own `--repo owner/name` host semantics apply instead") misstates the mechanism; the same misstatement sits in the SPEC row
  [sec-C M1 comment-half + L2 · contested by sec-A ("not false") · sec-B (silent on this clause)]
site: .claude/hooks/helpers/ac_closeout_gate.sh#_evidence_issue_gql
axis-key: .claude/hooks/helpers/ac_closeout_gate.sh#_evidence_issue_gql::comment-accuracy
verdict: confirmed
evidence: executed=(1) `gh api --help | grep -E '^\s+(-R|--repo|--hostname)'` mode=script-file matches-artifact-mode=yes result=**only** `--hostname string  The GitHub hostname for the request (default "github.com")` — gh 2.92.0's `api` subcommand has **no** `-R`/`--repo` flag at all. (2) awk-extracted `_evidence_issue_gql` body at head, grepped for selector tokens → the argv is `api graphql ${host_flag[@]+"${host_flag[@]}"} -f query=… -f owner="$owner" -f name="$name" -F number="$n"`; the only `--repo`/`-R` strings in the whole function body are **inside the comment**. owner/name travel as GraphQL *variables*, not as a repo selector.
  So: no selector reaches `gh api graphql`, and its host is gh's default host — not anything derived from `--repo` semantics. Decisive corroboration without leaving the function: the branch-B comment ten lines below says "`gh api` does no repo inference — unpinned it resolves gh's DEFAULT host." The two comments in one function contradict each other, and branch-B's is the true one (it is also the issue body's and #610's finding).
action: fix-now
remedy-survives: yes for sec-C's remedy (i) — "correct the comment to say the branch resolves gh's default host". **Does NOT survive: sec-C's remedy (ii)** ("record the host-prefixed-selector divergence as a documented residual") and **sec-B's** ("carry it honestly in the §6.1 rows") — see the axis below. **sec-A's ruling that there is no false statement to fix is refuted** on this clause.
axis: narrow↔wide
target-position: low (narrow)
justification: both endpoints named. **Low** = substitute the mechanism in place, equal-or-fewer words, no new sentence in either surface. **High** = add a documented-residual paragraph to the comment and the §6.1 rows. Low is chosen independently, on two grounds, neither of them the direction of any prior failure: (a) §1.10(c), SPEC.md:597 — a residual paragraph describing a hole scheduled for a follow-up is a claim that expires before the artifact, so writing it now buys a finding later; (b) the residual's durable home is the follow-up Issue (J4), and §9's thin-pointer norm puts the contract in one place, not two. The wide end is not forbidden in principle — it is rejected because its cost is measurable and its content is J4's.
swing: not-evaluated
```

```
finding: J3b — the trailing clause ("the same host the sibling REST read in `completion_evidence_present` resolves for this selector, so both halves of a gate agree on one host per target")
  [sec-C M1 calls it wrong · sec-A calls it true · sec-B calls it true but "reads as an all-clear"]
site: .claude/hooks/helpers/ac_closeout_gate.sh#_evidence_issue_gql
axis-key: .claude/hooks/helpers/ac_closeout_gate.sh#_evidence_issue_gql::comment-accuracy-trailing
verdict: refuted (the clause is TRUE as written)
evidence: executed=read `completion_evidence_present` and `gh_issue_target` at head via `git show 8a40e52:…` mode=script-file matches-artifact-mode=yes result=the sibling REST read is `_ac_run_gh issue view "$n" --repo "$repo" …`, and `$repo` is **always hostless** by the time it arrives — `gh_issue_target` (:566) and `resolve_gh_issue_target` (issue_type.sh:352-360) both run `case "$repo" in */*/*) repo="${repo#*/}" ;;` (#283). A hostless `--repo o/r` on `gh issue view` resolves gh's default host; an unpinned `gh api graphql` resolves gh's default host. Same host. Corroborated by all three reviewers' independent measurement that a GHES cwd + `--repo o/n` returns rc=2 on **both** predicates — i.e. both reads landed on the default host and NOT_FOUNDed together.
action: n/a (refuted — retained in the record, excluded from the fix list, not convertible to defer-to-issue)
remedy-survives: n/a
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
note: the three-way split settles as — **sec-A is right that this clause is true and right that it never claims gate-host == mutation-host; sec-A is wrong that the sentence therefore contains no false statement (J3a does).** sec-B is right on the truth value; its "reads as an all-clear" is an editorial judgment about J3a's false leading clause, which J3a's remedy removes. sec-C is right about the sentence and wrong about this clause. One imprecision not worth a record: "both halves of **a gate**" literally holds only for the completion gate — the activation gate has one read, no REST half.
```

```
finding: J4 — a host-prefixed `-R`/`--repo HOST/owner/name` or URL selector has its host stripped, so the gate's evidence lookup can resolve a different host than the one gh routes the mutation to (forged-evidence wrong-ALLOW)
  [sec-A M1 · sec-B Medium · sec-C M1 behaviour-half · code-reviewer N5]
site: .claude/hooks/helpers/ac_closeout_gate.sh#gh_issue_target
axis-key: .claude/hooks/helpers/ac_closeout_gate.sh#gh_issue_target::cross-host-confusion
verdict: confirmed — and confirmed **unchanged by this diff**
evidence: executed=`git diff 7027f90…8a40e52 --stat` + a grep of the ac_closeout_gate.sh hunk for `gh_issue_target`/`#283`/host-prefixed mode=script-file matches-artifact-mode=yes result=6 files, +288/-7; `.claude/hooks/helpers/issue_type.sh` (home of `resolve_gh_issue_target`) is **not in the diff at all**; the grep over the changed hunks returns nothing — `gh_issue_target` is untouched. The explicit-selector branch of `_evidence_issue_gql` gains only comment lines; its argv is unchanged because `${host_flag[@]+"${host_flag[@]}"}` expands to zero words on an empty array (all three security reviewers probed this on /bin/bash 3.2.57 under `set -u`). Both stripping sites read at head and quoted above (J3b). The exploit shape all three constructed is coherent and I did not attempt it against a live host.
action: defer-to-issue
remedy-survives: yes — the shared remediation (thread the selector's own host through `gh_issue_target`/`resolve_gh_issue_target` and pin `--hostname` to it, or fail closed when it differs from `_ac_repo_host`). **sec-B's "carry it in the §6.1 rows" arm does NOT survive** — see J3a's axis.
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
note: **nothing is owed in this PR — not a code change and not a SPEC line.** Grounds: (a) genuinely pre-existing and byte-unchanged, verified above; (b) #745's own Out-of-scope names `_ac_repo_host`'s resolution rules, and the fix changes a resolver shared by several §6.1 matchers — that is a separate Execution Issue, not a carry on a `fix` PR; (c) the SPEC-line option is the wrong shape, because the row is not silent on this branch, it is **wrong** about it (J3a) — the owed SSOT edit is the correction, not an addition. **A follow-up Issue is owed**, because the PR body's Risks bullet already points at this residual and J8 trims its overstatement; leaving neither the claim nor the issue would drop the thread. Filing it does not gate `gh pr ready`.
```

```
finding: J5 — §189-6's static content-lock greps the function body for a bare `--hostname`, and this PR's Code commit added ten comment lines inside that same body
  [code-reviewer N1 · sec-C L3]
site: scripts/test/smoke.d/90-ghes-host.sh#§189-6
axis-key: scripts/test/smoke.d/90-ghes-host.sh#§189-6::vacuous-pass
verdict: confirmed
evidence: executed=the arm's own three lines (awk body-extractor + `grep -q -- '--hostname'` + the `api graphql` token grep), lifted verbatim from 90-ghes-host.sh:693-696 and run over three gate bodies in scratch mode=script-file matches-artifact-mode=yes result=
  gate_head.sh          : pin=1 tok=1 -> ok (GREEN)
  gate_base.sh          : pin=0 tok=1 -> ng (RED)      ← the arm's declared RED
  gate_base_comment.sh  : pin=1 tok=1 -> ok (GREEN)    ← **vacuous pass**
  The mutant's entire delta over base is one line (`diff` output: `587a588 >   # NOTE: unpinned; see #610 for the --hostname idiom`) — a pure comment, no code. So an unpinned round-trip greens the arm.
  `scripts/test/smoke.sh:18-23` names this by name as vacuous-pass anti-pattern #1 ("Comment-satisfiable grep … passes when `token` appears only in a COMMENT/prose") and prescribes the fix ("Anchor the grep to the code form … not a bare token").
action: fix-now
remedy-survives: yes — **the comment-strip arm only** (strip `^[[:space:]]*#` lines before the grep). **The `grep -q -- 'host_flag=(--hostname'` code-form-anchor arm does NOT survive**: §189-6's own header commits the arm to "`--hostname` must appear (the host pin, **in whatever shape the Code chooses** — a literal flag pair, an array, a variable run)", and that anchor pins one shape, breaking the arm's stated contract.
axis: permissive↔strict
target-position: interior
justification: both endpoints measured/quoted. **Too permissive** — the bare token greens on a comment over unpinned code (measured above, `pin=1 tok=1` on a comment-only mutant). **Too strict** — a literal `host_flag=(--hostname` anchor contradicts the arm's own declared shape-tolerance, quoted from its header. The interior position (strip comments, keep the shape-tolerant token) is the only one satisfying both. Independent of any prior failure's direction; the ground is the suite's own documented rule plus the measurement.
swing: not-evaluated
note on "must it be fixed now, given four sibling arms carry measured detection": **yes.** The sibling arms establish that there is no correctness exposure *today*; they say nothing about the arm's job, which is tomorrow. Three things make this PR's own problem: the arm is **new in this PR**; it violates a rule **this suite's own header** states with its fix; and the PR's Code commit made that exact function body comment-dense, so the drift path is not hypothetical here — it is the shape the body already has. The cost is one edit with no prose growth, and it rides a head advance that J1/J2/J3a force anyway (see head-advance below), so its marginal cost is zero. Per the repo's nit norm this is fix-and-ship: no re-review of the arm itself, no escalation.
```

```
finding: J6 — the PR body's "Alternatives considered → B1" asserts "measured it against A across a five-case matrix … and found them identical on all five" with no evidence shipped
  [code-reviewer N2]
site: PR#746-body#Alternatives-considered
axis-key: PR#746-body#Alternatives-considered::self-claim-evidence
verdict: confirmed
evidence: executed=`gh pr view 746 --json body -q .body` + `sed -n '597,615p' SPEC.md` mode=script-file matches-artifact-mode=yes result=the sentence is present verbatim and carries no pinned command, no output, no artifact; the B1 diff was withheld, so nothing in the repo witnesses the comparison. SPEC §1.10(d) at :610-615 puts it exactly: a coverage claim "ships the counted number and the command that counted it", and "the bare-prose form of the same claim is refused at authoring — not hedged, not deferred to review, not written."
action: fix-now
remedy-survives: yes (delete the sentence; keep "withheld as a zero-gain diff")
axis: none — discrete (§1.10(d) is explicitly two-sided — ship the evidence or do not write it, "not hedged" — so there is no intermediate position)
target-position: n/a
justification: §1.10(d) as quoted. Deletion rather than evidence-attachment because the compared artifact was never committed: there is no corpus a pinned command could re-run against, so the evidence half is unavailable and the rule's other side applies.
swing: n/a
```

```
finding: J7 — the PR body's Risks call-count figure "3 → 4 on the gated path" is path-specific
  [code-reviewer N3]
site: PR#746-body#Risks
axis-key: PR#746-body#Risks::volatile-figure
verdict: confirmed
evidence: executed=call-site count on `git show 8a40e52:.claude/hooks/helpers/ac_closeout_gate.sh` mode=script-file matches-artifact-mode=yes result=activation path = repo view owner, repo view name, **new** `_ac_repo_host` repo view url, graphql → 3→4. Completion path = `gh issue view --json labels` (:666) **first**, then the same four → 4→5. The figure is true for one of the two gated paths. Independently matches sec-B's measured "activation 3→4 bounded calls, completion 4→5".
action: fix-now
remedy-survives: yes — **the deletion arm and the generalise-to-"one more call per gated fire" arm both survive; correcting the figure to "4 → 5" does NOT.**
axis: none — discrete (§1.10(c) licenses exactly two remedy forms for a volatile figure — remove, or pin — a closed pair)
target-position: n/a
justification: §1.10(c), SPEC.md:597 and its first bullet ("An exact count belongs in a test assertion, where going stale fails the arm instead of leaving the prose lying"). The figure additionally expires *by the PR body's own account* — the same bullet says #741 AC3's fold "retires it back to 3" — so a corrected count is false on a landing the body already anticipates. Check 6's recurrence test is `not-evaluated` here (no resolved prior round), so this ruling rests on §1.10(c) directly, not on recurrence.
swing: not-evaluated
```

```
finding: J8 — the PR body's claim that the cross-host explicit-selector residual is "documented in the §6.1 rows as present-tense behaviour" overstates the rows
  [code-reviewer N4]
site: PR#746-body#Risks
axis-key: PR#746-body#Risks::claim-accuracy
verdict: confirmed — and stronger than the reviewer put it
evidence: executed=`grep -n "keeps gh's own selector host semantics" SPEC.md` + full read of the activation row at head mode=script-file matches-artifact-mode=yes result=one hit, SPEC.md:2002 (the completion row at :2003 carries no such clause at all). The row states a **mechanism** and never the consequence — nothing in either row says a GHES target named by `-R`/URL hard-blocks, and nothing says the evidence host can diverge from the mutation host. And per J3a the mechanism it does state is **false**, so the rows do not merely under-document the residual, they mis-describe the branch.
action: fix-now
remedy-survives: yes (trim the PR-body claim). code-reviewer's paired judgment — "I would **not** add prose to SPEC for this" — survives and is the correct half; see J3a's axis.
axis: none — discrete (the row either states the consequence or it does not; measured above that it does not)
target-position: n/a
justification: §1.10(a)/(d) — a body claim about this change's own SSOT edit that does not resolve at the attributed file. The trim must follow J3a's SPEC correction so the bullet describes what the row then actually says.
swing: not-evaluated
```

```
finding: J9 — the unbounded-`gh` fallback when neither `timeout` nor `gtimeout` is on PATH; this PR adds one call to the chain (measured 90s at head vs 60s at base)
  [sec-B L2 · sec-C L4 · sec-A I1]
site: .claude/hooks/helpers/ac_closeout_gate.sh#_ac_run_gh
axis-key: .claude/hooks/helpers/ac_closeout_gate.sh#_ac_run_gh::unbounded-fallback
verdict: confirmed (the fallback and the +1 call), and confirmed **pre-existing as a class**
evidence: executed=read `_ac_run_gh` at head (:63-77) mode=script-file matches-artifact-mode=yes result=`if [ -n "$timeout_bin" ]; then "$timeout_bin" 5 gh "$@"; else gh "$@"; fi` — the bare-`gh` else arm is present and unchanged by this diff. The +1 call is J7's arithmetic. sec-B's 90s/60s is a stopwatch on that same +1, and it terminated fail-closed in both shapes.
action: drop
remedy-survives: n/a (no reviewer proposed a change here; all three call it out of scope, and sec-B names the real hardening — a bash-level watchdog in `_ac_run_gh` — as out of scope)
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
note: not the author's to fix in this PR. The PR body's Risks section **already discloses it accurately** ("The `timeout 5` bound is not a guarantee … that unbounded-stall surface widens by one call until #741 AC3's fold retires it back to 3"), and #741 AC3 is the named home for the retirement. sec-B's escalation (a blackholed host outrunning the PreToolUse budget → a killed hook → fail-open) is a real class and is true at base as well as head, by sec-B's own account — it belongs to `_ac_run_gh`, not to #745. If the author wants it durable, it rides J4's follow-up; it does not gate this PR.
```

```
finding: J10 — `_ac_repo_host`'s charset guard admits the literal `null` (and a leading-`-` token); this PR widens its reachability
  [sec-B L1]
site: .claude/hooks/helpers/ac_closeout_gate.sh#_ac_repo_host
axis-key: .claude/hooks/helpers/ac_closeout_gate.sh#_ac_repo_host::charset-guard
verdict: confirmed
evidence: executed=`for h in null -x gh.example.com "a/b" "a b"; do case "$h" in ''|*[!A-Za-z0-9.:-]*) echo REJECTED;; *) echo ACCEPTED;; esac; done` (the guard's exact pattern, read from :94-96) mode=script-file matches-artifact-mode=yes result=`null` ACCEPTED, `-x` ACCEPTED, `gh.example.com` ACCEPTED, `a/b` REJECTED, `a b` REJECTED. Not an injection vector: `/`, space, quote and `$` are rejected, and `host_flag=(--hostname "$host")` binds the token as one argv word, so no second flag can be smuggled. Consequence is fail-closed (no auth for the bogus host → non-zero → rc 2 → block).
action: drop
remedy-survives: n/a (sec-B proposes no change and labels it pre-existing #610)
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
note: not the author's to fix. `_ac_repo_host`'s resolution rules are named in **both** the Issue's and the PR's Out-of-scope lists ("used exactly as #610 defined them"), the consequence is in the safe direction, and the one place it was actually doing harm — the §76 shim staying green on `--hostname null` — this PR already closed by adding `url` to the shim's emit blob. Nothing left owed. If the author files J4, it can ride along as a #610 residual.
```

```
finding: J11 — the new §189 shim prepends `$S745_BIN` to PATH but never asserts `command -v gh` resolves inside it
  [sec-A I2]
site: scripts/test/smoke.d/90-ghes-host.sh#§189-shim
axis-key: scripts/test/smoke.d/90-ghes-host.sh#§189-shim::path-residency
verdict: confirmed (the assertion is absent), but correctly scoped by the reviewer as suite-wide hygiene
evidence: executed=`grep -n 'S745_BIN\|command -v gh\|PATH=' 90-ghes-host.sh` at head mode=script-file matches-artifact-mode=yes result=§189 exports `PATH="$S745_BIN:$PATH"` at :581/:591/:601 with no residency assertion — byte-identical idiom to the pre-existing §153/§154 sites at :181/:189/:199/:361/:411, none of which assert it either.
action: drop
remedy-survives: n/a (sec-A proposes none; "not a #745 defect" is its own label)
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
note: not the author's to fix here — the PR conforms to the established idiom, and changing it would be a five-site suite-wide edit outside #745.
```

```
finding: J12 — §189-4 is green before and after the Code commit, i.e. vacuous as a RED
  [sec-C L5 · noted by code-reviewer and sec-B]
site: scripts/test/smoke.d/90-ghes-host.sh#§189-4
axis-key: scripts/test/smoke.d/90-ghes-host.sh#§189-4::declared-no-break-lock
verdict: confirmed as a fact, and confirmed already-disclosed
evidence: executed=read the §189 section header at head (90-ghes-host.sh:442-463) mode=script-file matches-artifact-mode=yes result=the header states it in advance — "driving a case through `--repo` exercises the branch the fix leaves alone (green before AND after — vacuous as a RED) … the explicit-selector case (§189-4) is a no-break lock, not a defect witness." Two reviewers additionally measured it green on their revert-mutants.
action: drop
remedy-survives: n/a (all three reviewers rule the labelling honest; none proposes a change)
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
```

```
finding: J13 — the PR body under-claims: it frames the change as an availability fix, while two security reviewers independently measured that it closes a wrong-ALLOW present at base
  [caller-posed, from sec-B's §189-2 base-vs-head measurement and sec-C's same-named-repo probe]
site: PR#746-body#Goal
axis-key: PR#746-body#Goal::under-claim
verdict: the underlying measurement is confirmed; the **under-claim is refuted as a defect**
evidence: executed=read the PR body, the §189 section header and §189-2's comment at head mode=script-file matches-artifact-mode=yes result=the base wrong-ALLOW is real (a GHES cwd derives owner/name from its own repo, then the unpinned graphql resolves those names at gh's default host; a same-named dotcom repo carrying a trusted line-1 pass satisfies the gate while the mutation lands on GHES) and it is independently witnessed twice. But **no statement in the PR body is false because of the omission** — the Goal describes the lookup-failure consequence, which is true, and claims no security-neutrality. The durable surfaces already carry the property in present tense: SPEC.md:2002 states "an unresolvable host aborts before the call and never falls back to gh's default host", and §189-2's own comment states "a silent default-host fallback would surface there as a wrong ALLOW". §1.10 binds false, unevidenced and expiring claims; an omission is none of the three.
action: drop
remedy-survives: no — **the author must NOT add the security framing to the PR body.**
axis: n/a
target-position: n/a
justification: n/a
swing: n/a
note: adding it would be strictly self-defeating under this repo's own rules. The sentence would be a same-change self-claim about the change's own security effect, so §1.10(d) (SPEC.md:610-615) would require it to ship "one attempted violation and its red" — i.e. the base-vs-head measurement pinned into the body. That evidence **already lives where §1.10(c) says it belongs**: §189-2 is a committed arm, RED at base and GREEN at head, which is precisely the machine-pair form. Writing the prose copy duplicates a test into a body and grows the artifact for nothing. This is the prose growth the norm pushes back on. Leave it.
```

**Rolled up, no record (assertions of cleanliness, not findings):** sec-A I3 (no new deps/egress), sec-B I1/I2/I3 (no sensitive disclosure; escape text byte-identical; bash 3.2 array idiom correct), sec-C L6/L7 (no injection surface; changelog fragment well-formed). I spot-checked the two that could have been wrong — the bash-3.2 empty-array expansion and the changelog fragment (`changelog_unreleased/fixed/745.md`: category `fixed`, stem `745` = closing issue, one `- ` bullet containing `(#745)`) — both hold.

---

## What must happen before `gh pr ready`

**Fix now — head-advancing (bundle into ONE commit):**
- **J1** SPEC.md:2002 — delete the duplicated verb.
- **J2** SPEC.md:2002 — delete `verbatim` (not: restate the message).
- **J3a** `ac_closeout_gate.sh` `_evidence_issue_gql` comment **and** SPEC.md:2002 — replace the false mechanism clause with gh's-default-host in place; add nothing.
- **J5** `90-ghes-host.sh` §189-6 — strip comment lines before the `--hostname` grep (not: anchor to a code form).

**Fix now — no head advance (PR body only; refetch the remote body first and abort on external edit, per PR-as-living-doc):**
- **J6** delete the five-case-matrix sentence. **J7** delete or generalise the call-count figure. **J8** trim the "documented in the §6.1 rows" claim, after J3a lands.

**Owed but not in this PR:** **J4** — file the follow-up Issue for the host-stripped selector (threading the selector's host through `gh_issue_target`/`resolve_gh_issue_target`, or failing closed on divergence). No head advance; does not gate ready.

**Must not be touched:** J4's code and its SPEC-residual paragraph · J9 unbounded-`gh` fallback (already disclosed; #741 AC3's) · J10 `_ac_repo_host` charset guard (out of scope in both Issue and PR) · J11 shim PATH residency (conforms to the existing idiom) · J12 §189-4 (already declared) · **J13 the PR body's framing — do not add the wrong-ALLOW claim.**

**Head-advance consequence.** The four head-advancing fixes touch three files and cost **exactly one** head advance if committed together; there is no cheaper split, because J1/J2/J3a alone already force it. That advance staleifies all four `reviewed-head: 8a40e52…` pins (§4.5's fail-closed pin; `/ship` step 1's blind compare invalidates a vote at a stale head), so **one re-review round at the new head** is owed before `gh pr ready` regardless of how many of these land. That is the economic argument for including J5 rather than deferring it: its marginal head cost is zero. The three PR-body fixes cost nothing and can land after the commit (order: commit → PR body update). Ship mode is `attended`, so the terminal step is `gh pr ready` and the `merge-review` matcher is not in the path.

**Not altered by me:** `code-reviewer`'s `ship after fix` and the three `approve` tokens stand as written. I cast no vote and emit no verdict token.

**Closing observation (not a judged record — no reviewer raised it).** The false clause J3a rules on exists in **two** places, and only the code comment was flagged. `SPEC.md:2002` carries the same misattribution — "the call **keeps gh's own selector host semantics** — the same host the sibling REST read in the completion arm resolves for that selector" — verified by `grep -n "keeps gh's own selector host semantics" SPEC.md` → one hit, the activation row; the completion row at :2003 has no such clause. If the author fixes only the comment, the SSOT keeps the false mechanism and the next round re-finds it. It sits inside the same row J1 and J2 already require editing, so correcting it costs no extra edit and no extra head advance — but it is the single most likely thing to be missed in the fix pass.

